PR #22, opened 2022-10-05
Revert Apache License

10allday:
Don't apply software licenses to code that has no copyright.


glommer:
Sorry, as I have explained to you on Discord, we're making different choices, since we have different goals. As one of our goals is to maximize contributions, an explicit OSS license works better.

PR #85, opened 2022-12-10
libsql build process automation for RHEL

tjyang:
Adding less intrusive bootstrap script for RHEL under contrib sub directory.


tjyang:
let me resubmit another one later.

PR #176, opened 2023-06-25
release pipeline for linux x86_64, aarch64, darwin arm64

dyasny:
See the generated artefacts at https://github.com/libsql/libsql-dev/releases/tag/v0.1.39

The new release flow will be as follows: 
- On a push to main, a new tag will be created as a release draft, the tag is autogenerated based on the current largest tag in the repo, e.g. with 0.2.1 right now, the new one will be 0.2.2. This will be a draft release, and every PR pushed to main will be added to the draft. 
- Once we are ready to release, edit the draft release (change the tag if the automatic one isn't right) and hit the "publish" button. 
- This will create a release with the given tag and autogenerated note, and trigger a workflow that builds the artefacts.
- The workflow creates an aarch64 AWS VM, and also uses our self hosted linux and mac machines. 
- As each build completes it will add the generated artefacts to the release.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,31 @@
+name-template: 'v$RESOLVED_VERSION'
+tag-template: 'v$RESOLVED_VERSION'
+categories:
+  - title: '🚀 Features'
+    labels:
+      - 'feature'
+      - 'enhancement'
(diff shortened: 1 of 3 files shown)


dyasny:
@psarna the failing check is a chicken/egg situation, you'll just have to ignore and accept it :)

PR #210, opened 2023-07-17
libsql: Fix Python package name

penberg:
We want developers to `import libsql` in Python, so fix up `Cargo.toml` to reflect that.


Abdur-rahmaanJ (approved):
Yes, see i am using it here for example

https://github.com/Abdur-rahmaanJ/phoning-home/blob/69f86240d7d3f46783a108d76c088fa0963c462f/phoning_home/ph.py#L1

Issue #187, opened 2023-07-07
LibSQL API for Rust


penberg:
This is a tracking issue for the whole LibSQL API for Rust.

**Features**

- [x] Positional parameters
- [x] Finalize prepared statements
- [x] Named parameters
  - [x] Owned/init impl
  - [x] By reference
- [x] Blobs types
- [x] `query_map()` API
- [x] Improve error handling
- [x] Transactions

penberg:
I am closing this ticket because we have a Rust API that works fine. Will open tickets for further improvements.

PR #234, opened 2023-07-25
libsql/core: Add named parameter support

LucioFranco:
This adds basic named parameter support via `Vec<(String, Value)>`.

Ref #187

### DIFF
--- a/crates/core/tests/integration_tests.rs
+++ b/crates/core/tests/integration_tests.rs
@@ -1,4 +1,4 @@
-use libsql::{params, Connection, Database, Params};
+use libsql::{named_params, params, Connection, Database, Params};
 
 fn setup() -> Connection {
     let db = Database::open(":memory:").unwrap();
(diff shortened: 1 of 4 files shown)


crates/core/tests/integration_tests.rs
```diff
@@ -57,7 +81,9 @@ fn check_insert(conn: &Connection, sql: &str, params: Params) {
     conn.execute(sql, params).unwrap();
     let rows = conn.execute("SELECT * FROM users", ()).unwrap().unwrap();
     let row = rows.next().unwrap().unwrap();
-    assert_eq!(row.get::<i32>(0).unwrap(), 1);
+    // Use two since if you forget to insert an id it will automatically
+    // be set to 1 which defeats the purpose of checking it here.
+    assert_eq!(row.get::<i32>(0).unwrap(), 2);
```
LucioFranco, line 86: I made this change because I was really confused when my test was not failing when I was not inserting the first column (It gets auto added via primary key in the schema). So I changed this to ensure we capture any missed params.

PR #261, opened 2023-08-09
 libsql/core: Add support for transactions

penberg:


crates/core/src/connection.rs
```diff
@@ -103,6 +103,19 @@ impl Connection {
         }
     }
 
+    /// Begin a new transaction in DEFERRED mode, which is the default.
+    pub fn transaction(&self) -> Result<Transaction> {
+        self.transaction_with_behavior(TransactionBehavior::Deferred)
+    }
+
+    /// Begin a new transaction in the given mode.
+    pub fn transaction_with_behavior(
```
LucioFranco, line 112: imho I would add a version of this fn for each enum variant and then make the enum private. Its 1) cleaner and 2) more backwards compat w/ adding new variant (which I assume in this case is likely never to happen).

crates/core/src/transaction.rs
```diff
@@ -0,0 +1,91 @@
...
+        match self.drop_behavior {
+            DropBehavior::Rollback => {
+                self.do_rollback().unwrap();
+            }
+            DropBehavior::Commit => {
+                self.do_commit().unwrap();
+            }
+            DropBehavior::Ignore => {}
+            DropBehavior::Panic => {
+                panic!("Transaction dropped without being committed or rolled back");
```
LucioFranco, line 37: You need to check if we are panicking already https://doc.rust-lang.org/stable/std/thread/fn.panicking.html before you can panic or else a double panic is really bad.  And same with the unwraps above.

crates/core/src/transaction.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+impl Deref for Transaction {
+    type Target = Connection;
+
+    #[inline]
+    fn deref(&self) -> &Connection {
+        &self.conn
+    }
+}
```
LucioFranco, line 91: Why this over exposing a `as_connection` fn? iirc using deref for this is bad practice.

LucioFranco (dismissed):

crates/core/src/transaction.rs
penberg: This is just mirroring the rusqlite API.

crates/core/src/connection.rs
penberg: This is just mirroring the rusqlite API.

PR #274, opened 2023-08-11
Update critierion to remove atty

LucioFranco:


LucioFranco:
We need to wait for `cbindgen` to upgrade from clap 3 to 4 to completely remove atty. https://github.com/libsql/libsql/security/dependabot/11

PR #287, opened 2023-08-16
libsql: Use `SQLITE_TRANSIENT` for text/blob

LucioFranco:


crates/libsql-sys/src/statement.rs
```diff
@@ -52,7 +52,7 @@ impl Statement {
                 idx,
                 value.as_ptr() as *const c_char,
                 value.len() as i32,
-                None,
+                SQLITE_TRANSIENT(),
```
LucioFranco, line 55: Rusqlite is smarter here and checks if the string is empty and uses the static version but I think for now this is fine.

PR #289, opened 2023-08-16
libsql/shell: commands completion

piotrkira:
Simple implementation of tab completion logic based on interfaces provided by rustyline. List of commands to complete is hardcoded but I guess we could put list of all commands in some shared variable and reuse it where needed.


penberg:
Very cool, thanks @piotrkira!

PR #298, opened 2023-08-21
libsql/core: Add `v2` api

LucioFranco:
This PR adds a new in-progress v2 rewrite of the libsql API to support both a sqlite3 and http based backend. None of this code affects the current API and is written in such a way to migrate over easily.

The current approach to abstract both the sqlite3 and http api is to use internal trait objects so that the public types do not contain generics. In the future, if we find that trait objects are too slow we can move to enums.

### DIFF
--- a/crates/core/src/v2/mod.rs
+++ b/crates/core/src/v2/mod.rs
@@ -1,0 +1,117 @@
+mod hrana;
+mod rows;
+mod statement;
+
+use std::sync::Arc;
+
+use crate::{Params, Result};
+pub use hrana::{Client, HranaError};
+
+pub use rows::{Row, Rows};
+use statement::LibsqlStmt;
+pub use statement::Statement;
+
+// TODO(lucio): Improve construction via
+//      1) Move open errors into open fn rather than connect
+//      2) Support replication setup
+enum DbType {
+    Memory,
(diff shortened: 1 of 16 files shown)


crates/libsql-sys/src/statement.rs
```diff
@@ -104,9 +104,6 @@ impl Statement {
 
     pub fn column_origin_name(&self, idx: i32) -> Option<&str> {
         let raw_name = unsafe { crate::ffi::sqlite3_column_origin_name(self.raw_stmt, idx) };
-        if raw_name.is_null() {
-            return None;
-        }
```
penberg, line 109: What's the reason for this change?

crates/core/src/v2/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    pub fn open(db_path: impl Into<String>) -> Result<Database> {
+        Ok(Database {
+            db_type: DbType::Path {
+                path: db_path.into(),
+            },
+        })
+    }
+
+    pub fn open_http(url: impl Into<String>) -> Result<Self> {
```
penberg, line 42: Probably should be `open_remote()` because you pass `libsql://` URLs to it. Can be fixed later too.

PR #325, opened 2023-08-26
Use `read_exact` instead of `read_exact_at`

avinassh:
I am trying to use libsql API library and aiming to cross compile my app across OS. Currently it uses a *nix OS specific import and can't be used on Windows. This patch tries to fix that. 

- We are using `read_exact_at` but at offset zero, so it is same as `read_exact`
- Earlier, it was reading twice into buffer. Any particular reason for that? I changed it to read it once

### DIFF
--- a/crates/replication/src/replica/meta.rs
+++ b/crates/replication/src/replica/meta.rs
@@ -1,7 +1,7 @@
 use std::fs::{File, OpenOptions};
-use std::io::ErrorKind;
+use std::io::{ErrorKind, Seek};
 use std::mem::size_of;
-use std::os::unix::prelude::FileExt;
+use std::io::Read;
 use std::path::Path;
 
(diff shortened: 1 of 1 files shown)


avinassh:
This PR currently breaks, but #324 fixes the import error

penberg:
Hey @avinassh! The change looks good to me, but I also wonder why we're reading twice. Let's try to get @MarinPostma or @psarna to have a look at this before merging.

MarinPostma:
Looks good. Idk why we were reading twice, likely a mistake

MarinPostma:
Do we want to bother with making sqld compile on windows? Just asking

penberg:
@MarinPostma Sure, I think it's good to keep stuff running on Windows when possible.

Issue #344, opened 2023-09-01
Support for read-only databases in the Rust crate


penberg:

xonx4l:
so this is what I came up with commit-3eaa309102694fb24171382d920c8dc65477781b. Waiting for your feedback @penberg . Thanks for Guiding.

PR #347, opened 2023-09-01
Added:Support for read-only databases in the Rust crate.

xonx4l:
issue -:#344

### DIFF
--- a/crates/core/src/database.rs
+++ b/crates/core/src/database.rs
@@ -47,18 +47,19 @@ pub struct Database {
     pub db_path: String,
     #[cfg(feature = "replication")]
     pub replication_ctx: Option<ReplicationContext>,
+    pub readonly: bool,
 }
 
 impl Database {
     /// Open a local database file.
-    pub fn open<S: Into<String>>(db_path: S) -> Result<Database> {
+    pub fn open<S: Into<String>>(db_path: S, readonly:bool) -> Result<Database> {
         let db_path = db_path.into();
         if db_path.starts_with("libsql:") || db_path.starts_with("http:") {
             Err(ConnectionFailed(format!(
                 "Unable to open remote database {db_path} with Database::open()"
             )))
         } else {
-            Ok(Database::new(db_path))
+            Ok(Database::new(db_path,readonly))
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/core/src/database.rs
```diff
@@ -47,18 +47,19 @@ pub struct Database {
     pub db_path: String,
     #[cfg(feature = "replication")]
     pub replication_ctx: Option<ReplicationContext>,
+    pub readonly: bool,
 }
 
 impl Database {
     /// Open a local database file.
-    pub fn open<S: Into<String>>(db_path: S) -> Result<Database> {
+    pub fn open<S: Into<String>>(db_path: S, readonly:bool) -> Result<Database> {
```
penberg, line 55: We should do a `open_with_flags()` associate function like Rusqlite does for this instead of adding a boolean to `open()` that everybody has to pass.

penberg (changes requested):

crates/core/src/database.rs
```diff
@@ -87,19 +88,36 @@ impl Database {
...
         Connection::connect(self)
+         if self.readonly {
+        Connection::open_with_flags(&self.db_path, rusqlite::OpenFlags::SQLITE_RO)
+    } else {
+        Connection::open(&self.db_path)
     }
+    .map_err(|e| ConnectionFailed(format!("{:?}", e)))
+    }
+
+    pub fn use_database(&self) -> Result<()> {
```
penberg, line 112: This looks like some left-over code.

crates/core/src/database.rs
```diff
@@ -87,19 +88,36 @@ impl Database {
...
+            readonly,
         }
     }
 
     pub fn close(&self) {}
 
     pub fn connect(&self) -> Result<Connection> {
         Connection::connect(self)
+         if self.readonly {
+        Connection::open_with_flags(&self.db_path, rusqlite::OpenFlags::SQLITE_RO)
```
penberg, line 105: You should use `libsql-sys` and `SQLITE_OPEN_READONLY` here.

penberg:
Implemented in commit befd7357c.

PR #358, opened 2023-09-06
libsql/core: Make v1 compeltely private

LucioFranco:
This upgrades all our examples/doc tests to use the new v2 api that now lives at the root of the project.


crates/core/src/v1/mod.rs
```diff
@@ -1,47 +1,3 @@
-//! # libSQL API for Rust
```
LucioFranco, line 1: This module is now private and these comments exist at lib.rs

crates/core/src/v1/connection.rs
```diff
@@ -89,7 +89,7 @@ impl Connection {
     /// Convenience method to run a prepared statement query.
     /// ## Example
     ///
-    /// ```rust,no_run
+    /// ```rust,no_run,ignore
```
LucioFranco, line 92: Ignored for now since these are no longer public, we will eventually completely remove some of these APIs as we will only require what is needed for the async api.

psarna (approved):

PR #363, opened 2023-09-08
libsql/core: Add write delegationt test

LucioFranco:


LucioFranco:
This will fail CI since there is no version of sqld running there but should be possible to run this locally to reproduce.

PR #368, opened 2023-09-11
add open_remote_with_connector

MarinPostma:
add an `open_remote_with_connector` method to use libsql remote with a custom http acceptor.

### DIFF
--- a/crates/core/src/v2/mod.rs
+++ b/crates/core/src/v2/mod.rs
@@ -10,9 +10,13 @@ use crate::v1::{params::Params, TransactionBehavior};
 use crate::Result;
 pub use hrana::{Client, HranaError};
 
+use hyper::Uri;
+use hyper::client::HttpConnector;
+use hyper::service::Service;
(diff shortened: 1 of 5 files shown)


crates/core/src/v2/mod.rs
```diff
@@ -35,23 +39,80 @@ impl Default for OpenFlags {
 // TODO(lucio): Improve construction via
 //      1) Move open errors into open fn rather than connect
 //      2) Support replication setup
-enum DbType {
+enum DbType<C> {
     Memory,
     File { path: String, flags: OpenFlags },
     Sync { db: crate::v1::Database },
-    Remote { url: String, auth_token: String },
+    Remote { url: String, auth_token: String, connector: C },
```
LucioFranco, line 46: Can we box the connector and remove the generic here?

LucioFranco (changes requested):

LucioFranco (approved):

PR #493, opened 2023-10-21
snapshot on scale to zero and shutdown

avinassh:
closes https://github.com/libsql/sqld/issues/757

- I renamed `Database.shutdown` to `Database.destroy`, which sends only the close signal. The new `Database.shutdown` now sends the close signal + waits for snapshot to complete

### DIFF
--- a/libsql-server/src/database.rs
+++ b/libsql-server/src/database.rs
@@ -4,13 +4,20 @@ use crate::connection::libsql::LibSqlConnection;
 use crate::connection::write_proxy::{RpcStream, WriteProxyConnection};
 use crate::connection::{Connection, MakeConnection, TrackedConnection};
 use crate::replication::{ReplicationLogger, ReplicationLoggerHook};
+use async_trait::async_trait;
 
+pub type Result<T> = anyhow::Result<T>;
+
+#[async_trait]
 pub trait Database: Sync + Send + 'static {
     /// The connection type of the database
     type Connection: Connection;
 
     fn connection_maker(&self) -> Arc<dyn MakeConnection<Connection = Self::Connection>>;
-    fn shutdown(&self);
+
+    fn destroy(self);
+
+    async fn shutdown(self) -> Result<()>;
 }
 
 pub struct ReplicaDatabase {
@@ -18,6 +25,7 @@ pub struct ReplicaDatabase {
         Arc<dyn MakeConnection<Connection = TrackedConnection<WriteProxyConnection<RpcStream>>>>,
 }
 
+#[async_trait]
 impl Database for ReplicaDatabase {
     type Connection = TrackedConnection<WriteProxyConnection<RpcStream>>;
 
@@ -25,7 +33,11 @@ impl Database for ReplicaDatabase {
         self.connection_maker.clone()
     }
 
-    fn shutdown(&self) {}
+    fn destroy(self) {}
+
+    async fn shutdown(self) -> Result<()> {
+        Ok(())
+    }
 }
 
 pub type PrimaryConnection = TrackedConnection<LibSqlConnection<ReplicationLoggerHook>>;
@@ -35,6 +47,7 @@ pub struct PrimaryDatabase {
     pub connection_maker: Arc<dyn MakeConnection<Connection = PrimaryConnection>>,
 }
 
+#[async_trait]
 impl Database for PrimaryDatabase {
     type Connection = PrimaryConnection;
 
@@ -42,7 +55,18 @@ impl Database for PrimaryDatabase {
         self.connection_maker.clone()
     }
 
-    fn shutdown(&self) {
+    fn destroy(self) {
         self.logger.closed_signal.send_replace(true);
+    }
+
+    async fn shutdown(self) -> Result<()> {
+        self.logger.closed_signal.send_replace(true);
+        if let Some(replicator) = &self.logger.bottomless_replicator {
+            let replicator = replicator.lock().unwrap().take();
+            if let Some(mut replicator) = replicator {
+                replicator.wait_until_snapshotted().await?;
+            }
+        }
+        Ok(())
     }
 }
--- a/libsql-server/src/lib.rs
+++ b/libsql-server/src/lib.rs
@@ -1,6 +1,8 @@
 #![allow(clippy::type_complexity, clippy::too_many_arguments)]
 
+use std::future::Future;
 use std::path::{Path, PathBuf};
+use std::pin::Pin;
 use std::process::Command;
 use std::str::FromStr;
 use std::sync::{Arc, Weak};
@@ -379,6 +381,7 @@ where
         let snapshot_callback = self.make_snapshot_callback();
         let auth = self.user_api_config.get_auth().map(Arc::new)?;
         let extensions = self.db_config.validate_extensions()?;
+        let namespace_store_shutdown_fut: Pin<Box<dyn Future<Output = Result<()>> + Send>>;
 
         match self.rpc_client_config {
             Some(rpc_config) => {
@@ -394,6 +397,10 @@ where
                 let (namespaces, proxy_service, replication_service) = replica.configure().await?;
                 self.rpc_client_config = None;
                 self.spawn_monitoring_tasks(&mut join_set, stats_receiver, namespaces.clone())?;
+                namespace_store_shutdown_fut = {
+                    let namespaces = namespaces.clone();
+                    Box::pin(async move { namespaces.shutdown().await })
+                };
 
                 let services = Services {
                     namespaces,
@@ -430,6 +437,10 @@ where
                 let (namespaces, proxy_service, replication_service) = primary.configure().await?;
                 self.rpc_server_config = None;
                 self.spawn_monitoring_tasks(&mut join_set, stats_receiver, namespaces.clone())?;
+                namespace_store_shutdown_fut = {
+                    let namespaces = namespaces.clone();
+                    Box::pin(async move { namespaces.shutdown().await })
+                };
 
                 let services = Services {
                     namespaces,
@@ -453,6 +464,7 @@ where
         tokio::select! {
             _ = self.shutdown.notified() => {
                 join_set.shutdown().await;
+                namespace_store_shutdown_fut.await?;
                 // clean shutdown, remove sentinel file
                 std::fs::remove_file(sentinel_file_path(&self.path))?;
                 tracing::info!("sqld was shutdown gracefully. Bye!");
--- a/libsql-server/src/namespace/mod.rs
+++ b/libsql-server/src/namespace/mod.rs
@@ -2,6 +2,7 @@ use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::fmt;
 use std::path::{Path, PathBuf};
+use std::sync::atomic::{AtomicBool, Ordering};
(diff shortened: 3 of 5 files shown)


libsql-server/src/lib.rs
```diff
@@ -365,6 +365,7 @@ where
         let snapshot_callback = self.make_snapshot_callback();
         let auth = self.user_api_config.get_auth().map(Arc::new)?;
         let extensions = self.db_config.validate_extensions()?;
+        let mut primary_namespace_store: Option<NamespaceStore<PrimaryNamespaceMaker>> = None;
```
avinassh, line 368: I couldn't think of a better place than here to keep the NamespaceStore reference

MarinPostma: There is another way you can do that, which doesn't leak the primary implementation details to a wider scope: you put it in a future. You rename this to `namespace_store_shutdown_fut: Pin<Box<dyn Future<Output = Result<()>>>>`, and declare it as:
```rust
namespace_store_shutdown_fut = Box::pin(async move {
    namespace_store.shutdown().await
})
```

in both the replica and primary branch :)

The advantage of that is that if we decide  to extend the shutdown method for replica, everything is already in place to do that

libsql-server/src/namespace/mod.rs
```diff
@@ -495,6 +497,15 @@ impl<M: MakeNamespace> NamespaceStore<M> {
         Ok(())
     }
 
+    pub async fn shutdown(&self) -> crate::Result<()> {
```
MarinPostma, line 500: you should mark the store as shut down to prevent other namespaces to be reallocated after this is called

libsql-server/src/replication/primary/logger.rs
```diff
@@ -266,13 +273,16 @@ unsafe impl WalHook for ReplicationLoggerHook {
...
             let runtime = tokio::runtime::Handle::current();
             if let Some(replicator) = ctx.bottomless_replicator.as_mut() {
-                let mut replicator = replicator.lock().unwrap();
-                let _prev = replicator.new_generation();
-                if let Err(e) =
-                    runtime.block_on(async move { replicator.snapshot_main_db_file().await })
-                {
-                    tracing::error!("Failed to snapshot the main db file during checkpoint: {e}");
-                    return SQLITE_IOERR_WRITE;
+                if let Some(replicator) = replicator.lock().unwrap().as_mut() {
```
MarinPostma, line 276: We want to abort the operation if we find out that the replicator is not there anymore, by returning an error. Same in all cases where we check for the replicator, including in `on_frame`

MarinPostma (changes requested):

libsql-server/src/namespace/mod.rs
avinassh: Could you elaborate? Do you simply mean keep some variable like `has_shutdown`, set it to true at shutdown and use this boolean at other places while creating/forking namespace?

MarinPostma: you won't be able to take ownership because you only have a ref, so yes, the first option sounds great :)

avinassh:
@MarinPostma I have addressed the review comments. I have run into a bug. Here is what I did: 

1. Connected to a DB using `turso db shell`
2. I added a sleep statement right after shutdown future is called - https://github.com/tursodatabase/libsql/blob/d07e3f9/libsql-server/src/lib.rs#L447 
3. Shutdown the server and when it was sleeping, I tried to open a new DB, which failed as expected
4. However the old connection (shell opened earlier) still accepted new reads / writes 

How do I prevent this? I would like to close all connections before calling `namespace_store_shutdown_fut`

MarinPostma:
@avinassh in the implementation of `on_frames` in the logger, you want to check that the replicator is still there before doing anything: https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/replication/primary/logger.rs#L81

libsql-server/src/replication/primary/logger.rs
```diff
@@ -84,6 +85,21 @@ unsafe impl WalHook for ReplicationLoggerHook {
...
+            replicator_lock = replicator.lock().unwrap();
+            if replicator_lock.is_none() {
+                tracing::error!("fatal error: expected a replicator, exiting");
+                std::process::abort()
+            } else {
+                replicator_mutex = Option::from(replicator_lock.as_mut().unwrap());
+            }
+        } else {
+            replicator_mutex = None;
+        }
```
avinassh, line 101: in `on_frame`, I wanted to check if bottomless is enabled and if so, hold the lock till the operation is done. Let me know if there is a cleaner way to do above

libsql-server/src/database.rs
```diff
@@ -35,14 +47,26 @@ pub struct PrimaryDatabase {
...
+
+    async fn shutdown(self) -> Result<()> {
         self.logger.closed_signal.send_replace(true);
+        if let Some(replicator) = &self.logger.bottomless_replicator {
+            let replicator = replicator.lock().unwrap().take();
+            if let Some(mut replicator) = replicator {
+                replicator.wait_until_snapshotted().await?;
+            }
+        }
+        Ok(())
```
LucioFranco, line 70: Is this cancel safe? @MarinPostma do we have shutdown timeouts? What happens if this gets canceled? Do we have a half backed up backup? This is just waiting so I assume the actual backup happens on a bg thread that can't get cancelled while doing IO by tokio?

libsql-server/src/lib.rs
```diff
@@ -456,6 +456,7 @@ where
                 namespace_store_shutdown_fut.await?;
                 // clean shutdown, remove sentinel file
                 std::fs::remove_file(sentinel_file_path(&self.path))?;
+                println!("server shutdown complete -- clean exit");
```
MarinPostma, line 459: please use `tracing::info!` for that

libsql-server/src/database.rs
MarinPostma: @Horusiath ☝️

MarinPostma: afaik, there are recovery mecanisms when we detect primary is more up to date than bottomless

libsql-server/src/lib.rs
avinassh: hey yes, I have reverted it to `trace`. I wanted to check if the tests time out, so I pushed something quickly. 

@LucioFranco I think the timeout setting is not working as expected. There is a deadlock in my PR and this particular `test::bottomless::rollback_restore` test never finishes. However, its been more than 10 mins and CI hasn't cancelled it.

libsql-server/src/database.rs
avinassh: > Is this cancel safe? @MarinPostma do we have shutdown timeouts? What happens if this gets canceled? Do we have a half backed up backup? This is just waiting so I assume the actual backup happens on a bg thread that can't get cancelled while doing IO by tokio?

In unusual circumstances, it can get cancelled in the middle of a shutdown.. may be cos of OOM or from Fly. I had tested this scenario earlier, IIRC, when the machine reboots, restoration from the last back up fails. This is easy to test via Nemesis, I will do that and report back.

avinassh:
>  in the implementation of on_frames in the logger, you want to check that the replicator is still there before doing anything

Yes, I did this now. However, the issue was happening because I missed checking `has_shutdown` in `with` / `authenticated` methods and it was initiating new bottomless by using `try_new_primary` 

I have addressed all the review comments. But the one test `test::bottomless::rollback_restore` never finishes. The last log I see is 

```
[2023-11-10T15:08:21Z DEBUG bottomless::replicator] Uploaded to S3: testbackuprestore/ns-testbackuprestore:default-e4e4c070-f7bf-7097-961c-a98e1312c48b/000000001157-000000002000-1699628900.gz in 3.487429ms
[2023-11-10T15:08:22Z TRACE bottomless::replicator] Requesting flush
[2023-11-10T15:08:22Z TRACE bottomless::replicator] Confirmed commit of frame no. 2000 (waited for >= 2000)
[2023-11-10T15:08:22Z DEBUG bottomless::replicator] waiting for generation snapshot e4e4c070-f7bf-7097-961c-a98e1312c48b to complete
```

I am suspecting I introduced a deadlock somewhere. Do you see anything obvious in my PR?

libsql-server/src/namespace/mod.rs
```diff
@@ -348,6 +350,7 @@ struct NamespaceStoreInner<M: MakeNamespace> {
     /// The namespace factory, to create new namespaces.
     make_namespace: M,
     allow_lazy_creation: bool,
+    has_shutdown: RwLock<bool>,
```
LucioFranco, line 353: Why is this a RwLock and not a Mutex? And infact we can probably use AtomicBool instead?

avinassh: > and Ill add to this why is it an async RwLock and not a sync one?

I didn't give much thought to it. The existing code used async lock, since the module had async operations, I used async lock. `has_shutdown`, we can use sync lock/mutex. Should we always default to using sync mutex if that works?

> Why is this a RwLock and not a Mutex? 
The value would be mutated only once, rest all are read operations. So I thought RWLock would be better 

> And infact we can probably use AtomicBool instead?
I guess we can

MarinPostma: yes, an `AtomicBool` is sufficient

libsql-server/src/replication/primary/logger.rs
```diff
@@ -84,6 +85,21 @@ unsafe impl WalHook for ReplicationLoggerHook {
...
         let ctx = Self::wal_extract_ctx(wal);
 
+        let mut binding = ctx.bottomless_replicator.clone();
+        let replicator_mutex: Option<&mut Replicator>;
+        let mut replicator_lock: MutexGuard<Option<Replicator>>;
+        if let Some(replicator) = binding.as_mut() {
+            replicator_lock = replicator.lock().unwrap();
+            if replicator_lock.is_none() {
+                tracing::error!("fatal error: expected a replicator, exiting");
+                std::process::abort()
```
MarinPostma, line 95: do not abort the process here, just return an io error instead

MarinPostma: we just need to prevent the write, no need to shutdown the whole instance

libsql-server/src/replication/primary/logger.rs
```diff
@@ -162,12 +177,16 @@ unsafe impl WalHook for ReplicationLoggerHook {
...
+                if let Some(replicator) = replicator.lock().unwrap().as_mut() {
+                    let prev_valid_frame = replicator.peek_last_valid_frame();
+                    tracing::trace!(
                     "Savepoint: rolling back from frame {prev_valid_frame} to {last_valid_frame}",
                 );
-                replicator.rollback_to_frame(last_valid_frame);
+                    replicator.rollback_to_frame(last_valid_frame);
+                } else {
+                    tracing::error!("fatal error: expected a replicator, exiting");
+                    std::process::abort()
```
MarinPostma, line 188: same here

libsql-server/src/replication/primary/logger.rs
avinassh: Thats a big oops from my side. I had done this at four places and replaced all of them to return `SQLITE_IOERR`

MarinPostma (approved):

PR #503, opened 2023-10-23
libsql-shell: fix panic on empty statement

pjhades:
If the given statement is empty like ` ;`, `run_statement` panicks because in this case rusqlite would return an `Err(SqliteFailure(...))` but we `unwrap` it.

```
$ ./target/debug/libsql
libSQL version 0.2.0
libsql>  ;
thread 'main' panicked at libsql-shell/src/main.rs:407:28:
called `Result::unwrap()` on an `Err` value: SqliteFailure(Error { code: ApiMisuse, extended_code: 21 }, Some("not an error"))
```

This commit fixes the panic with the following changes:
- Prevent the iterator of `StrStatements` from returning empty statements.
- Skip a row if returned by rusqlite if it's `Err(...)`.

With the above changes:

```
$ cargo build && ./target/debug/libsql
    Finished dev [unoptimized + debuginfo] target(s) in 0.49s
;libSQL version 0.2.0
libsql>  ;
libsql>  ;        ;      ;     ;
libsql> ;;;;;;;;;;;;
libsql>    ;;;;;;;;; ;;;;;;;;;
libsql> .quit
```

### DIFF
--- a/libsql-shell/src/main.rs
+++ b/libsql-shell/src/main.rs
@@ -47,7 +47,7 @@ impl Iterator for StrStatements {
                 continue;
             }
             let str_statement = self.value[pos..index + 1].to_string();
-            if str_statement.starts_with(';') || str_statement.is_empty() {
+            if str_statement.trim().starts_with(';') {
                 pos = index + 1;
                 continue;
             }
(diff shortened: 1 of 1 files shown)


MarinPostma (approved):
thanks!

PR #504, opened 2023-10-24
make libsql client a valid target for cloudflare workers

Horusiath:
This PR introduces new feature: `cloudflare-worker`, which enables Hrana protocol to be served over Cloudflare workers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 .vscode/
+.idea/
 *.sqld
--- a/libsql/src/connection.rs
+++ b/libsql/src/connection.rs
@@ -6,7 +6,8 @@ use crate::statement::Statement;
 use crate::transaction::Transaction;
 use crate::{Result, TransactionBehavior};
 
-#[async_trait::async_trait]
+#[cfg_attr(feature = "cloudflare-worker", async_trait::async_trait(?Send))]
+#[cfg_attr(not(feature = "cloudflare-worker"), async_trait::async_trait)]
 pub(crate) trait Conn {
     async fn execute(&self, sql: &str, params: Params) -> Result<u64>;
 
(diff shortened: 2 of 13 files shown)


libsql/src/connection.rs
```diff
@@ -6,7 +6,8 @@ use crate::statement::Statement;
 use crate::transaction::Transaction;
 use crate::{Result, TransactionBehavior};
 
-#[async_trait::async_trait]
+#[cfg_attr(feature = "cloudflare-worker", async_trait::async_trait(?Send))]
+#[cfg_attr(not(feature = "cloudflare-worker"), async_trait::async_trait)]
```
Horusiath, line 10: Any advice on how to make it cleaner? @LucioFranco ?

libsql/src/local/impls.rs
```diff
@@ -15,7 +15,8 @@ pub(crate) struct LibsqlConnection {
     pub(crate) conn: super::Connection,
 }
 
-#[async_trait::async_trait]
+#[cfg_attr(feature = "hrana-cf", async_trait::async_trait(?Send))]
+#[cfg_attr(not(feature = "hrana-cf"), async_trait::async_trait)]
```
LucioFranco, line 19: I don't think this is the right approach we should take since we need these features to be additive.

LucioFranco (changes requested):
No real request changes yet because I don't know exactly what path we should take.

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 .vscode/
+.idea/
```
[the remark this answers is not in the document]
Horusiath: Same could be said about `.vscode/` path ;)

Horusiath:
1.

PR #507, opened 2023-10-25
streaming proxy

MarinPostma:
This PR introduces an extension to the proxy protocol, that uses streams to maintain the connection between the replica and the primary. The two advantages are:
- We don't need to store and garbage collect connections on the primary, the lifetime of the connection is tied to the stream, and when the stream is closed, the link is closed as well.
- We stream the result builder steps, so rows can be streamed from the primary to the replica, without needing to fit them all in ram.

This is an extension to the protocol, for now only sqld implements it. Eventually, libsql should implement it as well.


libsql-server/src/connection/write_proxy.rs
```diff
@@ -264,6 +233,161 @@ impl WriteProxyConnection {
...
 }
 
+struct RemoteConnection<R = Streaming<ExecResp>> {
+    response_stream: R,
+    request_sender: mpsc::Sender<ExecReq>,
+    current_request_id: u32,
+    builder_config: QueryBuilderConfig,
+}
+
+impl RemoteConnection {
```
LucioFranco, line 243: Can we remove the default type and do `RemoteConnection<Streaming<ExecResp>>` instead here?

LucioFranco (approved):

PR #533, opened 2023-10-30
Small bugfix for wrong link to libsql-server

cookiengineer:
Small bugfix for wrong link to libsql-server


MarinPostma (approved):
thanks!!

PR #534, opened 2023-10-30
Add basic namespace stats to /metrics

athoscouto:


MarinPostma (approved):
we should get rid of the stats module and do all that in place I think, but good to go

athoscouto:
Any chance https://github.com/tursodatabase/libsql/pull/534/commits/1cf2d65bc2cc4771f9443b28f890283e48ccf53a breaks de-serialization? Since we're using JSON I figured there is no problem, but I want to double-check with you.

MarinPostma:
Should be OK, i see no reason why those number would ever go negative

athoscouto:
I'm afraid https://github.com/tursodatabase/libsql/pull/534/commits/ed15e00c7a53af98ce05651ce0a133d5267285b8 can possibly cause a memory leak since we never delete the counters.
My understanding is that the [`sql`](https://github.com/tursodatabase/libsql/pull/534/files#diff-003d5c811700d8e4ed177e55297207792a1f6acc68a34d23607bb1f864bcae14R216) variable contains the query template, not the query per see.
So the cardinality would be limited by the number of templates, not the number of queries.
But if the user is not using prepared statements it makes no difference, right?

LucioFranco (approved):
LGTM!

PR #554, opened 2023-11-02
libsql: implement Debug trait for Row

### LINKED ISSUE #283: Can't debug format Row like rusqlite
I can debug row items in a rusqlite query like this:

```rs
let mut stmt = conn.prepare("select * from foo")?;
let mut rows = stmt.query([])?;
while let Some(row) = rows.next()? {
    println!("{:?}", row);
}
```

However, the same code in libsql yields this compiler error:

```
error[E0277]: `libsql::rows::Row` doesn't implement `Debug`
  --> src/libsql/stuff.rs:16:26
   |
16 |         println!("{:?}", row);
   |                          ^^^ `libsql::rows::Row` cannot be formatted using `{:?}` because it doesn't implement `Debug`
   |
   = help: the trait `Debug` is not implemented for `libsql::rows::Row`
   = note: this error originates in the macro `$crate::format_args_nl` which comes from the expansion of the macro `println` (in Nightly builds, run with -Z macro-backtrace for more info)
```

`Row` could use a Debug trait implementation similar to what rusqlite provides.

pjhades:
Resolves #283, following [what rusqlite does](https://github.com/rusqlite/rusqlite/blob/master/src/row.rs#L355).


LucioFranco (approved):
Thanks!

PR #564, opened 2023-11-03
Allow custom heartbeat url for namespace

haaawk:
This is needed for putting namespaces for different users on the same cluster and still correctly reporting the usage.

### DIFF
--- a/libsql-server/src/heartbeat.rs
+++ b/libsql-server/src/heartbeat.rs
@@ -45,19 +45,43 @@ async fn send_stats(
     for (ns, stats) in watched.iter() {
         if let Some(stats) = stats.upgrade() {
             let body = StatsResponse::from(stats.as_ref());
+            let mut heartbeat_url;
             if let Some(url) = url {
-                let mut url = url.clone();
-                url.path_segments_mut().unwrap().push(ns.as_str());
-                let request = client.post(url);
-                let request = if let Some(ref auth) = auth {
-                    request.header("Authorization", auth.to_string())
-                } else {
-                    request
-                };
-                let request = request.json(&body);
-                if let Err(err) = request.send().await {
-                    tracing::warn!("Error sending heartbeat: {}", err);
+                heartbeat_url = url.clone();
+            } else {
+                match namespaces.config_store(ns.clone()).await {
+                    Ok(config_store) => {
+                        let config = config_store.get();
+                        if let Some(url) = config.heartbeat_url.as_ref() {
+                            heartbeat_url = url.clone();
+                        } else {
+                            tracing::warn!(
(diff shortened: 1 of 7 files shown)


libsql-server/src/http/admin/mod.rs
```diff
@@ -244,7 +254,17 @@ async fn handle_fork_namespace<M: MakeNamespace, C>(
...
+    app_state
+        .namespaces
+        .fork(from.clone(), to.clone(), timestamp)
+        .await?;
+    let from_store = app_state.namespaces.config_store(from).await?;
+    let from_config = from_store.get();
+    let to_store = app_state.namespaces.config_store(to).await?;
+    let mut to_config = (*to_store.get()).clone();
+    to_config.max_db_pages = from_config.max_db_pages;
+    to_config.heartbeat_url = from_config.heartbeat_url.clone();
```
MarinPostma, line 266: Is that correct? We now send the heartbeats for the fork to the same URL as the original namespace. How do we disambiguate them?

haaawk: It's fine. Now heartbeat_url contains namespace already. We can do that since now we have url per namespace.

haaawk: Good question. We probably need to take heartbeat_url as a parameter to fork as well. Good catch! Thanks @MarinPostma

libsql-server/src/heartbeat.rs
```diff
@@ -36,16 +37,43 @@ pub async fn server_heartbeat(
...
+                    Err(e) => {
+                        tracing::warn!(
+                            "Error fetching config for namespace {}. Can't send stats: {}",
+                            ns.as_str(),
+                            e
+                        );
+                        continue;
+                    }
+                }
+            }
```
MarinPostma, line 74: nit for followup
```suggestion
            let mut heartbeat_url = if let Some(url) = url {
                url.clone()
            } else {
                match namespaces.config_store(ns.clone()).await {
                    Ok(config_store) => {
                        let config = config_store.get();
                        if let Some(url) = config.heartbeat_url.as_ref() {
                             url.clone()
                        } else {
                            tracing::warn!(
                                "No heartbeat url for namespace {}. Can't send stats!",
                                ns.as_str()
                            );
                            continue;
                        }
                    }
                    Err(e) => {
                        tracing::warn!(
                            "Error fetching config for namespace {}. Can't send stats: {}",
                            ns.as_str(),
                            e
                        );
                        continue;
                    }
                }
            }
```

MarinPostma: actually a `match` would look even nicer here

MarinPostma (approved):

PR #574, opened 2023-11-06
fix bug in lock-stealing

MarinPostma:
This PR fixes a bug in lock stealing that was two-fold:

- When downgrading a transaction we took the state slot without making sure that the state lock was ours, so we could end up taking some other txn slot instead, leaving an empty slot.
- When stealing a txn, we didn't make sure that the slot we stole was the same slot as the one we intended to steal.

This fix was tested by running the following code for 1M iterations:
<details>
  <summary>Click me</summary>
  
  ```rust
#![allow(dead_code)]

use rand::Rng;
use serde::{ Deserialize,Serialize };
use tokio::task::JoinSet;

type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;


impl From<&[&str]> for Req {
    fn from(s: &[&str]) -> Self {
        let requests = s.into_iter().map(|&s| Op::Execute { stmt: Sql { sql: s.into() } }).collect();
        Req {
            baton: None,
            requests,
        }
    }
}

#[derive(Serialize)]
struct Req {
    baton: Option<String>,
    requests: Vec<Op>,
}

#[derive(Serialize)]
#[serde(tag = "type", rename_all = "camelCase")]
enum Op {
    Execute {
        stmt: Sql,
    },
    Close(CloseReq),
}

#[derive(Serialize)]
struct CloseReq {}

#[derive(Serialize)]
struct Sql {
    sql: String,
}


#[derive(Deserialize, Debug)]
#[serde(untagged)]
enum RespOrError {
    Resp {
        #[serde(flatten)]
        resp: Resp
    },
    Error {
        error: String,
    }
}

#[derive(Deserialize, Debug)]
struct Resp {
    baton: Option<String>,
    base_url: Option<String>,
    results: Vec<RespResult>,
}

#[derive(Deserialize, Debug)]
#[serde(tag = "type", rename_all = "camelCase")]
enum RespResult {
    Ok {
        response: OpResult,
    }
}

#[derive(Deserialize, Debug)]
#[serde(tag = "type", rename_all = "camelCase")]
enum OpResult {
    Execute {
        result: ExecuteResult,
    },
    Close,
}

#[derive(Deserialize, Debug)]
struct ExecuteResult {
    cols: Vec<Col>,
    rows: Vec<Row>,
    affected_row_count: u64,
    last_inserted_row_id: Option<i64>,
    replication_index: u64,
}

#[derive(Deserialize, Debug)]
struct Col {
    name: String,
    decltype: Option<String>,
}

type Row = ();

async fn send_request(s: &[&str], url: &str, baton: Option<String>, close: bool, client: &reqwest::Client) -> Result<Resp> {
    let mut req = Req::from(s);
    req.baton = baton;

    if close {
        req.requests.push(Op::Close(CloseReq {  }));
    }
    let resp = client.post(url).json(&req).send().await?;
    

    let body = resp.text().await?;
    // pcintln!("{body}\n");
    let json = serde_json::from_str::<RespOrError>(&body)?;
    match json {
        RespOrError::Resp { resp } => Ok(resp),
        RespOrError::Error { error } => {
            println!("{error}");
            panic!();
        }
    }
}

async fn make_txn(host: &str, client: &reqwest::Client) -> Result<()> {
    let ret = send_request(&["begin immediate", "insert into test values (11)"], host, None, false, client).await?;
    let baton = ret.baton.unwrap();
    let rand = rand::thread_rng().gen_range(0..100);
    if rand > 1 {
        send_request(&["insert into test values (11)", "commit"], host, Some(baton), true, client).await?;
    }

    Ok(())
}

#[tokio::main]
async fn main() -> Result<()> {
    let host = "http://localhost:8080/v2/pipeline";
    let mut join_set = JoinSet::new();
    let (send, mut rcv) = tokio::sync::mpsc::unbounded_channel::<()>();
    join_set.spawn(async move {
        let mut count = 0;
        while let Some(_) = rcv.recv().await {
            count += 1;

            print!("\rrequest count: {count:>8}");
            
        }

    });

    for _ in 0..3 {
        join_set.spawn({
            let snd = send.clone();
            async move {
                let client = reqwest::Client::new();
                loop  {
                    make_txn(host, &client).await.unwrap();
                    let _ = snd.send(());
                }
            } 
        });
    } 

    join_set.join_next().await.unwrap().unwrap();

    Ok(())
}
```
</details>

### DIFF
--- a/libsql-server/src/connection/libsql.rs
+++ b/libsql-server/src/connection/libsql.rs
@@ -938,10 +938,12 @@ where
 #[cfg(test)]
 mod test {
     use itertools::Itertools;
+    use rand::Rng;
     use sqld_libsql_bindings::wal_hook::TRANSPARENT_METHODS;
     use tempfile::tempdir;
     use tokio::task::JoinSet;
 
+    use crate::connection::Connection as _;
     use crate::query_result_builder::test::{test_driver, TestBuilder};
     use crate::query_result_builder::QueryResultBuilder;
     use crate::DEFAULT_AUTO_CHECKPOINT;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: libsql-server/src/connection/libsql.rs [hunk 1]
```diff
@@ -389,24 +389,32 @@ unsafe extern "C" fn busy_handler<W: WalHook>(state: *mut c_void, _retries: c_in
...
-                        // steal.
-                        assert!(lock.take().is_some());
-
-                        slot.abort();
-                        tracing::info!("stole transaction lock");
+                    if let Some(ref s) = *lock {
+                        // The state contains the same lock as the one we're attempting to steal
+                        if Arc::ptr_eq(s, &slot) {
+                            // We check that slot wasn't already stolen, and that their is still a slot.
+                            // The ordering is relaxed because the atomic is only set under the slot lock.
```
LucioFranco, line 399: Tracks 👍

libsql-server/src/connection/libsql.rs
```diff
@@ -537,8 +545,20 @@ impl<W: WalHook> Connection<W> {
                 }
                 // lock was downgraded, notify a waiter
                 (Tx::Write, Tx::None | Tx::Read) => {
-                    state.slot.write().take();
-                    lock.slot.take();
+                    let old_slot = lock
```
LucioFranco, line 548: I wonder what are the cases here of a race? I don't know the full picture so hard to say but how can we ensure this is true?

LucioFranco (approved):
Left one question, but overall lgtm

libsql-server/src/connection/libsql.rs
MarinPostma: The slot could have been stolen after we had executed the statement but before we updated the slot in the state. In this case, the stealing txn would rollback our txn, but that's fine, because we already committed at that point, so it's no-op. In the meantime, the stealer would have put their slot in the state, which we would take here

libsql-server/src/connection/libsql.rs
```diff
@@ -389,24 +389,32 @@ unsafe extern "C" fn busy_handler<W: WalHook>(state: *mut c_void, _retries: c_in
...
-                        // detects a timeout, so if we arrive to this point, then there is
-                        // necessarily a slot, and this slot has to be the one we attempted to
-                        // steal.
-                        assert!(lock.take().is_some());
-
-                        slot.abort();
-                        tracing::info!("stole transaction lock");
+                    if let Some(ref s) = *lock {
+                        // The state contains the same lock as the one we're attempting to steal
+                        if Arc::ptr_eq(s, &slot) {
```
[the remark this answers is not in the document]
MarinPostma: We're comparing 2 `Arc`s so we need to have them at hand. This means that the two arcs have a ref count of at least 1 when we compare them, so they could not have been deallocated, hence not reallocated with the same address. But that's a great remark I didn't think about that!

PR #587, opened 2023-11-08
Add client version reporting metrics

### LINKED ISSUE #546: Client SDK and version telemetry
Let's add a "user agent" header to the libSQL remote protocol in the SDKs and track that on the libSQL server side (perhaps exposing prometheus metrics). This improves observability when debugging production issues: what SDKs and what versions of them are used, in particular when there are errors happening.

LucioFranco:
This commit adds a new `x-libsql-client-version` header emitted by clients. The server will collect these values and emit them as a `libsql_client_version{version="libsql-hrana-0.1.11"}`. This also exposes special doc hidden functions that our other clients that use the rust one to emit their own metric.

Closes #546

### DIFF
--- a/libsql-server/src/rpc/mod.rs
+++ b/libsql-server/src/rpc/mod.rs
@@ -7,8 +7,12 @@ use rustls::server::AllowAnyAuthenticatedClient;
 use rustls::RootCertStore;
 use tonic::Status;
 use tower::util::option_layer;
+use tower::ServiceBuilder;
+use tower_http::trace::DefaultOnResponse;
+use tracing::Span;
 
 use crate::config::TlsConfig;
+use crate::metrics::CLIENT_VERSION;
 use crate::namespace::{NamespaceName, NamespaceStore, PrimaryNamespaceMaker};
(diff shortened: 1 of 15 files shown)


libsql-server/src/rpc/mod.rs
```diff
@@ -122,3 +151,20 @@ fn extract_namespace<T>(
...
+        req.uri(),
+        req.headers()
+    );
+
+    if let Some(v) = req.headers().get("x-libsql-client-version") {
+        if let Ok(s) = v.to_str() {
+            metrics::increment_counter!(CLIENT_VERSION, "version" => s.to_string());
+        }
+    }
+}
```
MarinPostma, line 170: can't we just reuse this function?

MarinPostma (approved):

libsql-server/src/rpc/mod.rs
LucioFranco: yeah, we can in a follow up but turns out this was for internal where we don't send any versions so this won't even get called. We prob wanna remove this.

PR #601, opened 2023-11-09
make CI faster, more timeouts sooner

LucioFranco:
- Merges make wasm and maketest into a single sqlite3 test and only runs if we change any files in the `libsql-sqlite3` directory.
- Removes needing to build sqlite3 from scratch before running cargo build in the go tests, we don't actually use any of the stuff generated by the make command since we manually copy the amalgamated C file into our rust code.
- Added some config for `cargo nextest` to catch timeout tests so we don't get 3h+ CI jobs


.github/workflows/sqlite3.yml
```diff
@@ -39,12 +44,3 @@ jobs:
...
       run: make rusttestwasm
-
-    - name: reconfigure with Wasm in WasmEdge mode
-      run: make clean && ./configure --enable-wasm-runtime-wasmedge
-
-    - name: Run tests
-      run: make test
-
-    - name: Run tests with Wasm
-      run: make rusttestwasm
```
MarinPostma, line 50: why is this going?

.github/workflows/maketest.yml
```diff
@@ -1,41 +0,0 @@
...
-    - name: configure
-      run: ./configure
-      
-    - name: Run tests
-      run: make test
-
-    # Rust tests are run under maketestwasm
-
-    - name: Run API tests
-      run: make testlibsql
```
MarinPostma, line 41: why is this going as well?

.github/workflows/sqlite3.yml
LucioFranco: I talked with @psarna this morning, we don't need wasmedge right now and the first pass covers enough for all our C code. We were duplicating a lot of work.

.github/workflows/maketest.yml
LucioFranco: Yup as I said above, a lot of our make tests did a lot of the same so if we just do it with + wasm then it covers basically everything. This was after a chat with @psarna.

PR #637, opened 2023-11-16
update metadata for releasing w/project renames

ashleygwilliams:


bottomless-cli/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.14"
 edition = "2021"
 license = "MIT"
 keywords = ["libsql", "sqlite", "s3", "cli", "replication"]
-repository = "https://github.com/libsql/sqld"
+repository = "https://github.com/libsql/libsql"
```
LucioFranco, line 7: Well get this fixed eventually as well as the org is wrong now too, but well do that.

ashleygwilliams: fixed while i was waiting, heh

MarinPostma (commented):
I don't see the old insta snapshot files being removed, I think the rename left them behind?

ashleygwilliams:
oh huh @MarinPostma that looks correct actually. i'll clean that up. additionally, there's a test i can't run locally due to a lack of a runtime dep i think- so that's the origin of the test failure i believe. will try to sort that but welcome any pointers!

ashleygwilliams:
(i'll also manage the conflict)

ashleygwilliams:
(hopping in a call now but will fix my rebase botch when i'm out)

MarinPostma:
Thanks for the effort @ashleygwilliams 😄

PR #645, opened 2023-11-17
Add nemesis test runner

avinassh:
This PR adds a Github Workflow to run [Nemesis tests](https://github.com/tursodatabase/tursotest/tree/main/internal/nemesis) in local mode. In local mode, it spawns multiple libsql-server and minio processes and runs the test suite. 

I have added Nemesis test workflow separately than `Run Tests`. There is a possibility of false positive alert (the code is new) and separate work flow makes it easier to distinguish.

### DIFF
--- a/.github/workflows/nemesis.yml
+++ b/.github/workflows/nemesis.yml
@@ -1,0 +1,80 @@
+name: Rust
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+  merge_group:
+    branches: [ "main" ]
+
+env:
+  CARGO_TERM_COLOR: always
+  PROTOC_VERSION: 3.23.4
+  # FIXME: There are some warnings depending on certain feature flags that
+  # we need to fix before we can enable this.
+  # RUSTFLAGS: "-D warnings"
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/nemesis.yml
```diff
@@ -0,0 +1,76 @@
+name: Rust
+
+on:
+  push:
```
MarinPostma, line 4: we probably want to be more specific on what events this is triggered. Do we also want that to be a mandated workflow for merging? maybe we should

avinassh: good catch. WDYT doing same as Rust Tests: 

```
  push:
    branches: [ "main" ]
  pull_request:
  merge_group:
    branches: [ "main" ]
```

avinassh: >  Do we also want that to be a mandated workflow for merging? maybe we should

We should.

MarinPostma: sounds good to me :)

avinassh: I have updated the PR 🫡

PR #653, opened 2023-11-19
libsql: remote hrana transactions

Horusiath:
This PR introduces transactions in libsql client using http+json. It introduces a Hrana concept of HttpStream, which is opened for sequential execution of subsequent requests.


libsql/src/hrana/stream.rs
```diff
@@ -0,0 +1,375 @@
...
+            .client
+            .http_send(self.base_url.clone(), self.auth_token.clone(), body)
+            .await?;
+        if let Some(base_url) = response.base_url.take() {
+            self.base_url = base_url;
+        }
+        match response.baton.take() {
+            None => {
+                tracing::trace!("client stream has been closed by the server");
+                self.status = StreamStatus::Closed
```
MarinPostma, line 254: why not call `self.done()` here?

libsql/src/hrana/stream.rs
```diff
@@ -0,0 +1,375 @@
...
+                response.results
+            )))?;
+        }
+        if response.results.len() > N {
+            // One with actual results, one closing the stream
+            Err(HranaError::UnexpectedResponse(format!(
+                "Unexpected multiple responses from server: {:?}",
+                response.results
+            )))?;
+        }
```
MarinPostma, line 274: what if `results.len < N`, we would panic at 277

libsql/src/hrana/stream.rs
```diff
@@ -0,0 +1,375 @@
...
+        let resp = client
+            .send(StreamRequest::StoreSql(StoreSqlStreamReq { sql, sql_id }))
+            .await?;
+        match resp {
+            StreamResponse::StoreSql => {
+                drop(client);
+                Ok(StoredSql::new(self.clone(), sql_id))
+            }
+            other => {
+                client.sql_id_generator = client.sql_id_generator.wrapping_sub(1);
```
MarinPostma, line 120: I'm not entirely sure about that, can we trust that the server hasn't stored the statement, despite returning an error? Wouldn't it be safe to just discard this sql_id altogether?

libsql/src/hrana/pipeline.rs
```diff
@@ -20,25 +20,47 @@ pub struct ServerMsg {
...
+#[derive(Serialize, Debug)]
+pub struct StoreSqlStreamReq {
+    pub sql: String,
+    pub sql_id: i32,
+}
+
+#[derive(Serialize, Debug)]
+pub struct CloseSqlStreamReq {
+    pub sql_id: i32,
+}
```
MarinPostma, line 63: we really need to factorize the proto types with libsql-server

MarinPostma (commented):
Reading the code for transactions I'm starting to wonder if it's a good idea to have connection's `execute` take `&self`, and be Sync and Send. If you're sharing a connection between threads, you're likely doing something wrong.

On the surface it looks fine from the remote perspective, because a new sqlite connection is created by the server for each execute request, but locally, the connection is backed by the same sqlite connection, so there is an inadequation in the semantics I think. I don't think connection should be shared, and if you really need to, I think the connection itself should be wrapped into a Mutex to synchronize accesses.

The transaction code made that even more evident to me, since we need a mutex around the HTTP stream to serialize requests, when it's clearly a code smell to have sharable access to a running transaction.

libsql/src/hrana/transaction.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    pub async fn open(stream: HttpStream<T>, tx_behavior: TransactionBehavior) -> Result<Self> {
+        let begin_stmt = match tx_behavior {
+            TransactionBehavior::Deferred => "BEGIN DEFERRED",
+            TransactionBehavior::Immediate => "BEGIN IMMEDIATE",
+            TransactionBehavior::Exclusive => "BEGIN EXCLUSIVE",
+            TransactionBehavior::ReadOnly => "BEGIN READONLY",
+        };
+        stream.execute(Stmt::new(begin_stmt, false)).await?;
+        Ok(HttpTransaction { stream })
```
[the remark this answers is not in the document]
Horusiath: On one side yes. On the other - this may screw happen before relationships as opening the transaction won't actually start a new transaction ie.:

1. Have a database in state S1.
2. Open transaction T1 (current db state: S1).
3. Open transaction T2.
4. Within T2, modify database state S1&rarr;S2.
5. Read something in T1: since this is first time when actual request is being made - T1 will point to S2.

libsql/src/hrana/hyper.rs
```diff
@@ -166,3 +184,79 @@ impl crate::statement::Stmt for crate::hrana::Statement<HttpSender> {
...
+    async fn transaction(
+        &self,
+        _tx_behavior: crate::TransactionBehavior,
+    ) -> crate::Result<crate::transaction::Transaction> {
+        todo!("sounds like nested transactions innit?")
+    }
+
+    fn is_autocommit(&self) -> bool {
+        // TODO: Is this correct?
+        false
```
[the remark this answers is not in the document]
Horusiath: I didn't want to change the existing API (is_autocomit is not async), but on the other side when transaction is opened, we could attach GetAutocommitReq to it.

libsql/src/hrana/transaction.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[derive(Debug, Clone)]
+pub(crate) struct HttpTransaction<T>
+where
+    T: for<'a> HttpSend<'a>,
+{
+    stream: HttpStream<T>,
+}
+
+impl<T> HttpTransaction<T>
```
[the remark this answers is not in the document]
Horusiath: Yes, but that implies sending rollback request in non-async method context, which can also fail (and possibly deadlock in some cases). IMO rollback on drop should be implemented on server side - and probably dependent on is_autocommit behaviour as well.

libsql/src/hrana/transaction.rs
```diff
@@ -0,0 +1,64 @@
...
+        self.stream.execute(stmt).await
+    }
+
+    pub async fn execute_batch(
+        &self,
+        stmts: impl IntoIterator<Item = Stmt>,
+    ) -> Result<BatchResult> {
+        let mut batch = Batch::new();
+        for stmt in stmts.into_iter() {
+            batch.step(None, stmt);
```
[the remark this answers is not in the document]
Horusiath: This is how it's realised in libsql-client-rs. I'm also using libsql-client-ts for reference, and have problems with that implementation: namely every batch starts with `BEGIN ...` statement, yet `END`/`COMMIT`/`ROLLBACK` are not attached to batch.

libsql/src/hrana/connection.rs
```diff
@@ -72,162 +64,83 @@ where
...
+
+pub(crate) enum CommitBehavior {
+    Commit,
+    Rollback,
+}
+
+pub(super) fn stmts_to_batch(protocol_v3: bool, stmts: impl IntoIterator<Item = Stmt>) -> Batch {
+    let mut batch = Batch::new();
+    let mut step = -1;
+    for stmt in stmts.into_iter() {
```
LucioFranco, line 130: could use https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.enumerate maybe here to do the counter for you so you don't need to do that logic manually.

libsql/src/hrana/hyper.rs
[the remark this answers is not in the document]
LucioFranco: @Horusiath just for note, at this stage if we need to make it async its also an ok option.

libsql/src/hrana/hyper.rs
```diff
@@ -166,3 +184,79 @@ impl crate::statement::Stmt for crate::hrana::Statement<HttpSender> {
...
+
+    fn last_insert_rowid(&self) -> i64 {
+        self.last_insert_rowid()
+    }
+
+    fn close(&mut self) {
+        //TODO: nobody is using this trait method and since it's not async
+        // there's no reason to do it anyway
+        todo!()
+    }
```
LucioFranco, line 261: We might want to consider removing this if its not needed, we do have drop that we can use to clean things up.

libsql/src/hrana/mod.rs
```diff
@@ -51,18 +53,50 @@ pub enum HranaError {
     Api(String),
 }
 
-pub struct Statement<T> {
-    client: HttpConnection<T>,
+enum StatementExecutor<T: for<'a> HttpSend<'a>> {
```
LucioFranco, line 56: Its possible to remove the T: HttpSend bound here as well and it should clean up some of the bounds.

libsql/src/hrana/stream.rs
```diff
@@ -0,0 +1,372 @@
+use crate::hrana::pipeline::{
+    BatchStreamReq, ClientMsg, CloseSqlStreamReq, DescribeStreamReq, ExecuteStreamReq, Response,
+    SequenceStreamReq, ServerMsg, StoreSqlStreamReq, StreamRequest, StreamResponse,
+    StreamResponseOk,
+};
+use crate::hrana::proto::{Batch, BatchResult, DescribeResult, Stmt, StmtResult};
+use crate::hrana::{HranaError, HttpSend, Result};
+use futures::lock::Mutex;
```
LucioFranco, line 8: We should use the tokio mutex if we need an async mutex and not the one inside futures crate. Its a bit unsafe and not as fully tested as the tokio one which I trust much more.

libsql/src/hrana/stream.rs
```diff
@@ -0,0 +1,372 @@
...
+}
+
+pub type SqlId = i32;
+
+/// A representation of Hrana HTTP stream. Since streams rely on sequential execution of requests,
+/// it's realized internally as a mutex lock.
+#[derive(Debug)]
+pub struct HttpStream<T>
+where
+    T: for<'a> HttpSend<'a>,
```
LucioFranco, line 28: Can also remove the bound here to clean this up.

libsql/src/hrana/stream.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    pub async fn store_sql(&self, sql: String) -> Result<StoredSql<T>> {
+        let mut client = self.inner.stream.lock().await;
+        let sql_id = client.next_sql_id();
+        let resp = client
+            .send(StreamRequest::StoreSql(StoreSqlStreamReq { sql, sql_id }))
+            .await?;
+        match resp {
+            StreamResponse::StoreSql => {
+                drop(client);
```
LucioFranco, line 116: Why do we drop explicitly here rather than let it implicitly drop. Was there some issue? If so we should document that.

libsql/src/hrana/stream.rs
```diff
@@ -0,0 +1,372 @@
...
+    pub async fn execute(&self) -> Result<()> {
+        self.stream.sequence(self).await
+    }
+
+    pub async fn close(self) -> Result<()> {
+        self.stream.close_sql(self.sql_id).await
+    }
+}
+
+pub trait SqlDescriptor {
```
LucioFranco, line 349: Could you add a comment explaining this abstraction?

libsql/src/hrana/stream.rs
```diff
@@ -0,0 +1,372 @@
...
+{
+    fn sql_description(&self) -> SqlDescription {
+        SqlDescription::SqlId(self.sql_id)
+    }
+}
+
+#[derive(Debug)]
+pub enum SqlDescription {
+    Sql(String),
+    SqlId(SqlId),
```
LucioFranco, line 371: What about String and Remote? Instead of these two names that seem quite similar?

libsql/src/hrana/transaction.rs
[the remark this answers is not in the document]
LucioFranco: Can we tokio::spawn and rollback? So its done async to the drop and that shouldn't cause any deadlocks?

libsql/src/hrana/transaction.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub fn stream(&self) -> &HttpStream<T> {
+        &self.stream
+    }
+
+    pub async fn open(stream: HttpStream<T>, tx_behavior: TransactionBehavior) -> Result<Self> {
+        let begin_stmt = match tx_behavior {
+            TransactionBehavior::Deferred => "BEGIN DEFERRED",
+            TransactionBehavior::Immediate => "BEGIN IMMEDIATE",
+            TransactionBehavior::Exclusive => "BEGIN EXCLUSIVE",
+            TransactionBehavior::ReadOnly => "BEGIN READONLY",
```
LucioFranco, line 29: We probably need to map this to `DEFERRED` since I don't think the server supports this syntax. We should make sure we test this.

libsql/src/wasm/mod.rs
```diff
@@ -73,15 +77,84 @@ where
...
+{
+    inner: HttpTransaction<T>,
+}
+
+impl<T> Transaction<T>
+where
+    T: for<'a> HttpSend<'a>,
+{
+    pub async fn query(&self, sql: &str, params: impl IntoParams) -> crate::Result<Rows> {
+        tracing::trace!("querying `{}`", sql);
```
LucioFranco, line 118: nit: could prefix the logs here like so to make it easier to debug in the future.
```suggestion
        tracing::trace!("txn querying `{}`", sql);
```

LucioFranco (changes requested):
Overall, really nice work! I left a few comments and some questions.

libsql/src/hrana/transaction.rs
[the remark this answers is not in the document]
Horusiath: Can we be sure that tokio task we complete if ie. program ends right after transaction drop?

libsql/src/hrana/connection.rs
```diff
@@ -1,29 +1,22 @@
...
-    baton: Option<String>,
-    base_url: Option<String>,
-}
+use std::sync::Arc;
 
 #[derive(Debug)]
-pub struct HttpConnection<T>(Arc<InnerClient<T>>);
+pub struct HttpConnection<T>(Arc<InnerClient<T>>)
+where
+    T: for<'a> HttpSend<'a>;
```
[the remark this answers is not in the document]
Horusiath: I didn't defined it because I wanted it to be fancy. I defined it this way, because it was the only way I've found to make borrow checker cooperate. If there's a better way I'm open for suggestions.

libsql/src/hrana/stream.rs
Horusiath: I named them this way, since they map directly onto Hrana protocol fields.

libsql/src/hrana/transaction.rs
Horusiath: I was suspecting that since replication client supports this (this code is copied from there), it should be supported in hrana as well.

libsql/src/hrana/hyper.rs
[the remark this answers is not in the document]
Horusiath: Ok, I made `is_autocommit` into async.

libsql/src/hrana/hyper.rs
Horusiath: Removed. We'd be better to rely on Drop.

libsql/src/hrana/transaction.rs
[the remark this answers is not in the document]
Horusiath: I've introduced `Drop for libsql::Transaction` together with optional close callback invoked there. It's basically used to schedule rollback task on HranaTransaction. That was the best option I could think of, given that HranaTransaction may be running in context where `tokio::spawn` is not available, and `futures::block_on` basically deadlocks the tests (or single thread runtime).

I haven't figured yet how to rollback on transaction drop in Cloudflare - it has a context method that can be used to schedule async method instead of using tokio, but the access pattern to it is very invasive in the API.

Horusiath:
If you don't mind guys, unless you have some strong opinions and must-fix remarks about existing code, we could resolve remaining questions in subsequent PRs. I don't think it's good to have this PR rot. I've created a #723 for topics opened but not covered here.

libsql/src/hrana/connection.rs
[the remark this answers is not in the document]
LucioFranco: What I mean is that if your struct foo<T> doesn't use the trait inside of the struct def you don't need to bound T by the trait. You can just leave it as T. This will then flow throughout the rest of the code since you no longer need to bound T: Trait for each struct that embeds the struct that contains T. 

Trait bounds should only be set at the impl of where the function is used. The only time you want to bound the struct generic is if you are using an associated type with in the struct.

libsql/src/hrana/transaction.rs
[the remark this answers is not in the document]
LucioFranco: @Horusiath I found this https://github.com/cloudflare/workers-rs/blob/main/examples/tokio-postgres/src/lib.rs#L19 I wonder if we can use this function to spawn in wasm?

libsql/src/hrana/transaction.rs
LucioFranco: No clue, we should test that it actually works I have a feeling it might not but I could also be wrong.

LucioFranco (approved):
I am approving, I added a few more items in the follow up issue.

PR #656, opened 2023-11-20
libsql: port row deserializer from rust client

### LINKED ISSUE #614: Port de.rs from libsql-client-rs

pjhades:
This PR ports [de.rs](https://github.com/libsql/libsql-client-rs/blob/main/src/de.rs) from libsql-client-rs so that `Row`s can be deserialized, attempting to resolve #614.

I'm adding a `column_count` method to the `RowInner` trait because I find it convenient for telling when to stop the iteration on the `Row` inside the `MapAccess` implementation. Alternatively, we can maintain a column index and stop the iteration as soon as `row.column_value` or `row.column_name` returns `Err` or `None`. (That would require some changes on the existing `RowInner` implementations.)

### DIFF
--- a/libsql/src/lib.rs
+++ b/libsql/src/lib.rs
@@ -87,6 +87,9 @@ mod statement;
 mod transaction;
 mod value;
 
+#[cfg(feature = "serde")]
+pub mod deserialize_row;
+
 pub use value::{Value, ValueRef, ValueType};
 
 cfg_hrana! {
--- a/libsql/src/deserialize_row.rs
+++ b/libsql/src/deserialize_row.rs
@@ -1,0 +1,99 @@
+use crate::{Row, Value};
+use serde::de::{
+    value::{Error as DeError, SeqDeserializer},
+    Error, IntoDeserializer, MapAccess, Visitor,
+};
+use serde::{Deserialize, Deserializer};
+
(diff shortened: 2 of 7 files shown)


libsql/src/de.rs
```diff
@@ -0,0 +1,99 @@
+use crate::{Row, Value};
+use serde::de::{
+    value::{Error as DeError, SeqDeserializer},
+    Error, IntoDeserializer, MapAccess, Visitor,
+};
+use serde::{Deserialize, Deserializer};
+
+struct De<'de> {
+    row: &'de Row,
+}
```
MarinPostma, line 10: Let's rename that to `RowDeserializer`

libsql/src/lib.rs
```diff
@@ -87,6 +87,9 @@ mod statement;
 mod transaction;
 mod value;
 
+#[cfg(feature = "serde")]
+pub mod de;
```
MarinPostma, line 91: let's rename that to `deserialize_row`

MarinPostma (changes requested):
Thanks for helping with that! A few details I would like addressed, and I'll let @LucioFranco give his approval

MarinPostma (approved):

PR #699, opened 2023-11-25
Fix remote connection panic

wyhaya:
```rust
let conn = Database::open_remote("", "").unwrap().connect().unwrap();
// !!! Painc
conn.query("SELECT 1", ()).await;
```
```
thread 'main' panicked at libsql/src/hrana/hyper.rs:40:14:
called `Result::unwrap()` on an `Err` value: http::Error(InvalidUri(InvalidFormat))
```

### DIFF
--- a/libsql/src/hrana/hyper.rs
+++ b/libsql/src/hrana/hyper.rs
@@ -37,23 +37,19 @@ impl HttpSender {
             .header(AUTHORIZATION, auth)
             .header("x-libsql-client-version", self.version.clone())
             .body(hyper::Body::from(body))
-            .unwrap();
+            .map_err(|err| HranaError::Http(format!("{:?}", err)))?;
 
-        let res = self.inner.request(req).await.map_err(HranaError::from)?;
+        let res = self.inner.request(req).await?;
 
         if res.status() != StatusCode::OK {
-            let body = hyper::body::to_bytes(res.into_body())
-                .await
-                .map_err(HranaError::from)?;
-            let body = String::from_utf8(body.into()).unwrap();
-            return Err(HranaError::Api(body));
+            let body = hyper::body::to_bytes(res.into_body()).await?;
+            let msg = String::from_utf8(body.into()).unwrap_or_else(|err| err.to_string());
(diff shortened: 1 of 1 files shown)


libsql/src/hrana/hyper.rs
```diff
@@ -37,23 +37,19 @@ impl HttpSender {
             .header(AUTHORIZATION, auth)
             .header("x-libsql-client-version", self.version.clone())
             .body(hyper::Body::from(body))
-            .unwrap();
+            .map_err(|err| HranaError::Http(format!("{:?}", err)))?;
```
wyhaya, line 40: I'm using a `Debug` value here instead of a `Display` value. Its display value is `invalid format`, which is very difficult to troubleshoot. However, the debug value is `http::Error(InvalidUri(InvalidFormat))`, which is clear enough.

libsql/src/hrana/hyper.rs
```diff
@@ -37,23 +37,19 @@ impl HttpSender {
...
+        let res = self.inner.request(req).await?;
 
         if res.status() != StatusCode::OK {
-            let body = hyper::body::to_bytes(res.into_body())
-                .await
-                .map_err(HranaError::from)?;
-            let body = String::from_utf8(body.into()).unwrap();
-            return Err(HranaError::Api(body));
+            let body = hyper::body::to_bytes(res.into_body()).await?;
+            let msg = String::from_utf8(body.into()).unwrap_or_else(|err| err.to_string());
```
MarinPostma, line 46: maybe a bit more context could be added here to explain that the string error comes from an invalid payload

MarinPostma (changes requested):

libsql/src/hrana/hyper.rs
wyhaya: After change: HranaError::Api(`Invalid payload: invalid utf-8 sequence of 1 bytes from index 1`)

MarinPostma:
looks good to me thanks!

MarinPostma (approved):

PR #706, opened 2023-11-27
fix(release): remove musl target

ashleygwilliams:


MarinPostma:
thanks for fixing this!

PR #707, opened 2023-11-27
libsql: fix for rust-cache

mistydemeo:
The rust-cache GitHub action uses a TOML parser which chokes on the extra space before the comma in one line of this `Cargo.toml`. While it's actually valid syntax, just removing the extra comma should fix running the job.


MarinPostma:
thanks for finding this one! :)

 @LucioFranco @psarna pinging your guys so you're aware

PR #714, opened 2023-11-28
bindings/c: Implement libsql_open_{file,remote} API

penberg:
The libsql_open_ext() API does not support auth token. Let's add a libsql_open_remote() API for establishing connection to remote database, but also add a libsql_open_file() for what libsql_open_ext() does so that we can deprecate it in the future.

### DIFF
--- a/bindings/c/include/libsql.h
+++ b/bindings/c/include/libsql.h
@@ -42,6 +42,10 @@ int libsql_open_sync(const char *db_path,
 
 int libsql_open_ext(const char *url, libsql_database_t *out_db, const char **out_err_msg);
 
+int libsql_open_file(const char *url, libsql_database_t *out_db, const char **out_err_msg);
+
+int libsql_open_remote(const char *url, const char *auth_token, libsql_database_t *out_db, const char **out_err_msg);
+
 void libsql_close(libsql_database_t db);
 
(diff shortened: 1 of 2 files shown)


haaawk:
This breaks the compilation of golang bindings:
```
error[E0425]: cannot find value `url` in this scope
   --> bindings/c/src/lib.rs:104:22
    |
104 |     libsql_open_file(url, out_db, out_err_msg)
    |                      ^^^ not found in this scope

error[E0425]: cannot find value `url` in this scope
   --> bindings/c/src/lib.rs:113:49
    |
113 |     let url = unsafe { std::ffi::CStr::from_ptr(url) };
    |                                                 ^^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
error: could not compile `sql-experimental` (lib) due to 2 previous errors
```

and C bindings:

```
error[E0425]: cannot find value `url` in this scope
   --> bindings/c/src/lib.rs:104:22
    |
104 |     libsql_open_file(url, out_db, out_err_msg)
    |                      ^^^ not found in this scope

error[E0425]: cannot find value `url` in this scope
   --> bindings/c/src/lib.rs:113:49
    |
113 |     let url = unsafe { std::ffi::CStr::from_ptr(url) };
    |                                                 ^^^ not found in this scope
```

tvandinther:
It appears that the offending lines require a variable rename of `url` -> `path` :blush:

PR #719, opened 2023-11-28
Make snapshot at shutdown opt in

avinassh:


libsql-server/src/lib.rs
```diff
@@ -101,6 +101,7 @@ pub struct Server<C = HttpConnector, A = AddrIncoming, D = HttpsConnector<HttpCo
     pub heartbeat_config: Option<HeartbeatConfig>,
     pub disable_namespaces: bool,
     pub shutdown: Arc<Notify>,
+    pub snapshot_at_shutdown: bool,
```
MarinPostma, line 104: can you move that to the `DbConfig` struct?

MarinPostma (changes requested):

libsql-server/src/namespace/mod.rs
```diff
@@ -645,9 +647,11 @@ impl<T: Database> Namespace<T> {
...
     }
 
-    async fn shutdown(mut self) -> anyhow::Result<()> {
+    async fn shutdown(mut self, should_checkpoint: bool) -> anyhow::Result<()> {
         self.tasks.shutdown().await;
-        self.checkpoint().await?;
+        if should_checkpoint {
+            self.checkpoint().await?;
+        }
         self.db.shutdown().await?;
```
avinassh, line 655: I want to confirm one thing, I haven't changed `db::shutdown`, it has `replicator.wait_until_snapshotted().await?` This should be fine, right? As I understand, this one does not trigger a new upload rather wait for existing one to finish.

libsql-server/src/lib.rs
avinassh: Yup, done

libsql-server/src/namespace/mod.rs
MarinPostma: @Horusiath ?

Horusiath: Yes. When a new generation is made, we begin uploading a DB snapshot found at the beginning of that generation. `replicator.wait_until_snapshotted` is used to wait for that process to finish. This way we have an entire generation lifetime for it to complete which may be necessary since snapshot file may be massive.

Calling `checkpoint` twice in a row - once explicitly and once as part of db.shutdown - is basically a roundabout way to compress WAL into db file, trigger snapshot upload and wait for it.

MarinPostma (approved):

MarinPostma:
@v you pr fails against main, can you rebase and try to see what's going on?

avinassh:
@MarinPostma I rebased with main and the tests fail. However, I explicitly enabled snapshot at shutdown in bottomless tests and it is passing. But this is wrong, bottomless should work fine without the snapshot at shutdown

avinassh:
Seems there was an issue with the bottomless shutdown routine, fixed in #736 once that is merged, I will rebase and merge

PR #722, opened 2023-11-28
sqld: Add fault and http response metrics

LucioFranco:


libsql-server/src/http/user/trace.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    tracing::debug!(
+        "got request: {} {} {:?}",
+        req.method(),
+        req.uri(),
+        req.headers()
+    );
+    if let Some(v) = req.headers().get("x-libsql-client-version") {
+        if let Ok(s) = v.to_str() {
+            metrics::increment_counter!(CLIENT_VERSION, "version" => s.to_string());
```
MarinPostma, line 26: that allocation bothers me still 😄

MarinPostma (approved):

libsql-server/src/http/user/trace.rs
```diff
@@ -0,0 +1,103 @@
...
+                "libsql_server_user_http_response",
+                "protocol" => "grpc",
+                "status" => code,
+            );
+        }
+    }
+}
+
+pub(crate) fn failure<F>(_: F, _: Duration, _: &Span) {
+    metrics::increment_counter!("libsql_server_user_http_fault", "protcol" => "http");
```
MarinPostma, line 102: Maybe add the status code? @LucioFranco

libsql-server/tests/common/net.rs
```diff
@@ -141,6 +141,8 @@ pub trait SimServer {
 #[async_trait::async_trait]
 impl SimServer for TestServer {
     async fn start_sim(mut self, user_api_port: usize) -> anyhow::Result<()> {
+        let _ = tracing_subscriber::fmt::try_init();
```
MarinPostma, line 144: there is a `init_tracing` function already somewhere already, and it's called in multiple tests

libsql-server/src/http/user/trace.rs
LucioFranco: Yeah, another option is to use some sort of static lookup table or update metrics to accept integers.

libsql-server/src/http/user/trace.rs
LucioFranco: There is no status code here 😄 part of the issue. So we just need to know for gRPC right now metrics will be a little wonky. But in general I am way more concerned about our http faults than grpc.

libsql-server/tests/common/net.rs
LucioFranco: Yeah, I just called it here just to make sure it gets called. We should probably refactor some of our test harness code a bit since this spot is a bit hacky.

PR #724, opened 2023-11-29
chore(libsql-server): Remove reference to postgres listen address

oliverlambson:
Postgres support was nuked in 0.21.9 as per [this commit](https://github.com/libsql/sqld/pull/400/commits/b0a5855d80a0ef4a7d159919325df9957fc15100) after [this comment](https://github.com/libsql/sqld/pull/400#discussion_r1209967183)

[The Turso blog still talks about this feature](https://blog.turso.tech/sqlite-based-databases-on-the-postgres-protocol-yes-we-can-358e61171d65)—I suggest adding a banner notice at the top of the article that this is no longer supported (I spent the last hour trying to figure out why there was an env var available in the Dockerfile but the `-p` argument to the `sqld` CLI was failing)


MarinPostma:
Thanks for your contribution. Very sorry for the frustration. We have removed the feature months ago, and some occurrences are still in the wild.

We'll fix the blog 👍

oliverlambson:
Thanks!

PR #728, opened 2023-11-29
bottomless: wait for all background tasks to complete on graceful shutdown

Horusiath:
This PR adds a `graceful_shutdown` method to bottomless replicator, which will wait for all known committed frames to be uploaded to S3 and for all pending snapshots to be uploaded as well prior completion.

There's a matter when to call this method as `ReplicationLogger` doesn't seem to expose graceful shuttdown method on its own. @MarinPostma could maybe help here.


MarinPostma:
@Horusiath wouldn't here make sense? https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/database.rs#L66-L67

PR #751, opened 2023-12-04
Add path based routing

avinassh:
This PR adds a way to extract namespace from the path. This is done so that users can have smooth experience in `turso dev` and they can connect to multiple namespaces without editing `/etc/hosts` file. This patch does not require any changes in shell or libraries.

1. The expected path format is `/dev/<namespace>` I am not sure if we want to enable to this feature on the platform. So for now, I kept it behind the `/dev` sub path
2. Similarly, I have enabled only `/dev/:namespace/v2/pipeline` and `/dev/:namespace/v3/pipeline` paths.

I have tested this with turso shell and go sdk.

### DIFF
--- a/libsql-server/src/http/user/db_factory.rs
+++ b/libsql-server/src/http/user/db_factory.rs
@@ -64,6 +64,32 @@ pub fn namespace_from_headers(
     }
 }
 
+pub struct MakeConnectionExtractorPath<D>(pub Arc<dyn MakeConnection<Connection = D>>);
+#[async_trait::async_trait]
+impl<F> FromRequestParts<AppState<F>>
+    for MakeConnectionExtractorPath<<F::Database as Database>::Connection>
+where
+    F: MakeNamespace,
+{
+    type Rejection = Error;
+
+    async fn from_request_parts(
+        parts: &mut Parts,
+        state: &AppState<F>,
+    ) -> Result<Self, Self::Rejection> {
+        let auth = Authenticated::from_request_parts(parts, state).await?;
+        let ns = axum::extract::Path::<NamespaceName>::from_request_parts(parts, state)
+            .await
+            .map_err(|e| Error::InvalidPath(e.to_string()))?;
+        Ok(Self(
+            state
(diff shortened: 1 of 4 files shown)


libsql-server/src/http/user/db_factory.rs
```diff
@@ -64,10 +80,37 @@ pub fn namespace_from_headers(
...
+
 fn split_namespace(host: &str) -> crate::Result<NamespaceName> {
     let (ns, _) = host.split_once('.').ok_or_else(|| {
         Error::InvalidHost("host header should be in the format <namespace>.<...>".into())
     })?;
     let ns = NamespaceName::from_string(ns.to_owned())?;
     Ok(ns)
 }
+
+fn extract_namespace(parts: &Parts) -> Option<String> {
```
avinassh, line 107: there is probably an axum way to extract this namespace from the parts, but I could not figure out how to do it

MarinPostma: it's simpler than it seems: you need implement `Deserialize` for `NamespaceName`:
```rust
impl<'de> Deserialize for NamespaceName<'de> {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de> {
            struct V;

            impl<'de> Visitor<'de> for V {
                type Value = NamespaceName;

                fn expecting(&self, f: &mut fmt::Formatter) -> fmt::Result {
                    write!(f, "a valid namespace name");
                }

                fn visit_string<E>(self, v: String) -> Result<Self::Value, E>
                where
                    E: serde::de::Error, {
                        NamespaceName::from_string(v).map_err(|e| E::custom(e))
                }
            }
        }
}
```

when you have that, you can use the `Path` extractor: https://docs.rs/axum/latest/axum/extract/struct.Path.html

avinassh: Thank you so much, I shall try this 🙏

MarinPostma (approved):

PR #753, opened 2023-12-04
fix libsql Value deserialization

MarinPostma:
I noticed that the row deserialization was not implemented correctly, because `Value::Null` was deserialized with the unit deserializer, which is incorrect. Instead, I implemented a `ValueDeserializer` that handles deserializing values correctly.

### DIFF
--- a/libsql/src/value.rs
+++ b/libsql/src/value.rs
@@ -3,7 +3,7 @@ use std::str::FromStr;
 use crate::{Error, Result};
 
 #[derive(Clone, Debug)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
 pub enum Value {
     Null,
     Integer(i64),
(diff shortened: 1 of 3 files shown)


libsql/src/value.rs
```diff
@@ -426,3 +426,105 @@ impl TryFrom<libsql_replication::rpc::proxy::Value> for Value {
         bincode::deserialize(&value.data[..]).map_err(Error::from)
     }
 }
+
+#[cfg(feature = "serde")]
+mod serde_ {
```
LucioFranco, line 431: why do we need to put _ at the end?

libsql/src/value.rs
```diff
@@ -1,9 +1,9 @@
-use std::str::FromStr;
+use std::{marker::PhantomData, str::FromStr};
```
LucioFranco, line 1: Shouldn't this be imported inside the `serde_` module not outside?

MarinPostma: ah yes, I didn't compile so I forgot to move this one

Issue #758, opened 2023-12-05
Make `is_autocommit()` not async


penberg:
The `is_autocommit()` is now async, which makes it hard to use from the JavaScript and Python bindings. But more importantly, apparently the remote protocol implementation does a round-trip to the server, which is problematic because in the bindings `execute()` implies two round-trips.

penberg:
Fixed by e8500de99957ce9d789de9ab7f0320010794040c

PR #759, opened 2023-12-05
make libsql Conn::is_autocommit synchronous

Horusiath:
Potential rollback, depending on the result of discussion from #758 .

The original decision to change was this discussion: https://github.com/tursodatabase/libsql/pull/653#discussion_r1400789143

### DIFF
--- a/libsql/src/hrana/connection.rs
+++ b/libsql/src/hrana/connection.rs
@@ -3,7 +3,7 @@ use crate::hrana::proto::{Batch, BatchCond, BatchResult, Stmt, StmtResult};
 use crate::hrana::stream::HttpStream;
 use crate::hrana::{HranaError, HttpSend, Result, Statement};
 use crate::util::coerce_url_scheme;
-use std::sync::atomic::{AtomicI64, AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicI64, AtomicU64, Ordering};
 use std::sync::Arc;
 
 #[derive(Debug)]
@@ -21,6 +21,7 @@ where
     auth: String,
     affected_row_count: AtomicU64,
     last_insert_rowid: AtomicI64,
+    is_autocommit: AtomicBool,
 }
 
 impl<T> HttpConnection<T>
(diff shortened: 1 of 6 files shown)


libsql/src/hrana/hyper.rs
```diff
@@ -142,8 +142,8 @@ impl Conn for HttpConnection<HttpSender> {
         })
     }
 
-    async fn is_autocommit(&self) -> crate::Result<bool> {
-        Ok(true) // connection without transaction always commits at the end of execution step
+    fn is_autocommit(&self) -> bool {
+        true // connection without transaction always commits at the end of execution step
```
MarinPostma, line 146: this is not true, one can always execute `BEGIN TRANSACTION` on a connection, then this would be a bug

MarinPostma (commented):
I don't think we should do that, rather, we should accurately report whether a connection is in auto-commit state, either by tracking the state, or querying the server

libsql/src/hrana/hyper.rs
Horusiath: Isn't server-side transaction mapped 1-1 to client-side stream? Atm. if you do `conn.execute("BEGIN TRANSACTION")` it will open a new stream, then immediately send a pair of `"execute: {my_sql_stmt}", "close"` Hrana HTTP requests: I assumed that the close request immediately calls for any resources related to the stream (including transaction) to be released.

libsql/src/hrana/connection.rs
```diff
@@ -30,13 +31,14 @@ where
     pub fn new(url: String, token: String, inner: T) -> Self {
         // The `libsql://` protocol is an alias for `https://`.
         let base_url = coerce_url_scheme(&url);
-        let url_for_queries = format!("{base_url}/v2/pipeline");
+        let url_for_queries = format!("{base_url}/v3/pipeline");
```
Horusiath, line 34: `GetAutocommitReq` requires Hrana v3 to be used.

penberg (approved):
Looks sensible to me, but it's best to get explicit ACK from @MarinPostma and/or @LucioFranco.

libsql/src/hrana/stream.rs
```diff
@@ -274,10 +274,21 @@ where
         Ok(responses)
     }
 
-    async fn finalize(&mut self, req: StreamRequest) -> Result<StreamResponse> {
-        let [resp, _] = self.send_requests([req, StreamRequest::Close]).await?;
+    async fn finalize(&mut self, req: StreamRequest) -> Result<(StreamResponse, bool)> {
+        let [resp, _, get_autocommit] = self
+            .send_requests([req, StreamRequest::GetAutocommit, StreamRequest::Close])
```
MarinPostma, line 279: aren't the results in the wrong order? 🤔 
```suggestion
        let [resp, get_autocommit, _] = self
            .send_requests([req, StreamRequest::GetAutocommit, StreamRequest::Close])
```

MarinPostma: mind adding a sim test?

MarinPostma (changes requested):

libsql/src/hrana/stream.rs
Horusiath: This is tested, hence I flipped the order. I found this pretty weird myself.

MarinPostma (approved):

Issue #749, opened 2023-12-03
de.rs: Support deserialize to Option<T>


wyhaya:
```rs
#[derive(Debug, serde::Deserialize)]
struct Row {
    n: Option<i64>
}

let mut rows = conn.query("SELECT 1 as n", ()).await.unwrap();
// ...
de::from_row::<Row>(&row).unwrap();
```

### Error
```
`Err` value: Error("invalid type: integer `1`, expected option")
```

PR #762, opened 2023-12-06
Support deserialize to Option<T> and Unit

wyhaya:
```rs
let mut rows = conn.query("SELECT null as a, 1 as b, 2 as c, null as d, null as e", ()).await.unwrap();
// ...
#[derive(Deserialize)]
struct Row {
    // From Null: `None`
    a: Option<i64>,
    // From Integer: `Some(1)`
    b: Option<i64>,
    // From Integer: `2`
    c: i64,
    // From Null: `()`
    d: (),
    // From Null: `None`
    e: Option<()>,
}
```

fix: #749


MarinPostma:
Thanks! do you mind adding a couple of tests please 🙏

MarinPostma:
Great job thanks!

PR #773, opened 2023-12-10
libsql: Add `visit_some` for value deserializer

JonahLund:
This PR adds `visit_some` visitor for the `Value` deserializer

### DIFF
--- a/libsql/src/value.rs
+++ b/libsql/src/value.rs
@@ -444,7 +444,7 @@ mod serde_ {
             D: serde::Deserializer<'de>,
         {
             struct V;
-            impl Visitor<'_> for V {
+            impl<'de> Visitor<'de> for V {
                 type Value = Value;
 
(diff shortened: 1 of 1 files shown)


JonahLund:
Actually i'll probably add more ones, i noticed that the deserialization was failing for a couple of unimplemented values for me

libsql/src/value.rs
```diff
@@ -458,13 +458,76 @@ mod serde_ {
...
+                {
+                    Ok(Value::Integer(v as i64))
+                }
+
+                fn visit_u16<E>(self, v: u16) -> std::result::Result<Self::Value, E>
+                where
+                    E: serde::de::Error,
+                {
+                    Ok(Value::Integer(v as i64))
+                }
```
MarinPostma, line 508: shouldn't those already be handled by the delegation?

JonahLund: I don't believe so, i got an error when deserializing an integer other than i64 which was solved by this

MarinPostma (approved):

PR #798, opened 2023-12-18
Libsql client: introduce async row fetcher

Horusiath:
This PR works on top of #750 . 

It integrates Hrana cursor API into mainline of Rows abstraction in libsql client and enables to fetch cursors asynchronously without pre-fetching or buffering.

### DIFF
--- a/libsql-server/tests/embedded_replica/mod.rs
+++ b/libsql-server/tests/embedded_replica/mod.rs
@@ -499,7 +499,7 @@ fn replicate_with_snapshots() {
         let conn = db.connect().unwrap();
         conn.execute("create table test (x)", ()).await.unwrap();
         // insert enough to trigger snapshot creation.
-        for _ in 0..200 {
+        for _ in 0..ROW_COUNT {
             conn.execute("INSERT INTO test values (randomblob(6000))", ())
                 .await
                 .unwrap();
@@ -530,7 +530,7 @@ fn replicate_with_snapshots() {
                 .unwrap()
                 .as_integer()
                 .unwrap(),
-            200
+            ROW_COUNT
         );
 
         Ok(())
--- a/libsql-server/tests/embedded_replica/mod.rs
+++ b/libsql-server/tests/embedded_replica/mod.rs
@@ -283,6 +283,7 @@ fn replica_primary_reset() {
             .await
             .unwrap()
             .next()
+            .await
             .unwrap()
             .unwrap()
             .get_value(0)
@@ -296,6 +297,7 @@ fn replica_primary_reset() {
             .await
             .unwrap()
             .next()
+            .await
             .unwrap()
             .unwrap()
             .get_value(0)
@@ -339,6 +341,7 @@ fn replica_primary_reset() {
             .await
             .unwrap()
             .next()
+            .await
             .unwrap()
             .unwrap()
             .get_value(0)
@@ -352,6 +355,7 @@ fn replica_primary_reset() {
             .await
             .unwrap()
             .next()
+            .await
             .unwrap()
             .unwrap()
             .get_value(0)
@@ -519,6 +523,7 @@ fn replicate_with_snapshots() {
         let mut res = conn.query("select count(*) from test", ()).await.unwrap();
         assert_eq!(
             *res.next()
+                .await
                 .unwrap()
                 .unwrap()
                 .get_value(0)
--- a/libsql-server/tests/embedded_replica/mod.rs
+++ b/libsql-server/tests/embedded_replica/mod.rs
@@ -461,6 +461,7 @@ fn replicate_with_snapshots() {
         .tcp_capacity(200)
         .build();
 
+    const ROW_COUNT: i64 = 200;
     let tmp = tempdir().unwrap();
 
     init_tracing();
@@ -617,7 +618,7 @@ fn proxy_write_returning_row() {
             .await
             .unwrap();
 
-        rows.next().unwrap().unwrap();
+        rows.next().await.unwrap().unwrap();
 
         Ok(())
     });
--- a/libsql/src/hrana/hyper.rs
+++ b/libsql/src/hrana/hyper.rs
@@ -85,11 +85,8 @@ impl HttpSend for HttpSender {
     }
 
     fn oneshot(self, url: Arc<str>, auth: Arc<str>, body: String) {
-        if let Ok(rt) = tokio::runtime::Handle::try_current() {
-            rt.spawn(self.send(url, auth, body));
-        } else {
-            tracing::warn!("couldn't send Hrana oneshot request: no tokio runtime active")
-        }
+        let rt = tokio::runtime::Handle::current();
+        rt.spawn(self.send(url, auth, body));
     }
 }
 
--- a/libsql/src/hrana/hyper.rs
+++ b/libsql/src/hrana/hyper.rs
@@ -85,8 +85,7 @@ impl HttpSend for HttpSender {
     }
 
     fn oneshot(self, url: Arc<str>, auth: Arc<str>, body: String) {
-        let rt = tokio::runtime::Handle::current();
-        rt.spawn(self.send(url, auth, body));
+        tokio::spawn(self.send(url, auth, body));
     }
 }
 
--- a/libsql/src/hrana/hyper.rs
+++ b/libsql/src/hrana/hyper.rs
@@ -8,7 +8,6 @@ use crate::params::Params;
 use crate::transaction::Tx;
 use crate::util::ConnectorService;
 use crate::{Rows, Statement};
-use bytes::Bytes;
 use futures::future::BoxFuture;
 use futures::{Stream, TryStreamExt};
 use http::header::AUTHORIZATION;
--- a/libsql/src/hrana/hyper.rs
+++ b/libsql/src/hrana/hyper.rs
@@ -14,9 +14,10 @@ use futures::{Stream, TryStreamExt};
 use http::header::AUTHORIZATION;
 use http::{HeaderValue, StatusCode};
 use hyper::body::HttpBody;
+use std::io::ErrorKind;
 use std::sync::Arc;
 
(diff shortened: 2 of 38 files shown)


libsql-server/src/connection/write_proxy.rs
```diff
@@ -332,7 +338,23 @@ where
                         continue;
                     }
 
-                    if !response_cb(resp.response.ok_or(Error::PrimaryStreamMisuse)?)? {
+                    let should_continue = tokio::task::spawn_blocking({
```
Horusiath, line 341: This change is related to what @MarinPostma proposed in #775 (I think) - I used it his code, otherwise the tests would deadlock in tests when stream dropping schedules sending close request in background task.

libsql-server/tests/embedded_replica/mod.rs
```diff
@@ -452,6 +456,7 @@ fn replica_no_resync_on_restart() {
 
 #[test]
 fn replicate_with_snapshots() {
+    const ROW_COUNT: i64 = 120;
```
Horusiath, line 459: This would be a question for @MarinPostma : I had to change number of inserts in this tests from 200&rarr;120. When passing over 155 inserts on my local machine, the test fails as it says that the simulation cannot finish within 10sec.

The thing is that I've checked and these 150+ requests finish within ~600ms. I've checked if generated streams don't leak and it looks like they are properly dropped, causing stream close on the server side as well.

I'm still trying to find out why 200 inserts loop causes a problem.

MarinPostma: You can instead increase the simulation duration on the builder, just set it to an arbitrary duration

Horusiath: How is the simulation time being computed? My problem is that I'm hitting 10s simulation execution limit, yet the real time of the test is <1s:
image

MarinPostma: Time is ticked very fast in simultatins, but time doesn't tick at the same speed in blocking threads, just set "simulation_duration" in the builder to something large enough

libsql/src/hrana/mod.rs
```diff
@@ -198,33 +224,38 @@ impl RowsInner for Rows {
...
+    fn column_count(&self) -> i32 {
+        self.column_count()
+    }
+
+    fn column_name(&self, idx: i32) -> Option<&str> {
+        self.column_name(idx)
+    }
+
+    fn column_type(&self, _idx: i32) -> crate::Result<ValueType> {
+        todo!()
```
LucioFranco, line 258: are we tracking this todo in an issue somewhere?

libsql/src/hrana/stream.rs
```diff
@@ -350,30 +351,43 @@ where
...
+        tracing::trace!("closing stream");
+        let req = serde_json::to_string(&ClientMsg {
+            baton: self.baton.clone(),
+            requests: vec![StreamRequest::Close],
+        })
+        .unwrap();
+        self.client
+            .clone()
+            .oneshot(self.pipeline_url.clone(), self.auth_token.clone(), req);
+        self.done();
```
LucioFranco, line 378: related to my comment above I think it makes more sense here to do all the spawn logic, unless I am missing something with wasm that makes this slightly more complicated.

libsql/src/statement.rs
```diff
@@ -92,18 +93,37 @@ pub struct MappedRows<F> {
...
-            .transpose()
-            .map(|row_result| row_result.and_then(map))
+    fn poll_next(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Self::Item>> {
+        use futures::{ready, Future};
+
+        let mut rows = unsafe { self.as_mut().map_unchecked_mut(|pin| &mut pin.rows) };
+        let mut fut = Box::pin(rows.next());
```
LucioFranco, line 116: FYI can use this to avoid the allocations each time https://docs.rs/tokio-util/latest/tokio_util/sync/struct.ReusableBoxFuture.html

libsql/src/statement.rs
```diff
@@ -92,18 +93,37 @@ pub struct MappedRows<F> {
...
+        let res = ready!(fut.as_mut().poll(cx));
```
LucioFranco, line 117: I don' think this is correct, if the rows future is not available right away and returns pending we cancel it. What happens there? I think we need to store the future inside the stream itself.

LucioFranco (changes requested):
I have

libsql/src/hrana/hyper.rs
```diff
@@ -82,6 +83,14 @@ impl HttpSend for HttpSender {
         let fut = self.clone().send(url, auth, body);
         Box::pin(fut)
     }
+
+    fn oneshot(self, url: Arc<str>, auth: Arc<str>, body: String) {
+        if let Ok(rt) = tokio::runtime::Handle::try_current() {
+            rt.spawn(self.send(url, auth, body));
+        } else {
+            tracing::warn!("couldn't send Hrana oneshot request: no tokio runtime active")
+        }
```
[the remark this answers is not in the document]
Horusiath: Potentially, when we're closing the app, this object could be dropped after tokio runtime. While it's unlikely the case in regular Rust code, C-based bindings are different matter.

libsql/src/hrana/mod.rs
```diff
@@ -38,9 +38,12 @@ struct Cookie {
 }
 
 pub trait HttpSend: Clone {
-    type Stream: Stream<Item = Result<Bytes>> + Unpin;
+    type Stream: Stream<Item = std::io::Result<Bytes>> + Unpin;
     type Result: Future<Output = Result<Self::Stream>>;
     fn http_send(&self, url: Arc<str>, auth: Arc<str>, body: String) -> Self::Result;
+
+    /// Schedule sending a HTTP post request without waiting for the completion.
+    fn oneshot(self, url: Arc<str>, auth: Arc<str>, body: String);
```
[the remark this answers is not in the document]
Horusiath: As usual: cloudflare doesn't operate on tokio and has its own way of spawning tasks. Since this is called from `Drop::drop` it cannot be specialised between CloudflareSender and regular Hyper sender.

libsql/src/hrana/mod.rs
Horusiath: I don't think so, I've only started a discussion on Slack about this problem.

libsql/src/statement.rs
Horusiath: If I got this correctly, this type works essentially like iter Map. I modified it into most straightforward way possible. I don't think it should have a freedom of cancelling the underlying futures on its own. Maybe there's some other Stream-based mapper that we could use instead of this one?

libsql/src/statement.rs
```diff
@@ -92,18 +93,37 @@ pub struct MappedRows<F> {
...
-            .next()
-            .transpose()
-            .map(|row_result| row_result.and_then(map))
+    fn poll_next(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Self::Item>> {
+        use futures::{ready, Future};
+
+        let mut rows = unsafe { self.as_mut().map_unchecked_mut(|pin| &mut pin.rows) };
```
MarinPostma, line 115: you should use pin_project to avoid that

libsql/src/statement.rs
MarinPostma, line 116: `rows.next()` should already return a `Box<dyn Future>`

libsql/src/statement.rs
```diff
@@ -92,18 +93,37 @@ pub struct MappedRows<F> {
     map: F,
 }
 
-impl<F, T> Iterator for MappedRows<F>
+impl<F> MappedRows<F> {
+    pub fn new(rows: Rows, map: F) -> Self {
+        Self { rows, map }
+    }
+}
```
MarinPostma, line 100: Overall, I think it would be a lot easier to have a `into_stream` method on Rows, and then use `Stream`'s provided methods for mapping etc

MarinPostma: it would look something like:
```rust
fn into_stream(self) -> impl Stream<...> {
    async_stream! {
        yield self.next().await
     }
}
```

libsql/src/hrana/hyper.rs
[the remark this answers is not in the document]
LucioFranco: How is that possible if we have a runtime like this https://github.com/tursodatabase/libsql/blob/main/bindings/c/src/lib.rs#L15?

libsql/src/hrana/mod.rs
Horusiath: I changed that to work on the best effort basis - when we iterate over the first row, we'll infer and cache the column types from it. If user will try to call the `column_type` prior fetching first row, we'll return an error. Atm. it's the best that we can do.

libsql/src/hrana/hyper.rs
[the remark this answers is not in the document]
Horusiath: I changed that, we'll see how it goes.

libsql/src/rows.rs
```diff
@@ -74,6 +75,18 @@ impl Rows {
...
+    /// one by one. This stream can be further used with [futures::StreamExt]
+    /// operators.
+    #[cfg(feature = "stream")]
+    pub fn into_stream(mut self) -> impl futures::Stream<Item = Result<Row>> + Unpin {
+        Box::pin(async_stream::try_stream! {
+            if let Some(row) = self.next().await? {
+                yield row
+            }
+        })
+    }
```
MarinPostma, line 89: I don't think it's necessary to `Box::pin` the stream and require `Unpin`. The caller can choose to either box it, of simply `tokio::pin!()` it to the stack if that's what they want

Horusiath:
@MarinPostma I've applied all of your remarks.

MarinPostma (approved):

PR #802, opened 2023-12-18
libsql: full docs pass of all public members

LucioFranco:


libsql/README.md
```diff
@@ -10,51 +10,6 @@
 
 This repository contains the libSQL API for Rust.
 
-## Installation
```
notrab, line 13: @LucioFranco will chat to you on Slack about this part.

PR #821, opened 2023-12-21
add a ci task to build cr-sqlite binaries against libsql

tantaman:


tantaman:
How would you like this build to be triggered? It is currently triggered when a tag matching a certain pattern is pushed.

psarna (approved):

PR #825, opened 2023-12-22
Move vendored/rusqlite to libsql-rusqlite

MarinPostma:
In order to publish libsql, we need to publish rusqlite, so I moved the `vendored/rusqlite` to the root of the workspace and renamed it to `libsql-rusqlite`.

Migrating away from rusqlite is too much work, and ain't nobody got time for that (for now).


penberg (approved):

LucioFranco (changes requested):
I want to revisit this approach, i do not want to have 4 crates for the single client crate. I have some ideas that we discussed on slack so lets revisit in the new year.

PR #843, opened 2023-12-31
libsql: Add boolean visitor

JonahLund:
Adds a boolean visitor for value deserializer (not sure why i didn't add this previously)


MarinPostma (approved):
thanks!

Issue #795, opened 2023-12-16
Build docker images for arm machines


avinassh:
It seems the docker images we build work only on amd64

reported here: https://github.com/pressly/goose/pull/658#discussion_r1428806442

> > main_test.go:123: unexpected error: no matching manifest for linux/arm64/v8 in the manifest list entries

> Afaics the docker image only supports amd64,

> https://github.com/tursodatabase/libsql/pkgs/container/libsql-server/158573424

PR #845, opened 2024-01-01
ci: release Docker images for arm64

QuiiBz:
Closes https://github.com/tursodatabase/libsql/issues/795

This PR adds an additional `platform` to `docker buildx` to support arm64 (`linux/arm64`, the default for GitHub runners is `linux/amd64`).

[This commit](https://github.com/tursodatabase/libsql/pull/845/commits/c498c98e693d498f460a62fad6ed79b9cca737ec) will allow to test if it still builds successfully for both platforms before merging.

### DIFF
--- a/.github/workflows/publish-server.yml
+++ b/.github/workflows/publish-server.yml
@@ -26,6 +26,8 @@ jobs:
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
+        with:
+          platforms: linux/amd64,linux/arm64
 
       - name: Log in to the Container registry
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-server.yml
```diff
@@ -5,6 +5,8 @@ on:
     branches: ['main']
     tags:
       - libsql-server-v*.*.*
+  # REMOVE BEFORE MERGING
+  pull_request:
```
penberg, line 9: @QuiiBz I think now is the time to remove this 🤣

QuiiBz: Indeed, looks like the job passed!

Screenshot 2024-01-02 at 16 24 04

PR #847, opened 2024-01-02
Libsql single stream connection

### LINKED ISSUE #813: Commit fails with remote database although not in autocommit mode
When running the example against a remote database, `COMMIT` fails despite being in non-autocommit mode:

```
diff --git a/libsql/examples/example.rs b/libsql/examples/example.rs
index 9d898e0eda..0374c38c47 100644
--- a/libsql/examples/example.rs
+++ b/libsql/examples/example.rs
@@ -15,7 +15,7 @@ async fn main() {

     let conn = db.connect().unwrap();

-    conn.query("select 1; select 1;", ()).await.unwrap();
+    conn.query("select 1;", ()).await.unwrap();

     conn.execute("CREATE TABLE IF NOT EXISTS users (email TEXT)", ())
         .await
@@ -26,7 +26,10 @@ async fn main() {
         .await
         .unwrap();

+    conn.execute("BEGIN", ()).await.unwrap();
     stmt.execute(["foo@example.com"]).await.unwrap();
+    assert!(!conn.is_autocommit());
+    conn.execute("COMMIT", ()).await.unwrap();

     let mut stmt = conn
         .prepare("SELECT * FROM users WHERE email = ?1")
```

```
penberg@vonneumann libsql % export LIBSQL_URL=$(turso db show --url foo)
penberg@vonneumann libsql % export LIBSQL_AUTH_TOKEN=$(turso db tokens create foo)penberg@vonneumann libsql % cargo run --example example
   Compiling libsql v0.2.0-alpha.1 (/Users/penberg/src/libsql/libsql/libsql)
    Finished dev [unoptimized + debuginfo] target(s) in 2.19s
     Running `target/debug/examples/example`
thread 'main' panicked at libsql/examples/example.rs:32:38:
called `Result::unwrap()` on an `Err` value: Hrana(StreamError(StreamResponseError { error: Error { message: "SQLite error: cannot commit - no transaction is active" } }))
```

Horusiath:
This PR is build on top of #798 and should fix #813

This PR addresses an issue with Hrana-based connection behaviour that @penberg has noticed. Essentially, non-transactional execute requests are always running in auto-commit mode, even if they are executed in `begin;`+` ...;`+ `commit;` series of steps:

```rust
let db = Database::open_remote_with_connector(...)?;
let conn = db.connect()?;
conn.execute("create table test (x)", ()).await?;
conn.execute("begin;", ()).await?; 
conn.execute("insert into test values (12);", ()).await?; // this statement is auto-committed
conn.execute("rollback;", ()).await?; // this statement fails, because we're not in a transaction scope
```

This was caused by HTTP connection running each `conn.execute` as a separate Hrana stream - therefore each execute statement got its own new SQLite connection, which was committed right after.

This PR changes that by assigning one general stream for HTTP connection, used throughout its lifetime. If we're creating transactions explicitly via `conn.transaction()`, these are going to get their own unique streams.

### DIFF
--- a/libsql-server/tests/standalone/mod.rs
+++ b/libsql-server/tests/standalone/mod.rs
@@ -243,6 +243,43 @@ fn basic_query_fail() {
 }
 
 #[test]
+fn query_begin_rollback() {
+    let mut sim = turmoil::Builder::new().build();
+
+    sim.host("primary", make_standalone_server);
+
+    sim.client("test", async {
+        let db = Database::open_remote_with_connector("http://primary:8080", "", TurmoilConnector)?;
+        let conn = db.connect()?;
+
+        conn.execute("create table test (x)", ()).await?;
+
+        conn.execute("begin;", ()).await?;
+        conn.execute("insert into test values (12);", ()).await?;
+
+        // we can read the inserted row
+        let mut rows = conn.query("select count(*) from test", ()).await?;
+        assert_eq!(
+            rows.next().await.unwrap().unwrap().get_value(0).unwrap(),
+            Value::Integer(1)
+        );
+
+        conn.execute("rollback;", ()).await?;
+
+        // after rollback row is no longer there
+        let mut rows = conn.query("select count(*) from test", ()).await?;
+        assert_eq!(
+            rows.next().await.unwrap().unwrap().get_value(0).unwrap(),
+            Value::Integer(0)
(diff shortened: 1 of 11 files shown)


LucioFranco (approved):
Really nice work!

MarinPostma (approved):

PR #858, opened 2024-01-05
Update cargo dist to 0.7.0

### LINKED ISSUE #793: Homebrew formula is incorrect
We use axo's cargo dist to generate the homebrew formula for libsql-server. But the final formula generated is incorrect: https://github.com/libsql/homebrew-sqld/blob/05fac6e/Formula/sqld.rb#L2,#L8 

The formula checks only CPU architecture, but not OS. Because of this, it ends up installing incorrect binary on Linux, instead of failing. 

For instance, compare this to our CLI formula where we explicitly check for OS too: https://github.com/tursodatabase/homebrew-tap/blob/main/turso.rb#L10

This will likely to be fixed in the upcoming version of cargo dist - https://github.com/axodotdev/cargo-dist/releases/tag/v0.6.0-prerelease.4

avinassh:
closes #793

### DIFF
--- a/.github/workflows/release-libsql-server.yml
+++ b/.github/workflows/release-libsql-server.yml
@@ -58,7 +58,7 @@ jobs:
         with:
           submodules: recursive
       - name: Install cargo-dist
-        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.4.2/cargo-dist-installer.sh | sh"
+        run: "curl --proto '=https' --tlsv1.2 -LsSf https://github.com/axodotdev/cargo-dist/releases/download/v0.7.0/cargo-dist-installer.sh | sh"
       - id: plan
         run: |
           cargo dist plan ${{ !github.event.pull_request && format('--tag={0}', github.ref_name) || '' }} --output-format=json > dist-manifest.json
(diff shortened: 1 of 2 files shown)


avinassh:
I did a dry run (ran `cargo dist build --artifacts=global`) and it generated the expected homebrew formula: 

```ruby
class LibsqlServer < Formula
  version "0.22.7"
  on_macos do
    on_arm do
      url "https://github.com/tursodatabase/libsql/releases/download/v0.22.7/libsql-server-aarch64-apple-darwin.tar.xz"
    end
    on_intel do
      url "https://github.com/tursodatabase/libsql/releases/download/v0.22.7/libsql-server-x86_64-apple-darwin.tar.xz"
    end
  end
  on_linux do
    on_intel do
      url "https://github.com/tursodatabase/libsql/releases/download/v0.22.7/libsql-server-x86_64-unknown-linux-gnu.tar.xz"
    end
  end

  def install
    bin.install "sqld"

    # Homebrew will automatically install these, so we don't need to do that
    doc_files = Dir["README.*", "readme.*", "LICENSE", "LICENSE.*", "CHANGELOG.*"]
    leftover_contents = Dir["*"] - doc_files

    # Install any leftover files in pkgshare; these are probably config or
    # sample files.
    pkgshare.install *leftover_contents unless leftover_contents.empty?
  end
end
```

penberg (approved):

PR #861, opened 2024-01-05
Enable per namespace jwt key

haaawk:


libsql-server/src/namespace/mod.rs
```diff
@@ -764,6 +765,21 @@ impl<M: MakeNamespace> NamespaceStore<M> {
     ) -> crate::Result<MetaStoreHandle> {
         self.with(namespace, |ns| ns.db_config_store.clone()).await
     }
+
+    pub async fn jwt_key(
```
LucioFranco, line 769: So I actually don't really like this approach, to me I envisioned that this would exist on the `Namespace` struct itself and you would get access to it via `with`. So once you have parsed the namespace out of the http header, you check the store for it, it then gives you a `Namespace<T>`  and on that namespace you can then access the config. There is no need to have a separate function that threads through namespaces since its all type checked via the store/namespace struct. This imo provides a really nice encapsulation where you can only access the config from the `Namespace` struct that is tied to that namespace.

We can keep this for now but its quite leaky imo and we should fix this in the future and track that.

libsql-server/src/rpc/proxy.rs
```diff
@@ -493,13 +493,18 @@ impl Proxy for ProxyService {
...
+            .namespaces
+            .jwt_key(namespace.clone())
+            .await
+            .map_err(|_| tonic::Status::internal("Error fetching jwt key for a namespace"))?;
         let auth = if let Some(auth) = &self.auth {
-            auth.authenticate_grpc(&req, self.disable_namespaces)?
+            auth.authenticate_grpc(&req, self.disable_namespaces, namespace_jwt_key)?
         } else {
             Authenticated::from_proxy_grpc_request(&req, self.disable_namespaces)?
         };
```
LucioFranco, line 506: I think we can move this all to a fn and make this more abstracted? You don't need to know the T of the request so you can write a fn that accepts a `req: Request<T>`, that should allow us to unify the ones below.

LucioFranco (changes requested):
I left some feedback on two big ticket items but for me the most important thing is that we at least remove all the `context` calls and wrap those in explicit error variants.

Some of the feedback is nits that to me remove a few of the leaky abstractions that we have here, I am okay with merging without those fixes but we should track those.

libsql-server/src/http/admin/mod.rs
```diff
@@ -219,13 +221,19 @@ struct HttpDatabaseConfig {
...
 }
 
 async fn handle_post_config<M: MakeNamespace, C>(
     State(app_state): State<Arc<AppState<M, C>>>,
     Path(namespace): Path<String>,
     Json(req): Json<HttpDatabaseConfig>,
 ) -> crate::Result<()> {
+    if let Some(jwt_key) = req.jwt_key.as_deref() {
+        // Check that the jwt key is correct
+        parse_jwt_key(jwt_key).context("Could not parse JWT decoding key")?;
```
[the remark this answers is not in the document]
haaawk: Would you have an example of something similar that's already in the codebase, please? @LucioFranco

libsql-server/src/rpc/proxy.rs
haaawk: When I extracted:
```
+    async fn auth<T>(&self, namespace: NamespaceName, msg: &tonic::Request<T>) -> Result<Authenticated, tonic::Status> {
+        let namespace_jwt_key = self
+            .namespaces
+            .jwt_key(namespace)
+            .await
+            .map_err(|_| tonic::Status::internal("Error fetching jwt key for a namespace"))?;
+        if let Some(auth) = &self.auth {
+            auth.authenticate_grpc(&msg, self.disable_namespaces, namespace_jwt_key)
+        } else {
+            Authenticated::from_proxy_grpc_request(&msg, self.disable_namespaces)
+        }
+    }
```
The compiler complained with:
```
error: future cannot be sent between threads safely
   --> libsql-server/src/rpc/proxy.rs:508:73
    |
508 |       ) -> Result<tonic::Response<Self::StreamExecStream>, tonic::Status> {
    |  _________________________________________________________________________^
509 | |         let namespace = super::extract_namespace(self.disable_namespaces, &req)?;
510 | |         let namespace_jwt_key = self
511 | |             .namespaces
...   |
549 | |         Ok(tonic::Response::new(Box::pin(stream)))
550 | |     }
    | |_____^ future created by async block is not `Send`
    |
    = help: the trait `Sync` is not implemented for `(dyn tonic::codec::Decoder<Item = ExecReq, Error = tonic::Status> + std::marker::Send + 'static)`
note: future is not `Send` as this value is used across an await
   --> libsql-server/src/rpc/proxy.rs:521:55
    |
521 |         let auth = self.auth(namespace.clone(), &req).await?;
    |                                                 ----  ^^^^^ - `&req` is later dropped here
    |                                                 |     |
    |                                                 |     await occurs here, with `&req` maybe used later
    |                                                 has type `&tonic::Request<Streaming<ExecReq>>` which is not `Send`
```
so I ended up just extracting the `namespace_jwt_key()` function.

libsql-server/src/http/user/mod.rs
```diff
@@ -479,10 +479,18 @@ where
         parts: &mut Parts,
         state: &AppState<M>,
     ) -> Result<Self, Self::Rejection> {
+        let ns = db_factory::namespace_from_headers(
```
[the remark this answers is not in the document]
haaawk: It's just a single function call that's repeated. Is it really such a big deal?

libsql-server/src/namespace/mod.rs
haaawk: Moved jwt_key to Namespace.

libsql-server/src/http/admin/mod.rs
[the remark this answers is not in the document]
LucioFranco: In this case, it looks like we have a lot of detailed error messages here https://github.com/tursodatabase/libsql/blob/4db7a4ce3dad009b98c34861a33e5785d896e4f7/libsql-server/src/auth.rs#L327 so I would just remove the call to `context` here.

In general, the pattern inside `parse_jwt_key` we should completely move away from because it tends to hide a lot of the actual errors that are useful for debugging due to the way anyhow uses debug over display. Happy to explain this more on slack but essentially we should avoid using `context` and instead when we want more detailed contextual errors use a new error enum that has those encoded into the variant via `thiserror` macro annotations.

libsql-server/src/http/user/mod.rs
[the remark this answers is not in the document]
LucioFranco: No you're right this is fine after looking at it again, we have somewhat similar logic repeated but this one is quite independent.

libsql-server/src/rpc/proxy.rs
LucioFranco: this is quite odd...maybe its some weird interaction with the borrow but I feel like ive done this a million times without issues. But if we are hitting errors we can deal with that later.

libsql-server/src/namespace/mod.rs
```diff
@@ -803,6 +804,17 @@ impl<T: Database> Namespace<T> {
         self.db.shutdown().await?;
         Ok(())
     }
+
+    pub fn jwt_key(&self) -> crate::Result<Option<jsonwebtoken::DecodingKey>> {
+        let config = self.db_config_store.get();
+        if let Some(jwt_key) = config.jwt_key.as_deref() {
+            Ok(Some(
+                parse_jwt_key(jwt_key).context("Could not parse JWT decoding key")?,
```
MarinPostma, line 812: It would be best if you decoded the key only once, when the namespace is loaded. As per jsonwebtoken Documentation:
> All the different kind of keys we can use to decode a JWT. This key can be re-used so make sure you only initialize it once if you can for better performance.

Also, consider wrapping it in an `Arc`: cloning the key is not cheap.

haaawk: I was thinking about it but didn't have a good place to store the key. Let's do this in a follow up as this PR is hanging for some time now.

libsql-server/src/rpc/proxy.rs
haaawk: done

PR #872, opened 2024-01-09
fixing namespaces::delete_namespace uncovers a bug

Horusiath:
This PR uncovers and fixes issue with namespace deletion:

1. Original test `namespaces::delete_namespace` didn't work correctly, since:
    1. POST `namespaces/foo/destroy` is not a valid request/route.
    2. Test assertion was failing on error not because we sent a request to deleted namespace, but we tried to create table which already existed.
2. As it turns out, `PrimaryNamespace::destroy` removed all necessary entries, **except the namespace configuration in meta store**. The thing is that when a request was sent from a client to a deleted namespace, meta store still had a necessary config. In result we opened a new DB connection using that config: and since all sqld connections are opening DB file using open-or-create semantics, we recreated a deleted database.

### DIFF
--- a/libsql-server/src/namespace/mod.rs
+++ b/libsql-server/src/namespace/mod.rs
@@ -279,6 +279,7 @@ impl MakeNamespace for PrimaryNamespaceMaker {
         }
 
         if ns_path.try_exists()? {
+            tracing::debug!("removing database directory: {}", ns_path.display());
             tokio::fs::remove_dir_all(ns_path).await?;
         }
 
@@ -473,7 +474,7 @@ impl<M: MakeNamespace> NamespaceStore<M> {
                 &self.inner.metadata,
             )
             .await?;
-
+        self.inner.metadata.remove(&namespace)?;
         tracing::info!("destroyed namespace: {namespace}");
 
         Ok(())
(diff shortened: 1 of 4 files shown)


MarinPostma (approved):

libsql-server/src/namespace/mod.rs
```diff
@@ -473,7 +474,7 @@ impl<M: MakeNamespace> NamespaceStore<M> {
                 &self.inner.metadata,
             )
             .await?;
-
+        self.inner.metadata.remove(&namespace)?;
```
MarinPostma, line 477: I would remove the entry from the state before deallocating the in memory resources, because you don't hold a lock to it. What could happen here is:
- deallocate the namespace
- a request tries to allocate it again, it still exist in the store, so we can reallocate
- we perform the metadata deletion

MarinPostma (changes requested):

libsql-server/src/namespace/mod.rs
Horusiath: Apparently changing the order causes the destroy namespace test to fail. I need to check out why.

Horusiath: The reason why switching the order didn't work was that we first removed db config from MetaStore, but then in `PrimaryNamespace::destroy` we access it by get-or-create semantics, therefore reinserting it again. 

I've changed the behaviour, so not `PrimayNamespace::destroy` accesses the DB config by remove-and-return semantics from MetaStore.

MarinPostma (approved):

PR #876, opened 2024-01-10
Add Iterator and IntoIterator traits for Rows type

lpturmel:
This PR adds the implementation of the `Iterator` and `IntoIterator` traits to the Rows struct.

I don't know if currently there is a better way of doing something like this, but I wanted an API that looked a bit like what we have on the `libsql-client-rs` crate where you can interact with the returned rows as an iterator (in this case because rows is a Vec<Row>) like so:

```rust
    let rows = conn.query(
        "select * from hello",
        (),
    ).await.unwrap();
    let _ = rows
        .into_iter()
        .map(|r| from_row::<MyType>(&r.unwrap()).unwrap())
        .collect::<Vec<_>>();
```

Let me know what you guys think/what should be changed about the implementation.

### DIFF
--- a/libsql/src/rows.rs
+++ b/libsql/src/rows.rs
@@ -76,6 +76,28 @@ impl Rows {
     }
 }
 
+impl IntoIterator for Rows {
+    type Item = Result<Row>;
+    type IntoIter = RowsIntoIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        RowsIntoIter { rows: self }
+    }
+}
+
+/// An iterator over a set of rows.
+pub struct RowsIntoIter {
+    rows: Rows,
+}
+
+impl Iterator for RowsIntoIter {
+    type Item = Result<Row>;
+
+    fn next(&mut self) -> Option<Self::Item> {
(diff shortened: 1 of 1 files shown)


Horusiath:
Soon rows will enable async fetching (see #798). It also comes with an ability to convert Rows into Stream, with all operators available thanks to that conversion.

SET001:
@Horusiath so now, with that PR, how to iterate rows?

lpturmel:
I think this is how it is intended to be used:

```rust
use futures::StreamExt;
use libsql::de::from_row;
use libsql::Builder;

#[tokio::main]
async fn main() {
    // ... usual setup
    let todos = conn.query("SELECT * FROM todos", ()).await.unwrap();
    let stream = todos.into_stream();

    let todos = stream
        .map(|row| from_row::<Todo>(&row.unwrap()).unwrap())
        .collect::<Vec<_>>()
        .await;
}

```
This is using `StreamExt` from the `futures` crate.

SET001:
@lpturmel thank you. this work just fine

PR #883, opened 2024-01-10
Only return persisted stats from /v1/namespaces/:namespace/stats endpoint

athoscouto:


libsql-server/src/stats.rs
```diff
@@ -244,9 +288,15 @@ async fn try_persist_stats(stats: Weak<Stats>, path: &Path) -> anyhow::Result<()
         .create(true)
         .open(&temp_path)
         .await?;
+
+    let stats = stats.upgrade().unwrap();
```
LucioFranco, line 292: We should handle this unwrap likely, we can just do something like 

```
if let Some(stats) = stats.upgrade() {
   // create_snapshot
} else {
    tracing::error!("stats handle doesn't exist, not persisting stats");
}
```

libsql-server/src/stats.rs
```diff
@@ -104,60 +159,54 @@ impl Stats {
         Ok(this)
     }
 
+    /// creates a snapshot from current the stats
+    fn create_snapshot(&self) -> StatsData {
+        self.current.snapshot()
+    }
+
+    /// sets the latest snapshot of the stats
+    fn set_snapshot(&self, new: StatsData) {
```
LucioFranco, line 168: This is really good, keeps the critical section really low 👍

libsql-server/src/stats.rs
```diff
@@ -51,11 +51,15 @@ impl SlowestQuery {
     }
 }
 
-#[derive(Debug, Default, Serialize, Deserialize)]
+#[derive(Debug, Default)]
 pub struct Stats {
-    #[serde(skip)]
     namespace: NamespaceName,
+    current: StatsData,
+    snapshot: Arc<RwLock<StatsData>>,
```
LucioFranco, line 58: This is correct as well 👍 we want a RwLock incase we have a bunch of readers.

LucioFranco (changes requested):
This looks overall correct to me, I think we just need to handle that one unwrap and its good to be merged. Nice work!

PR #892, opened 2024-01-11
Lucio/config session

LucioFranco:


libsql-server/src/namespace/meta_store.rs
```diff
@@ -119,14 +131,17 @@ fn restore(db: &Connection) -> Result<HashMap<NamespaceName, Sender<Arc<Database
...
                         tracing::warn!("unable to convert config: {}", e);
                         continue;
                     }
                 };
 
-                let (tx, _) = watch::channel(Arc::new(config));
+                // We don't store the version in the sqlitedb due to the session token
+                // changed each time we start the primary, this will cause the replica to
+                // handshake again and get the latest config.
+                let (tx, _) = watch::channel(InnerConfig { version: 0, config });
```
MarinPostma, line 144: this is only ok for now, because we keep the config for all namespaces in memory at all time. When we allow the number of namespaces to grow, we'll want to tie the config's lifetime to the namespace's lifetime in memory, which won't work anymore. To keep in mind for when we implement that.

MarinPostma (approved):

PR #897, opened 2024-01-13
Add `test --doc` to `xtask test` and the CI

sanks02:
Check https://github.com/rust-lang/cargo/issues/6669 for context as it seems to be the issue here.

Nextest doesn't have the `--doc` flag, so the only option seems to be to use `cargo test --doc`.

Before:

```
---- libsql/src/lib.rs - (line 29) stdout ----
error[E0432]: unresolved import `libsql::Frames`
 --> libsql/src/lib.rs:32:24
  |
6 | use libsql::{Database, Frames};
  |                        ^^^^^^ no `Frames` in the root
  |
  = help: consider importing one of these items instead:
          libsql::replication::Frames
          libsql_replication::rpc::replication::Frames
          pprof::Frames

error: aborting due to previous error
```


.github/workflows/rust.yml
```diff
@@ -123,6 +123,10 @@ jobs:
     - name: Run tests
       run: cargo nextest run
 
+    - name: Run documentation tests
+      # Nextest doesn't have the `--doc` flag, so we are forced to use `cargo test` here.
+      run: cargo test --doc
+
```
MarinPostma, line 129: can you instead change the recipe in xtask for test [here](https://github.com/tursodatabase/libsql/blob/main/xtask/src/main.rs#L75), and then have the CI run `cargo xtask test`?

sanks02: Done

sanks02:
I see that this has been fixed in `main` as well, so I've updated this PR to only include the changes to CI and xtask.

MarinPostma:
It seems your new CI step has uncovered a failing example ;)

sanks02:
I wanted to update the PR, but was unable to because of https://github.com/tursodatabase/libsql/actions/runs/7621698836/job/20758481022.

PR #902, opened 2024-01-15
Libsql query multiple statements

### LINKED ISSUE #807: Local query accepts multiple statement, but remote does not
The following works with local database, but fails with a remote one:

```
    conn.query("select 1; select 1;", ()).await.unwrap();
```

The behavior needs to be the same in both cases.

Horusiath:
Build on top of #847 to fix #807


LucioFranco:
Lets get #847 merged then I will review this one.

libsql/src/hrana/mod.rs
```diff
@@ -123,9 +123,21 @@ impl<T> Statement<T>
...
+        match parse.next() {
+            None => Err(crate::Error::Misuse(
+                "no SQL statement provided".to_string(),
+            )),
+            Some(Err(e)) => Err(e.into()),
+            Some(Ok(stmt)) => {
+                let inner = Stmt::new(stmt.stmt, want_rows);
+                Ok(Statement { stream, inner })
+            }
+        }
```
MarinPostma, line 140: I think the issue is with the API for execute_batch. It shouldn't take a coma separated list of statements, but something that implement something like `IntoIterator<Item = Into<Query>>`. Query could be implemented for `(String, Params)`, so that parameters can be bound to each statement in the batch

MarinPostma: that's what hrana does. it's libsql client that do not follow those semantics. Hrana exposes batching mechanism that sits on top of sqlite semantics, with sole purpose to reduce network roundtrips

Horusiath: I don't think so. Sending StreamExecuteReq  with single statement `select 1; select 2` returns an error instead of executing the first statement. Same goes for BatchReq with single batch step having multiple SQLite statements inside of it.

MarinPostma: yes, we decided to return an error, rather than failing silently on this one. We just shouldn't support multiple statement per query-string

MarinPostma: the added value is strictly 0

MarinPostma: it's either we have an execute api like https://www.sqlite.org/c3ref/exec.html, but that doesn't support parameters, or we support parameters and not multi-statement strings

MarinPostma: but honnestly this api maybe makes sense in C, but in higher level languages, passing an array of statements is very easy

penberg: @MarinPostma The issue is not with C but other bindings also require the ability to execute multiple statements. For example, Python has this https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.executescript which we need to support.

LucioFranco (approved):

PR #918, opened 2024-01-22
libsql-server: Allow `index_info` pragma

### LINKED ISSUE #900: Broken SQLite API: `PRAGMA index_info` returns `SQL_PARSE_ERROR`
```
alembic 1.13.1
SQLAlchemy 2.0.25
sqlalchemy-libsql 0.1.0
```

I encountered this when trying to run `alembic revision --autogenerate` on my turso instance, it raised the following error:

```
sqlalchemy.exc.DatabaseError: (sqlite3.DatabaseError) SQL_PARSE_ERROR: SQL string could not be parsed: unsupported statement: PRAGMA main.index_info(sqlite_autoindex_userhistory_1);
[SQL: PRAGMA main.index_info(sqlite_autoindex_userhistory_1);]
```

I also tried running it directly, the same error was raised:

```python
import sqlalchemy
engine = sqlalchemy.create_engine(...)
with engine.connect() as conn:
    conn.execute(sqlachemy.text("PRAGMA main.index_info('sqlite_autoindex_userhistory_1');"))
```

tried both with and without the `main.` prefix.

This syntax is part of the [SQLite API](https://www.sqlite.org/pragma.html#pragma_index_info) 

The equivalent `SELECT * FROM pragma_index_info('sqlite_autoindex_userhistory_1');` works, but this is not what used by alembic.

penberg:
Fixes #900

### DIFF
--- a/libsql-server/src/query_analysis.rs
+++ b/libsql-server/src/query_analysis.rs
@@ -121,7 +121,7 @@ impl StmtKind {
         let name = name.name.0.as_str();
         match name {
             // always ok to be served by primary or replicas - pure readonly pragmas
-            "table_list" | "index_list" | "table_info" | "table_xinfo" | "index_xinfo"
+            "table_list" | "index_list" | "table_info" | "table_xinfo" | "index_info" | "index_xinfo"
             | "pragma_list" | "compile_options" | "database_list" | "function_list"
             | "module_list" => Some(Self::Read),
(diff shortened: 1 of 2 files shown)


psarna (changes requested):
A mirror change is needed in libsql/src/parser.rs - client side

psarna (approved):

PR #920, opened 2024-01-22
Go bindings add positional parameters

haaawk:


bindings/c/src/lib.rs
```diff
@@ -205,6 +205,182 @@ pub unsafe extern "C" fn libsql_disconnect(conn: libsql_connection_t) {
...
+) -> std::ffi::c_int {
+    let sql = unsafe { std::ffi::CStr::from_ptr(sql) };
+    let sql = match sql.to_str() {
+        Ok(sql) => sql,
+        Err(e) => {
+            set_err_msg(format!("Wrong SQL: {}", e), out_err_msg);
+            return 1;
+        }
+    };
+    let stmt = Box::leak(Box::new(libsql_stmt { stmt: stmt { sql: sql.to_string(), params: vec![]} }));
```
LucioFranco, line 222: I think you just want to use this https://doc.rust-lang.org/stable/std/boxed/struct.Box.html#method.into_raw instead of leak.

bindings/c/src/lib.rs
```diff
@@ -205,6 +206,187 @@ pub unsafe extern "C" fn libsql_disconnect(conn: libsql_connection_t) {
...
+    };
+    let value_len: usize = match value_len.try_into() {
+        Ok(v) => v,
+        Err(e) => {
+            set_err_msg(format!("Wrong param value len: {}", e), out_err_msg);
+            return 2;
+        }
+    };
+    let value = unsafe { core::slice::from_raw_parts(value, value_len) };
+    let value = Vec::from(value);
```
LucioFranco, line 351: This seems correct to me since its making a copy since we have no control over the allocator thus doing a Vec::from_raw_parts is not going to work.

bindings/c/src/types.rs
```diff
@@ -97,6 +97,57 @@ impl From<&mut libsql_connection> for libsql_connection_t {
...
+    pub fn is_null(&self) -> bool {
+        self.ptr.is_null()
+    }
+
+    pub fn get_ref(&self) -> &stmt {
+        &unsafe { &*self.ptr }.stmt
+    }
+
+    #[allow(clippy::mut_from_ref)]
+    pub fn get_ref_mut(&self) -> &mut stmt {
```
LucioFranco, line 131: I think this needs to be marked unsafe, converting a &_ to a &mut _ is a UB and violates LLVM aliasing. So this likely should return the raw *mut pointer as well and not convert it back into a safe rust &mut

bindings/c/src/lib.rs
haaawk: What's the difference? Box::leak seems to be widely used in this file.

bindings/c/src/types.rs
haaawk: Does that mean all the other get_ref_mut methods in this file are wrong too?

PR #931, opened 2024-01-24
introduce snapshot store

MarinPostma:
Introduce the snapshot store, that manages snapshots for all namespaces. It is backed by a sqlite table, which allows efficient searching. The resource is also separated from the namespaces, so that it's lifetime is not tied to that of the namespaces, and that it is shared.

### DIFF
--- a/libsql-server/src/replication/snapshot_store.rs
+++ b/libsql-server/src/replication/snapshot_store.rs
@@ -1,0 +1,333 @@
+#![allow(dead_code)]
+
+use std::io::{Seek, Write};
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
+use std::sync::Arc;
+
+use libsql_replication::frame::FrameBorrowed;
+use libsql_replication::snapshot::SnapshotFile;
+use libsql_replication::snapshot::SnapshotFileHeader;
+use libsql_sys::wal::{Sqlite3Wal, Sqlite3WalManager};
+use parking_lot::Mutex;
+use rusqlite::named_params;
+use rusqlite::OptionalExtension;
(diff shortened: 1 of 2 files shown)


libsql-server/src/replication/snapshot_store.rs
```diff
@@ -0,0 +1,333 @@
...
+
+        self.snapshot_file.write_all(frame.as_bytes()).unwrap();
+
+        Ok(())
+    }
+
+    pub fn finish(mut self, start_frame_no: FrameNo) -> Result<()> {
+        self.snapshot_file
+            .seek(std::io::SeekFrom::Start(0))
+            .unwrap();
```
LucioFranco, line 241: should we handle this suprrious io errors dont sound like fun

libsql-server/src/replication/snapshot_store.rs
```diff
@@ -0,0 +1,333 @@
...
+            start_frame_no,
+            end_frame_no,
+            snapshot_id,
+            || {
+                self.snapshot_file.persist(&snapshot_path).unwrap();
+                Ok(())
+            },
+        );
+
+        if ret.is_err() {
```
LucioFranco, line 273: worth at least a debug log maybe?

LucioFranco (approved):
LGTM

libsql-server/src/replication/snapshot_store.rs
MarinPostma: we sure should, but for now I'm only interested in having the skeleton code in tree, so that I can iron out the storage layer with @Horusiath

PR #937, opened 2024-01-24
Add UUID to stats

athoscouto:
This should allow any stats consumer to know when the file has been lost and the counters should be restarted.

Replaces https://github.com/tursodatabase/libsql/pull/883.

### DIFF
--- a/libsql-server/src/stats.rs
+++ b/libsql-server/src/stats.rs
@@ -8,6 +8,7 @@ use std::collections::BTreeSet;
 use tokio::io::AsyncWriteExt;
 use tokio::task::JoinSet;
 use tokio::time::Duration;
+use uuid::Uuid;
 
 use crate::namespace::NamespaceName;
 use crate::replication::FrameNo;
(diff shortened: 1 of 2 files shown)


libsql-server/src/stats.rs
```diff
@@ -56,6 +57,8 @@ pub struct Stats {
     #[serde(skip)]
     namespace: NamespaceName,
 
+    #[serde(default)]
+    id: Uuid,
```
MarinPostma, line 61:
```suggestion
    #[serde(default)]
    id: Option<Uuid>,
```

This is how we would do it in rust

libsql-server/src/stats.rs
```diff
@@ -93,6 +96,10 @@ impl Stats {
             Stats::default()
         };
 
+        if this.id.is_nil() {
+            this.id = Uuid::new_v4();
+        }
+
```
MarinPostma, line 102:
```suggestion
        if this.id.is_none() {
            this.id = Some(Uuid::new_v4());
        }

```

PR #940, opened 2024-01-24
libsql: add local sync write delegation

LucioFranco:


libsql-server/tests/embedded_replica/local.rs
```diff
@@ -0,0 +1,144 @@
...
+
+        tracing::info!("executing write delegated inserts");
+
+        for _ in 0..300 {
+            conn.execute("insert into test values (randomblob(4092))", ())
+                .await
+                .unwrap();
+        }
+
+        let snapshots_path = tmp_host_path.join("dbs").join("foo").join("snapshots");
```
lmalheiro, line 102: Wouldn't be interesting to execute another query before syncing to show that the local DB hasn't changed, even though the inserts were successful?

MarinPostma (approved):

libsql-server/tests/embedded_replica/local.rs
LucioFranco: Good call, added!

PR #942, opened 2024-01-24
chore: fix ffi build script again

LucioFranco:


libsql-ffi/build.rs
```diff
@@ -222,7 +228,17 @@ pub fn build_bundled(out_dir: &str, out_path: &Path) {
     println!("cargo:lib_dir={out_dir}");
 }
 
-fn build_multiple_ciphers(out_dir: &str, out_path: &Path) {
+fn copy_multiple_ciphers(out_dir: &str) {
```
LucioFranco, line 231: We need to separate the code that copies the needed linkable files into the out dir and the code that actually builds it. We originally had a cyclic issue where we were building code that then was placed right back into the bundled folder, which caused the build script to re-run and execute the same build. This cuts down recompiles from 45s to 0.14s.

PR #953, opened 2024-01-26
WIP: Custom page allocator for sqlite

MarinPostma:
In my quest to get memory under control, I'm experimenting with using a custom pager cache for sqlite connections.

All pages are allocated at the process startup, and the pool is shared among all the connections.

sqlite use of global sucks.

This doesn't work yet


libsql-server/src/pager.rs
```diff
@@ -0,0 +1,319 @@
...
+    let cache = unsafe { &mut *(cache as *mut Pager) };
+    let _new_page = unsafe { &mut *(data as *mut Page) };
+    let pages = &mut cache.pages;
+    let page = pages.remove(&old_key).expect("missing page when rekeying");
+    page.pin(new_key as u16);
+    if let Some(page) = pages.insert(new_key, page) {
+        assert!(!page.is_pinned());
+        cache.alloc.free(page);
+    }
+}
```
MarinPostma, line 252: @psarna It's not clear to me what I'm supported to do with new page. https://sqlite.org/c3ref/pcache_methods2.html suggests that we are just moving a page, but then why is that necessary to pass the data as well?

MarinPostma: my undertanding is that data is a ref to the old key's value already

psarna: let me browse the C code and see

psarna: @MarinPostma ok, so the stock implementation is here: https://github.com/tursodatabase/libsql/blob/ac6bf3a4a9eca21e004654b82a04149b1835850b/libsql-sqlite3/src/pcache1.c#L1113-L1147

... and I don't follow the question about why we need to pass the page data -- we need its metadata, which in the stock implementation is really a pointer to https://github.com/tursodatabase/libsql/blob/ac6bf3a4a9eca21e004654b82a04149b1835850b/libsql-sqlite3/src/pcache1.c#L117-L127 . `PgHdr1` has an intrusive hashtable built-in, so when we update the key and put it in another place in the data structure, we need the page instance to update the internal pointers. So I guess it's just an implementation detail, and since your hashmap isn't intrusive, you just don't need to touch the page pointer either.

MarinPostma:
@sarna, do you think it's safe to assume a pager per connection? Therefore no need for synchronization when accessing the pager?

psarna:
@MarinPostma no, multiple connections may share a `BtShared` object, which contains the pager:

`sqlite3` has a table of databases: https://github.com/tursodatabase/libsql/blob/b01500237f2a1cd137681ef1d86edf482e848c29/libsql-sqlite3/src/sqliteInt.h#L1658
each database has a btree: https://github.com/tursodatabase/libsql/blob/b01500237f2a1cd137681ef1d86edf482e848c29/libsql-sqlite3/src/sqliteInt.h#L1427
there's a shared part of it: https://github.com/tursodatabase/libsql/blob/b01500237f2a1cd137681ef1d86edf482e848c29/libsql-sqlite3/src/btreeInt.h#L332-L347
and here's the pager: https://github.com/tursodatabase/libsql/blob/b01500237f2a1cd137681ef1d86edf482e848c29/libsql-sqlite3/src/btreeInt.h#L425-L426

MarinPostma:
uhh that sucks
ok I'll add a lock

psarna:
Also depends on the access pattern, is it from our code or some callbacks from within SQLite? In the latter case we can often assume we already hold locks

libsql-server/src/pager.rs
```diff
@@ -0,0 +1,344 @@
...
+            extra_size as usize,
+            PAGER_CACHE_SIZE.load(SeqCst),
+        )
+    });
+    let pager = Pager {
+        alloc: allocator,
+        pages: HashMap::new().into(),
+    };
+
+    Box::leak(Box::new(pager)) as *mut Pager as *mut _
```
LucioFranco, line 191: into_raw reads better here

LucioFranco (approved):

PR #954, opened 2024-01-27
Update example.rs

evccyr:
Remove await due to this error:
```rs
Diagnostics:
1. `Result<Option<Row>, libsql::Error>` is not a future
   the trait `Future` is not implemented for `Result<Option<Row>, libsql::Error>`
   Result<Option<Row>, libsql::Error> must be a future or must implement `IntoFuture` to be awaited
   required for `Result<Option<Row>, libsql::Error>` to implement `IntoFuture` [E0277]
2. remove the `.await` [E0277]
```

### DIFF
--- a/libsql/examples/example.rs
+++ b/libsql/examples/example.rs
@@ -35,7 +35,7 @@ async fn main() {
 
     let mut rows = stmt.query(["foo@example.com"]).await.unwrap();
 
-    let row = rows.next().await.unwrap().unwrap();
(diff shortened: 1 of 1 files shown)


psarna:
@evccyr how did you run the example? This api is asynchronous, and works just fine with `cargo run --example example`

LucioFranco:
its likely that the version in main is more updated than the version you are using? We will be making a round of client releases quite soon that will probably include this breaking change.

PR #963, opened 2024-01-29
limit the number of retries of the busy handler

MarinPostma:
This PR fixes a series of bugs that caused hangs on the platform.

- Replace `is_stolen` `AtomicBool` by a `Mutex<bool>`. I'm not sure this one was necessary, but I was desperate.
- Always update the state for a connection, even in case of error during execution.
- An instance of `borrow_and_update` in a channel was borrowing too long, causing a dead_lock. This was caused by some obscure rules about temporaries, causing the reference to be dropped after the call, even though it was being dereferenced. This is best exemplified by https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3c7bf7fbc2471c9bce121c781ebc982b, and elements of explanation can be found here: https://blog.m-ou.se/super-let/

### DIFF
--- a/libsql-server/src/connection/libsql.rs
+++ b/libsql-server/src/connection/libsql.rs
@@ -649,8 +649,9 @@ impl<W: Wal> Connection<W> {
         {
             let mut lock = this.lock();
             let is_autocommit = lock.conn.is_autocommit();
+            let current_fno = *lock.current_frame_no_receiver.borrow_and_update();
             builder.finish(
-                *(lock.current_frame_no_receiver.borrow_and_update()),
+                current_fno,
                 is_autocommit,
             )?;
         }
--- a/libsql-server/src/connection/libsql.rs
+++ b/libsql-server/src/connection/libsql.rs
@@ -602,13 +602,7 @@ impl<W: Wal> Connection<W> {
         pgm: Program,
         mut builder: B,
     ) -> Result<B> {
-        use rusqlite::TransactionState as Tx;
-
-        let (state, txn_timeout) = {
-            let lock = this.lock();
-            let txn_timeout = lock.config_store.get().txn_timeout.unwrap_or(TXN_TIMEOUT);
-            (lock.state.clone(), txn_timeout)
-        };
+        let txn_timeout = this.lock().config_store.get().txn_timeout.unwrap_or(TXN_TIMEOUT);
 
         let mut results = Vec::with_capacity(pgm.steps.len());
         builder.init(&this.lock().builder_config)?;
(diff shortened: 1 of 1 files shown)


libsql-server/src/connection/libsql.rs
```diff
@@ -621,61 +642,71 @@ impl<W: Wal> Connection<W> {
...
+            // lock was downgraded, notify a waiter
+            (Tx::Write, Tx::None | Tx::Read) => {
+                let old_slot = self
+                    .slot
+                    .take()
+                    .expect("there should be a slot right after downgrading a txn");
+                let mut maybe_state_slot = self.state.slot.write();
+                // We need to make sure that the state slot is our slot before removing it.
+                if let Some(ref state_slot) = *maybe_state_slot {
+                    if Arc::ptr_eq(state_slot, &old_slot) {
```
LucioFranco, line 694: is there a case where a pointer gets reused and we run into a ABA problem?

LucioFranco (approved):

libsql-server/src/connection/libsql.rs
MarinPostma: no, we discussed that last time, but we're actually comparing the pointer with itself, and we're still holding a reference to both ends while comparing, so no risk of it being reused in the meantime :)

libsql-server/src/connection/libsql.rs
```diff
@@ -493,7 +492,17 @@ unsafe extern "C" fn busy_handler<T: Wal>(state: *mut c_void, retries: c_int) ->
                         if Arc::ptr_eq(s, &slot) {
                             // We check that slot wasn't already stolen, and that their is still a slot.
                             // The ordering is relaxed because the atomic is only set under the slot lock.
```
haaawk, line 494: This comment can probably go away, no?

libsql-server/src/connection/libsql.rs
```diff
@@ -612,12 +621,17 @@ impl<W: Wal> Connection<W> {
...
-                    has_timeout = true;
+            if !has_timeout {
+                if let Some(slot) = &lock.slot {
+                    let mut is_stolen = slot.is_stolen.lock();
+                    if *is_stolen || Instant::now() > slot.expires_at() {
+                        // we mark ourselves as stolen to notify any waiting lock thief.
+                        if !*is_stolen {
+                            lock.rollback();
+                        }
+                        *is_stolen = true;
```
haaawk, line 630: Shouldn't this be in opposite order?
```
let stolen = *is_stolen
*is_stolen = true;
if !stolen {
  lock.rollback();
}
```

haaawk (approved):

psarna (approved):
looks good, later we can also push a low priority follow-up that brings back `AtomicBool`, since it's probably just as good -- and test it on canary

PR #964, opened 2024-01-29
test ends with SIGSEGV

Horusiath:
Running this test locally ends with:

```
Caused by:
  process didn't exit successfully: `libsql/target/debug/deps/tests-937c14199c1f7a68 'hrana::batch::execute_results' --format=json --exact -Z unstable-options --show-output` (signal: 11, SIGSEGV: invalid memory reference)
```

Stack trace: 

image

With `encryption` feature turned off, it doesn't cause problems.


psarna:
(also, fyi, it only happens to fail on Apple arm64 hardware)

libsql-sqlite3/src/build.c
```diff
@@ -189,16 +189,19 @@ void sqlite3FinishCoding(Parse *pParse){
     }
     sqlite3VdbeAddOp0(v, OP_Halt);
 
-#if SQLITE_USER_AUTHENTICATION
```
psarna, line 192: the `#if SQLITE_USER_AUTHENTICATION` must stay, it's still valid

MarinPostma:
can we close that?

PR #972, opened 2024-01-31
scripted replication

### LINKED ISSUE #508: Snapshot script exiting with 1 causes compation to fail later
If a snapshot script exits with code 1, log compaction eventually fails, crashing the log compactor thread.

MarinPostma:
This is a revision of scripted backups. This code was left unused for a long time, and it started to rot a bit, so I refreshed it with some improvements.

- The script is called with 4 arguments:
    - The path where the snapshot can be found
    - The namespace for that snapshot
    - The starting replication index
    - The end replication index

- The path passed as an argument is a hard link to the snapshot file. It is expected that the script removes it after the snapshot was ingested. Sqld will check for the existence of the snapshot file, on success, if the file is still present, this is considered, it will panic. On error, if the snapshot file *is not present* sqld will also panic. On restart, the leftover snapshots will be passed again to the script. The script must, therefore, ensure idempotency.
- If the script file, with the assumption above, holds true, then sqld will re-enqueue the script for retry, with an exponential backoff.
- Snapshots for different namespaces are processed in an undefined order
- Snapshots for the same namespace are always processed in an order such that the first snapshot to be processed is that which has the lowest starting replication index. On ties, that which covers the most replication indexed is processed first, allowing the script to discard the smaller one if that's what they want to do.

## Testing:
I have tested this feature with the following script:
```bash
#!/bin/sh

SUCCESS=1

echo "$1 $2 $3 $4 $SUCCESS" >> log

if [ $SUCCESS == 1 ]
then
    rm $1
else
    exit 1
fi
```

start sqld with: `cargo r -- --snapshot-exec ./script.sh  --max-log-duration 5`
and while perform insertion with:
```
repeat 10000 echo '{"requests":[{"type": "execute", "stmt": {"sql": "insert into test values (12)"}}]}'| http post localhost:8080/v2/pipeline
```
flip the `SUCCESS` flag from 0 to 1 randomly to simulate failure.

here is a sample log:
```
data.sqld/script_backup/default:00000000000000000000-00000000000000000001.snap default 0 1 1
data.sqld/script_backup/default:00000000000000000002-00000000000000000002.snap default 2 2 1
data.sqld/script_backup/default:00000000000000000003-00000000000000000027.snap default 3 39 1
data.sqld/script_backup/default:00000000000000000000-00000000000000000027.snap default 0 39 1
data.sqld/script_backup/default:00000000000000000028-0000000000000000004c.snap default 40 76 1
data.sqld/script_backup/default:0000000000000000004d-00000000000000000071.snap default 77 113 0
data.sqld/script_backup/default:00000000000000000000-00000000000000000071.snap default 0 113 0
data.sqld/script_backup/default:00000000000000000000-00000000000000000071.snap default 0 113 0
data.sqld/script_backup/default:00000000000000000000-00000000000000000071.snap default 0 113 0
data.sqld/script_backup/default:00000000000000000000-00000000000000000071.snap default 0 113 0
data.sqld/script_backup/default:00000000000000000000-00000000000000000071.snap default 0 113 1
data.sqld/script_backup/default:0000000000000000004d-00000000000000000071.snap default 77 113 1
data.sqld/script_backup/default:00000000000000000072-00000000000000000095.snap default 114 149 1
data.sqld/script_backup/default:00000000000000000096-000000000000000000ba.snap default 150 186 1
data.sqld/script_backup/default:00000000000000000000-000000000000000000ba.snap default 0 186 1
data.sqld/script_backup/default:000000000000000000bb-000000000000000000de.snap default 187 222 1
data.sqld/script_backup/default:000000000000000000df-00000000000000000103.snap default 223 259 1
data.sqld/script_backup/default:00000000000000000000-00000000000000000103.snap default 0 259 1
data.sqld/script_backup/default:00000000000000000104-00000000000000000128.snap default 260 296 1
data.sqld/script_backup/default:00000000000000000129-0000000000000000014d.snap default 297 333 1
data.sqld/script_backup/default:00000000000000000000-0000000000000000014d.snap default 0 333 1
data.sqld/script_backup/default:0000000000000000014e-00000000000000000172.snap default 334 370 1
```

We can see that after failure, the snapshots are retried, as expected.

close #508

### DIFF
--- a/libsql-server/src/replication/script_backup_manager.rs
+++ b/libsql-server/src/replication/script_backup_manager.rs
@@ -381,7 +381,10 @@ mod test {
         impl Handler for FailHandler {
             async fn handle(&mut self, entry: &SnapshotEntry) -> Result<()> {
                 self.last_entry = Some(entry.clone());
-                Err(Error::HandlerFailure)
+                Err(Error::HandlerFailure {
+                    stdout: String::new(),
+                    stderr: String::new(),
+                })
             }
         }
 
@@ -460,7 +463,10 @@ mod test {
         impl Handler for FailHandler {
             async fn handle(&mut self, entry: &SnapshotEntry) -> Result<()> {
                 tokio::fs::remove_file(&entry.path).await.unwrap();
-                Err(Error::HandlerFailure)
+                Err(Error::HandlerFailure {
+                    stdout: String::new(),
+                    stderr: String::new(),
+                })
             }
         }
 
--- a/libsql-server/src/replication/script_backup_manager.rs
+++ b/libsql-server/src/replication/script_backup_manager.rs
@@ -78,6 +78,7 @@ impl Ord for SnapshotEntry {
     }
 }
 
+#[derive(Clone)]
 pub(crate) struct ScriptBackupManager {
     path: PathBuf,
     queue: Arc<Mutex<BinaryHeap<SnapshotEntry>>>,
--- a/libsql-server/src/replication/script_backup_manager.rs
+++ b/libsql-server/src/replication/script_backup_manager.rs
@@ -1,0 +1,429 @@
+use std::cmp::Ordering;
+use std::sync::Arc;
+use std::path::{Path, PathBuf};
+use std::collections::BinaryHeap;
+
+use once_cell::sync::Lazy;
+use parking_lot::Mutex;
+use regex::Regex;
+use tokio::sync::Notify;
+use tokio::time::Duration;
+
+use crate::namespace::NamespaceName;
+
+use super::FrameNo;
+
+const MAX_RETRIES_THRESHOLD: u32 = 64;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("io error: {0}")]
+    Io(#[from] std::io::Error),
+    #[error("Snaphot handler failure")]
+    HandlerFailure,
+    #[error("Could not parse snapshot path: {0}")]
+    InvalidSnapshotPath(PathBuf),
+}
+
+type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, Clone)]
+pub(crate) struct SnapshotEntry {
+    namespace: NamespaceName,
+    start_frame_no: FrameNo,
+    end_frame_no: FrameNo,
+    path: PathBuf,
+    retries: u32,
+}
+
+pub(crate) trait Handler {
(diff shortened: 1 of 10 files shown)


libsql-server/src/replication/script_backup_manager.rs
```diff
@@ -0,0 +1,573 @@
...
+                            assert!(entry.path.try_exists()?, "snapshot file was removed, but script returned an error. Can't ensure consistency");
+                            // exponential backoff
+                            tokio::time::sleep(
+                                Duration::from_millis(500) * 2u32.pow(entry.retries),
+                            )
+                            .await;
+
+                            entry.retries += 1;
+                            if entry.retries > MAX_RETRIES_THRESHOLD {
+                                todo!("failure to make any progress, what do we do?");
```
LucioFranco, line 153: panic crash, log each retry? I think anything here is better than a todo 😄

libsql-server/src/replication/script_backup_manager.rs
```diff
@@ -0,0 +1,573 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use proptest::prelude::*;
+    use tempfile::tempdir;
+    use uuid::Uuid;
+
+    proptest! {
```
LucioFranco, line 272: this is sexy

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
 profile = "default"
-channel = "1.73.0"
+channel = "1.75.0"
```
LucioFranco, line 3: probably should do this in a diff pr but its fine

LucioFranco (approved):
LGTM only concern is that one todo!

libsql-server/src/replication/script_backup_manager.rs
```diff
@@ -0,0 +1,573 @@
...
+    }
+
+    async fn process_one(&mut self) -> crate::Result<()> {
+        loop {
+            let entry = self.queue.lock().pop();
+            match entry {
+                Some(mut entry) => {
+                    match self.handler.handle(&entry).await {
+                        Ok(_) => {
+                            assert!(!entry.path.try_exists()?, "snapshot handler returned success, yet snapshot file is still present.");
```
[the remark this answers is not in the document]
MarinPostma: This one is an actual assertion, it's an implementation error if the script fails to do that,and would only caused crippled program. Better abort

PR #975, opened 2024-01-31
chore: update cargo-dist to 0.8.2

Gankra:
steps taken:
* remove allow-dirty from config
* cargo dist init
* rm ./.github/workflows/release-libsql-server.yml
* mv ./.github/workflows/release.yml ./.github/workflows/release-libsql-server.yml
* add libsql-server- prefix back to on:push:tags:
* add back allow-dirty

Note that this is applying a seemingly-pending change to your release notes. Previously you had set things up to use release-drafter, and `create-release=false` was set. Your config has a lingering `create-release=true` setting change.

By accepting this change, **it will now be an error for a draft release to exist**. cargo-dist will expect to create a release from scratch, and will use the changelogs it finds in your CHANGELOG/RELEASES files.


LucioFranco (approved):
Thanks!

PR #994, opened 2024-02-05
Allow vacuum block to be configured with SQLD_BLOCK_VACUUM env var

athoscouto:
This would allow `libsql-server` users to run vacuum locally.

### DIFF
--- a/libsql-server/src/main.rs
+++ b/libsql-server/src/main.rs
@@ -228,6 +228,9 @@ struct Cli {
     /// encryption_key for encryption at rest
     #[clap(long, env = "SQLD_ENCRYPTION_KEY")]
     encryption_key: Option<bytes::Bytes>,
+    /// encryption_key for encryption at rest
+    #[clap(long, env = "SQLD_BLOCK_VACUUM")]
+    block_vacuum: bool,
 
     #[clap(long, default_value = "128")]
     max_concurrent_connections: usize,
@@ -348,6 +351,7 @@ fn make_db_config(config: &Cli) -> anyhow::Result<DbConfig> {
         checkpoint_interval: config.checkpoint_interval_s.map(Duration::from_secs),
         snapshot_at_shutdown: config.snapshot_at_shutdown,
         encryption_key: config.encryption_key.clone(),
+        block_vacuum: config.block_vacuum,
     })
 }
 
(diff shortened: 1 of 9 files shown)


libsql-server/src/main.rs
```diff
@@ -228,6 +228,9 @@ struct Cli {
     /// encryption_key for encryption at rest
     #[clap(long, env = "SQLD_ENCRYPTION_KEY")]
     encryption_key: Option<bytes::Bytes>,
+    /// encryption_key for encryption at rest
+    #[clap(long, env = "SQLD_BLOCK_VACUUM")]
+    block_vacuum: bool,
```
psarna, line 233: Doesn't it make it default to `false`, unless you specify otherwise? It may make sense to introduce a `SQLD_ALLOW_USER_VACUUM` and make it default to false everywhere, and that would work naturally for all kinds of libraries, rpc, serializers, etc. Main risk of accidentally allowing vacuum here is that it's a very I/O heavy operation, and we should prefer to always prevent users from issuing it themselves, or at least throttle it

libsql-server/src/query_analysis.rs
```diff
@@ -108,6 +109,7 @@ impl StmtKind {
             }) => Some(Self::Write),
             Cmd::Stmt(Stmt::DropView { .. }) => Some(Self::Write),
             Cmd::Stmt(Stmt::Savepoint(_)) => Some(Self::Savepoint),
+            Cmd::Stmt(Stmt::Vacuum(_, _)) => Some(Self::Vacuum),
```
psarna, line 112: there's a mirror parser in our client code - `libsql/src/parser.rs`, and it should get updated in there as well

glommer:
@athoscouto do you plan to rebase this after sarna's comments?

libsql-server/src/main.rs
```diff
@@ -228,6 +228,9 @@ struct Cli {
     /// encryption_key for encryption at rest
     #[clap(long, env = "SQLD_ENCRYPTION_KEY")]
     encryption_key: Option<bytes::Bytes>,
+    /// encryption_key for encryption at rest
```
avinassh, line 231:
```suggestion
    /// config to allow running vacuum
```

penberg:
@athoscouto I am closing this because there has been no activity on the pull request for six months and it has tons of conflicts. Please open a new one if you still want to do this.

PR #995, opened 2024-02-06
libsql: add periodic background sync

LucioFranco:
This adds a new periodic sync configuration item to the `Builder` that will allow users to set a duration at which they want the background sync to sync.

### DIFF
--- a/libsql/src/replication/mod.rs
+++ b/libsql/src/replication/mod.rs
@@ -2,6 +2,7 @@
 
 use std::path::PathBuf;
 use std::sync::Arc;
+use std::time::Duration;
 
 pub use libsql_replication::frame::{Frame, FrameNo};
 use libsql_replication::replicator::{Either, Replicator};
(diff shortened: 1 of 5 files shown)


libsql/src/replication/mod.rs
```diff
@@ -95,27 +98,76 @@ impl Writer {
...
+                        }
+
+                        tokio::time::sleep(sync_duration).await;
+                    }
+                }
+                .instrument(tracing::info_span!("periodic_sync")),
+            );
+
+            replicator.bg_abort = Some(Arc::new(DropAbort(jh.abort_handle())));
+        }
```
MarinPostma, line 145: would it make sense to spawn this task in a join set instead?

LucioFranco: That is an option, I think the behavior is the same here just saves us some code. So maybe we can refactor to that in the future.

PR #1026, opened 2024-02-13
Introduce UserAuthStrategy to allow third party authentication implementation

jameswritescode:
## What are you trying to accomplish?

This PR seeks to open up authentication so that anyone using libsql-server as a library can provide their own authentication strategy.

## What approach did you choose and why?

We removed the `Auth` struct in favor of structs that implement `UserAuthStrategy` that can be configured via `UserApiConfig`. The existing strategies of HTTP Basic and JWT have been moved into their own strategy definitions. Anywhere `Auth` was previously being call for authentication has been switched out with `UserAuthStrategy#authenticate`.

## What should reviewers focus on?

* A handful of the changes in the `auth` mod with exception of the auth strategies are existing code from the original `auth.rs` but put into their own files.
* We would like to follow up in the future with changes to `AuthError` to provide more generic enum variants with data to give auth strategies more control over their errors.
* How do we feel about the changes in `hrana::ws`? How might we approach this differently otherwise?


penberg:
Looks good to me, but @MarinPostma or @LucioFranco ought to review this before merging.

MarinPostma:
I need a bit of time to look at this, but will do asap

libsql-server/src/auth/errors.rs
```diff
@@ -0,0 +1,52 @@
+use thiserror;
```
LucioFranco, line 1:
```suggestion
```
shouldn't need this with rust 2018 edition

libsql-server/src/hrana/ws/mod.rs
```diff
@@ -22,7 +22,7 @@ mod session;
 
 struct Server<F: MakeNamespace> {
     namespaces: NamespaceStore<F>,
-    auth: Arc<Auth>,
+    user_auth_strategy: Arc<Box<dyn UserAuthStrategy>>,
```
LucioFranco, line 25: An idea: We could potentially replace `Arc<Auth>` with `Auth` that internally holds an `Arc<dyn UserAuthStrategy>` that way we don't leak the internal implementation details (and there are no generics). This I think would then mean the rest of the code base just gets a `Arc<Auth>` -> `Auth` conversion and it should be really clean.

libsql-server/src/hrana/ws/session.rs
```diff
@@ -72,11 +74,21 @@ pub(super) async fn handle_initial_hello<F: MakeNamespace>(
...
+
     let authenticated = server
-        .auth
-        .authenticate_jwt(jwt.as_deref(), server.disable_namespaces, namespace_jwt_key)
+        .user_auth_strategy
+        .authenticate(UserAuthContext {
+            namespace,
+            user_credential,
+            namespace_credential: namespace_jwt_key,
+        })
```
LucioFranco, line 91: What are the thoughts on maybe having this function accept these each as individual parameters rather than a struct? I feel like this could reduce some of the noise here.

libsql-server/src/hrana/ws/session.rs
```diff
@@ -102,12 +114,23 @@ pub(super) async fn handle_repeated_hello<F: MakeNamespace>(
     }
     let namespace_jwt_key = server
         .namespaces
-        .with(namespace, |ns| ns.jwt_key())
+        .with(namespace.clone(), |ns| ns.jwt_key())
         .await??;
+
+    // Convert jwt token into a HeaderValue to be compatible with UserAuthStrategy
+    let user_credential = jwt
```
LucioFranco, line 121: unfortunate but I think this is fine for now, WS is not our primary protocol anyways for hrana so this I think is acceptable for now.

I think the only thing this really poses is should we accept a String rather than a HeaderValue in the auth trait?

LucioFranco (changes requested):
Overall this seems like a great improvement, I left a few comments/questions regarding some style things and api designs. Let me know if something isn't clear! thanks!

libsql-server/src/config.rs
```diff
@@ -60,8 +60,7 @@ pub struct UserApiConfig<A = AddrIncoming> {
     pub http_acceptor: Option<A>,
     pub enable_http_console: bool,
     pub self_url: Option<String>,
-    pub http_auth: Option<String>,
-    pub auth_jwt_key: Option<String>,
+    pub auth_strategy: Arc<Box<dyn UserAuthStrategy>>,
```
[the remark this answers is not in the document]
jameswritescode: This was my misunderstanding, I'd started by passing a `Box` and then brought in the `Arc`, didn't realize that it eliminated the need for `Box`. Thanks for pointing it out.

libsql-server/src/hrana/ws/session.rs
jameswritescode: I like the idea of `user_credential` being an `Option<String>` though a couple of things off the top of my head I think we'd need to account for:

* In HTTP context, do we pass the entire `Authorization` header value?
* In this WS context, do we still append an Authorization scheme or does the strategy need to do some extra homework to determine if it's being given a header value or something else?

libsql-server/src/hrana/ws/session.rs
jameswritescode: I think having a struct gives more flexibility of the shape of `UserAuthContext` long term with less impact on strategy implementations if it did unless we removed fields. I am open to moving these to parameters.

libsql-server/src/auth/user_auth_strategies/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+use super::{AuthError, Authenticated};
+
+pub struct UserAuthContext {
+    pub namespace: NamespaceName,
+    pub namespace_credential: Option<jsonwebtoken::DecodingKey>,
+    pub user_credential: Option<HeaderValue>,
+}
+
+pub trait UserAuthStrategy: Sync + Send {
+    fn authenticate(&self, context: UserAuthContext) -> Result<Authenticated, AuthError>;
```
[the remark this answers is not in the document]
jameswritescode: If we keep the struct around, can change!

libsql-server/src/hrana/ws/session.rs
LucioFranco: Okay I see, it would break changes to the consumer of this struct (which is just internal so that is okay) but wouldn't be a breaking change (to add new fields) to implementers which could exist outside of this code base. This makes sense to me!

libsql-server/src/auth/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+pub mod user_auth_strategies;
+
+pub use authenticated::Authenticated;
+pub use authorized::Authorized;
+pub use errors::AuthError;
+pub use parsers::{parse_http_auth_header, parse_http_basic_auth_arg, parse_jwt_key};
+pub use permission::Permission;
+pub use user_auth_strategies::{Disabled, HttpBasic, Jwt, UserAuthContext, UserAuthStrategy};
+
+pub type Auth = Arc<dyn UserAuthStrategy>;
```
LucioFranco, line 18: I think we can do this in a follow up and the code change should be pretty small, but I would like this to be a struct since type aliases tend to leak the internal implementation detail. So it could be a struct like:

```
#[derive(Clone)]
pub struct Auth {
    inner: Arc<dyn UserAuthStrategy + Send + Sync>,
}

impl Auth {
    // implement functions that we need to be public + constructor via `impl UserAuthStrategy`
}
```

libsql-server/src/hrana/ws/session.rs
LucioFranco: I think it would make sense to strip the prefix at the http/ws level and then only pass through the "token" to the Auth layer? I feel like that would make the most sense. What do you think? That way we can just set this as a string/bytes and not leak HeaderValue through to the auth layer.

LucioFranco (commented):
Overall looks good left a question and some feedback, we can merge this and then do that as follow up or yall could implement those changes and we can merge up to you.

libsql-server/src/auth/mod.rs
jeremywrowe: I'd like to take care of this in this pull request. I'll take a stab at it right now :D

jeremywrowe: This is taken care of here [ad18ef0](https://github.com/tursodatabase/libsql/pull/1026/commits/ad18ef0f56d653b68c0270a94a452ce23e5cf766)

MarinPostma:
folks, since @LucioFranco only has nits, and I need to do auth work, I'm merging this now, I'll let you address the rest in followups :)

PR #1038, opened 2024-02-16
Make encryption cipher configurable and switch default to SQLCipher

penberg:
This patch switches the default cipher to SQLCipher, which seems like a more robust default. However, as libSQL with wxSQLite3 AES 256 cipher is already out in the wild, let's make the cipher configurable so people can read their existing databases...


libsql/src/database.rs
```diff
@@ -4,6 +4,8 @@ mod builder;
 
 pub use builder::Builder;
 
+pub use libsql_sys::{Cipher, EncryptionConfig};
```
MarinPostma, line 7:
```suggestion
#[cfg(feature = "core")]
pub use libsql_sys::{Cipher, EncryptionConfig};
```

PR #1040, opened 2024-02-16
recover metastore from fs

MarinPostma:
To help migration to 0.23, this pr adds the flag `--allow-metastore-recovery`: if this flag is set, and the metastore is empty, then the metastore is seeded from the config files found in the `/dbs` subdirectories.

### DIFF
--- a/libsql-server/src/main.rs
+++ b/libsql-server/src/main.rs
@@ -17,8 +17,8 @@ use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::Layer;
 
 use libsql_server::config::{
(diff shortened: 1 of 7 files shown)


libsql-server/src/main.rs
```diff
@@ -234,6 +234,11 @@ struct Cli {
     // max number of concurrent requests across all connections
     #[clap(long, default_value = "128", env = "SQLD_MAX_CONCURRENT_REQUESTS")]
     max_concurrent_requests: u64,
+
+    /// Allow meta store to recover config from filesystem from older version, if meta store is
+    /// empty on startup
+    #[clap(long)]
```
LucioFranco, line 240: should we add an env var for this?

LucioFranco (approved):

PR #1047, opened 2024-02-19
Go bindings: separate query and exec

haaawk:


haaawk:
Currently fails with:
```
libsql_test.go:34: failed to execute query CREATE TABLE test_5698375202114199610_20240216101942 (a int, b int)
        error code = 2: Error executing statement: Hrana: `api error: `Cannot deserialize client message from JSON: invalid type: null, expected a string representing an integer, or null at line 1 column 275``
```

PR #1049, opened 2024-02-19
Parametrize namespace creation for shared schema db

Horusiath:
HTTP POST for `/v1/namespaces/:namespace/create` now accepts 2 new params: 
- `shared_schema: bool` to mark that this namespace is going to work as shared schema
- `shared_schema_name: Option<String>` to reference shared schema namespace for databases using it

### DIFF
--- a/libsql-replication/proto/metadata.proto
+++ b/libsql-replication/proto/metadata.proto
@@ -16,4 +16,6 @@ message DatabaseConfig {
     optional string jwt_key = 7;
     optional uint64 txn_timeout_s = 8;
     bool allow_attach = 9;
+    bool shared_schema = 10;
+    optional string shared_schema_name = 11;
 }
--- a/libsql-replication/src/generated/metadata.rs
+++ b/libsql-replication/src/generated/metadata.rs
@@ -23,4 +23,8 @@ pub struct DatabaseConfig {
     pub txn_timeout_s: ::core::option::Option<u64>,
     #[prost(bool, tag = "9")]
     pub allow_attach: bool,
+    #[prost(bool, tag = "10")]
+    pub is_shared_schema: bool,
+    #[prost(string, optional, tag = "11")]
+    pub shared_schema_name: ::core::option::Option<::prost::alloc::string::String>,
 }
--- a/libsql-server/src/connection/config.rs
+++ b/libsql-server/src/connection/config.rs
@@ -22,6 +22,10 @@ pub struct DatabaseConfig {
     pub txn_timeout: Option<Duration>,
     #[serde(default)]
     pub allow_attach: bool,
+    #[serde(default)]
+    pub is_shared_schema: bool,
+    #[serde(default)]
+    pub shared_schema_name: Option<String>,
 }
 
 const fn default_max_size() -> u64 {
@@ -40,6 +44,8 @@ impl Default for DatabaseConfig {
             jwt_key: None,
             txn_timeout: Some(TXN_TIMEOUT),
             allow_attach: false,
+            is_shared_schema: false,
+            shared_schema_name: None,
         }
     }
 }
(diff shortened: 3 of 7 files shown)


libsql-replication/proto/metadata.proto
```diff
@@ -16,4 +16,6 @@ message DatabaseConfig {
     optional string jwt_key = 7;
     optional uint64 txn_timeout_s = 8;
     bool allow_attach = 9;
+    optional bool shared_schema = 10;
```
haaawk, line 19: Why not just bool with default false instead of optional?

haaawk (approved):
LGTM but @MarinPostma and @LucioFranco should probably have a look too.

libsql-replication/proto/metadata.proto
Horusiath: How would it behave if we parse existing database config, where this tag was not defined? Wouldn't it error?

haaawk:
I'm not sure if you don't need to handle the new params in  handle_fork_namespace as well

libsql-replication/proto/metadata.proto
haaawk: if there's a default value then missing field should get default value.

haaawk: but this is not crucial and should block this PR.

MarinPostma:
Couple things:
- I think 'shared_schema' is a bit confusing, it's not evident if it's the db that stores the schema, or if it uses a schema define somewhere else. Maybe something like 'master_schema_db' or something around those lines would make it clearer?

- how should we handle deletion of a schema db? That would leave follower databases orphaned 
- do we need to propagate config for a schema database to the replicas? I don't think so

haaawk:
> Couple things:
> 
> * I think 'shared_schema' is a bit confusing, it's not evident if it's the db that stores the schema, or if it uses a schema define somewhere else. Maybe something like 'master_schema_db' or something around those lines would make it clearer?

Since shared_schema is a bool I think it's fine. I would prefer `shared_schema_db` over `master_schema_db` as master does not relate to any concept in the shared schema feature. I think that `*_db` param in db creation params is a bit redundant though.

> * how should we handle deletion of a schema db? That would leave follower databases orphaned

We should probably disallow deletion of schema db if more than 0 dbs depend on it

> * do we need to propagate config for a schema database to the replicas? I don't think so

Probably not

haaawk:
> > Couple things:
> > 
> > * I think 'shared_schema' is a bit confusing, it's not evident if it's the db that stores the schema, or if it uses a schema define somewhere else. Maybe something like 'master_schema_db' or something around those lines would make it clearer?
> 
> Since shared_schema is a bool I think it's fine. I would prefer `shared_schema_db` over `master_schema_db` as master does not relate to any concept in the shared schema feature. I think that `*_db` param in db creation params is a bit redundant though.
> 
> > * how should we handle deletion of a schema db? That would leave follower databases orphaned
> 
> We should probably disallow deletion of schema db if more than 0 dbs depend on it

Let's do this as a follow up. I will do that to allow @Horusiath to focus on the meat of the schema change logic

> 
> > * do we need to propagate config for a schema database to the replicas? I don't think so
> 
> Probably not

MarinPostma:
> We should probably disallow deletion of schema db if more than 0 dbs depend on it

We probably need to structure the meta store in such a way that we can efficiently lookup that information

MarinPostma:
Yeah I'm thinking maybe another table, many-to-one mapping share schema to their source of truth. We can make it indexable

haaawk:
> Yeah I'm thinking maybe another table, many-to-one mapping share schema to their source of truth. We can make it indexable

A new table sounds like a good idea. You mean new table in metastore, right?
I'm wondering how much work is it vs just having a comma separated list of dbs that depend on "me" under `affected_dbs` key?

MarinPostma (approved):

PR #1053, opened 2024-02-20
Make `ATTACH` enabled by default

avinassh:
+build


MarinPostma:
This is a security concern, we really don't want that to be anything but opt-in

haaawk:
> This is a security concern, we really don't want that to be anything but opt-in

PR #1054, opened 2024-02-20
Merge upstream SQLite 3.45.1

penberg:


nat-418:
this will be very useful

penberg:
I did a merge of this, but Github squashed it, making the repository incompatible with upstream SQLite. I dropped the merge so let's re-open.

PR #1067, opened 2024-02-21
Shared schema execution path

Horusiath:
Built on top of #1055

### DIFF
--- a/libsql-server/src/connection/mod.rs
+++ b/libsql-server/src/connection/mod.rs
@@ -1,3 +1,4 @@
+use async_trait::async_trait;
 use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::Arc;
 use tokio::time::{Duration, Instant};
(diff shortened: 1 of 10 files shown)


libsql-server/src/connection/mod.rs
```diff
@@ -189,6 +197,89 @@ where
...
+impl<DB: Connection> Connection for SharedSchemaConnection<DB> {
+    async fn execute_program<B: QueryResultBuilder>(
+        &self,
+        pgm: Program,
+        auth: Authenticated,
+        response_builder: B,
+        replication_index: Option<FrameNo>,
+    ) -> Result<B> {
+        //TODO: execute multi-database execution procedure
+        todo!()
```
Horusiath, line 249: This part will be responsible for executing multi-database execution. We'll need to modify current connection state with data necessary for driving the execution process, dry_runs etc.

Horusiath:
-

PR #1074, opened 2024-02-22
server: reduce size of slow read/write log

LucioFranco:


libsql-server/src/connection/libsql.rs
```diff
@@ -929,7 +929,7 @@ impl<W: Wal> Connection<W> {
                 "high read ({}) or write ({}) query: {}",
                 rows_read,
                 rows_written,
-                sql
+                &sql[..128]
```
athoscouto, line 932: Consider capping it to 512 or 1024. 
As a reference, a small snapshot compaction log has 220 chars with metadata.

PR #1076, opened 2024-02-22
WIP: multi-ns auth

MarinPostma:


haaawk:
Why can't we just allow attach on all namespaces? To attach a namespace JWT needs an appropriate claim isn't that enough?

libsql-server/src/auth/authenticated.rs
```diff
@@ -80,14 +44,31 @@ impl Authenticated {
...
+        namespace: &NamespaceName,
+        perm: Permission,
+    ) -> crate::Result<()> {
+        match self {
+            Authenticated::Anonymous => Err(crate::Error::NotAuthorized(
+                "anonymous access not allowed".to_string(),
+            )),
+            Authenticated::Authorized(a) => {
+                if !a.has_right(Scope::Namespace(namespace.clone()), perm) {
+                    return Err(crate::Error::NotAuthorized(format!(
```
haaawk, line 65: ultra nit: No need for return here.

libsql-server/src/auth/authorized.rs
```diff
@@ -1,8 +1,162 @@
...
+            (Permission::Read, Scope::Namespace(ref name)) => self.can_read_ns(name),
+            (Permission::Write, Scope::Namespace(ref name)) => self.can_write_ns(name),
+            (Permission::AttachRead, Scope::Namespace(ref name)) => self.can_attach_ns(name),
+        }
+    }
+
+    pub fn merge_legacy(&mut self, namespace: NamespaceName, perm: Permission) {
+        let scope = match perm {
+            Permission::Read => self.read_only.get_or_insert_with(Default::default),
+            Permission::Write => self.read_only.get_or_insert_with(Default::default),
```
haaawk, line 36: Shouldn't this be self.read_write?

libsql-server/src/auth/authorized.rs
```diff
@@ -1,8 +1,162 @@
...
-    pub namespace: Option<NamespaceName>,
-    pub permission: Permission,
+    #[serde(rename = "ro", default)]
+    pub read_only: Option<Scopes>,
+    #[serde(rename = "rw", default)]
+    pub read_write: Option<Scopes>,
+    #[serde(rename = "roa", default)]
+    pub read_only_attach: Option<Scopes>,
+    #[serde(rename = "rwa", default)]
+    pub read_write_attach: Option<Scopes>,
```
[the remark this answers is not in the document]
MarinPostma: this is fo when you have rw perm to the database, but also attach, it doesn't mean that you have write-attach perms, that we would write `rwaw` for example

libsql-server/src/auth/authorized.rs
MarinPostma: good catch

libsql-server/src/auth/authorized.rs
```diff
@@ -1,8 +1,162 @@
...
+
+        if let Some(Scopes {
+            namespaces: Some(ref ns),
+            ..
+        }) = self.read_write_attach
+        {
+            if ns.contains(name) {
+                return true;
+            }
+        }
```
haaawk, line 65: This can go away.

libsql-server/src/auth/authorized.rs
```diff
@@ -1,8 +1,162 @@
...
```
haaawk, line 141: This can go away.

haaawk (approved):
LGTM

Issue #1077, opened 2024-02-23
Rust suite tests are running twice


avinassh:
The PR #1027 added tests for extensions in the [github workflow ](https://github.com/tursodatabase/libsql/blob/d178de8/.github/workflows/extensions-test.yml), but it runs the entire rust suite again (which are already run as part of rust workflow). It should run only the `--extensions` tests: 

```
...
test extensions::test_sqlite_vss ... ok
```

https://github.com/tursodatabase/libsql/actions/runs/8015027296/job/21894572798

tungbq:
Hi @avinassh I would like to work on this issue, could you help to guide me the full command to run extension test?

avinassh:
@tungbq sure!

```
cd libsql-sqlite3/test/rust_suite
cargo test --features extensions
```

tungbq:
Thank you @avinassh I've proposed PR #1078 to fix the issue. Please review and let me know if we need to have more update.

PR #1078, opened 2024-02-23
ci: fix duplicated rust suites test

tungbq:
Closes: #1077 
ci: fix duplicated rust suites test
Only the `extensions` tests

### DIFF
--- a/.github/workflows/extensions-test.yml
+++ b/.github/workflows/extensions-test.yml
@@ -76,4 +76,5 @@ jobs:
       - name: test
         run: |
           cd libsql-sqlite3/test/rust_suite
-          cargo test --features extensions
+          # Only execute extensions tests
+          cargo test --features=extensions extensions


.github/workflows/extensions-test.yml
```diff
@@ -36,26 +36,6 @@ jobs:
...
-        uses: actions/checkout@v3
-      - name: build libsql
-        run: |
-          ./configure
-          make libsql
-      - name: test
-        run: |
-          cd ext/crr/rs/core
-          cargo test --features=loadable_extension
-
```
avinassh, line 58: hey, this should not be removed

.github/workflows/extensions-test.yml
```diff
@@ -76,4 +56,4 @@ jobs:
       - name: test
         run: |
           cd libsql-sqlite3/test/rust_suite
-          cargo test --features extensions
\ No newline at end of file
+          cargo test --features extensions
```
avinassh, line 59: this part targets all tests, including the one in extensions module. 

What we want is, just run the extensions module tests, but nothing else.

avinassh (changes requested):

.github/workflows/extensions-test.yml
tungbq: Actually, I do not familiar with the Rust and Rust test command, just trying to work on the CI side but it seems complicated than I thought. So was not able to figure out the right `cargo test` command to run the extensions module tests only so far. If you have any hint or document related to this, please let me know. Thanks!

avinassh: docs: https://doc.rust-lang.org/cargo/commands/cargo-test.html#target-selection

something like this would work:

```
cargo test --features extensions --test tests extensions
```

tungbq: Thanks, I will check it to find the appropriate command for this case.

tungbq: Finally, I've found the command to run only extension test, it works on my forked repo
https://github.com/tungbq/libsql/actions/runs/8028972291/job/21934748847#step:6:474
image
Hi @avinassh please help review again! Thanks

avinassh: @tungbq looks great! can you squash all the commits into a single one?

tungbq: It's done @avinassh

avinassh (approved):

PR #1080, opened 2024-02-24
Shared schema

haaawk:
Redo of https://github.com/tursodatabase/libsql/pull/1055

It basically stores the relationship between shared schema db and databases linked to it:

The relationships are stored in form of shared_schema_links table in the meta store of a shared-schema db.
When we create a dependent db, we store a link in shared-schema.
When we destroy dependent db, we remove link from shared-schema.
When we destroy shared-schema, we first check if all dependent dbs are already destroyed. We fail if they don't.

Some known decifiencies:
1. No way to fetch the schema links yet
2. ~~Forking a db that links to a shared schema should add a link between a new fork and a shared schema but doesn't do it right now.~~

### DIFF
--- a/libsql-server/src/http/admin/mod.rs
+++ b/libsql-server/src/http/admin/mod.rs
@@ -305,7 +305,7 @@ async fn handle_create_namespace<M: MakeNamespace, C: Connector>(
             ));
         }
         let namespace = NamespaceName::from_string(ns)?;
-        if !app_state.namespaces.exists(&namespace).await {
+        if !app_state.namespaces.exists(&namespace) {
             return Err(Error::NamespaceDoesntExist(namespace.to_string()));
         }
         Some(namespace.to_string())
--- a/libsql-server/src/http/admin/mod.rs
+++ b/libsql-server/src/http/admin/mod.rs
@@ -286,7 +286,7 @@ struct CreateNamespaceReq {
     shared_schema: bool,
     /// If some, this is a [NamespaceName] reference to a shared schema DB.
     #[serde(default)]
-    shared_schema_name: Option<String>,
+    shared_schema_name: Option<NamespaceName>,
 }
 
 async fn handle_create_namespace<M: MakeNamespace, C: Connector>(
@@ -304,11 +304,11 @@ async fn handle_create_namespace<M: MakeNamespace, C: Connector>(
                 "shared schema database cannot reference another shared schema".to_string(),
             ));
         }
-        let namespace = NamespaceName::from_string(ns)?;
-        if !app_state.namespaces.exists(&namespace) {
-            return Err(Error::NamespaceDoesntExist(namespace.to_string()));
+        // TODO: move this check into meta store
+        if !app_state.namespaces.exists(&ns) {
+            return Err(Error::NamespaceDoesntExist(ns.to_string()));
         }
-        Some(namespace.to_string())
+        Some(ns)
     } else {
         None
     };
@@ -334,7 +334,7 @@ async fn handle_create_namespace<M: MakeNamespace, C: Connector>(
     let mut config = (*store.get()).clone();
 
     config.is_shared_schema = req.shared_schema;
-    config.shared_schema_name = shared_schema_name;
+    config.shared_schema_name = shared_schema_name.as_ref().map(|x| x.to_string());
     if let Some(max_db_size) = req.max_db_size {
         config.max_db_pages = max_db_size.as_u64() / LIBSQL_PAGE_SIZE;
     }
@@ -379,6 +379,7 @@ async fn handle_fork_namespace<M: MakeNamespace, C>(
     let mut to_config = (*to_store.get()).clone();
     to_config.max_db_pages = from_config.max_db_pages;
     to_config.heartbeat_url = from_config.heartbeat_url.clone();
+    to_config.shared_schema_name = from_config.shared_schema_name.clone();
     to_store.store(to_config).await?;
     Ok(())
 }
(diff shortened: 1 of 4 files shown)


libsql-server/src/namespace/meta_store.rs
```diff
@@ -322,18 +374,40 @@ impl MetaStore {
...
+            let tx = guard.conn.transaction()?;
+            if let Some(shared_schema) = config.config.shared_schema_name.as_deref() {
+                tx.execute(
+                    "DELETE FROM shared_schema_links WHERE shared_schema_name = ? AND db_name = ?",
+                    [shared_schema, namespace.as_str()],
+                )?;
+            }
+            if config.config.is_shared_schema {
+                let res: rusqlite::Result<u32> = tx.query_row(
+                    "SELECT COUNT(*) FROM shared_schema_links WHERE shared_schema_name = ?",
```
MarinPostma, line 389: this should not be necessary if:
- we enable foreign keys on the connection (`PRAGMA foreign_keys=ON`)
- add a deletion restriction on the table
    - `FOREIGN KEY (shared_schema_name) REFERENCES namespace_configs (namespace) ON DELETE RESTRICT`
    - `FOREIGN KEY (db_name) REFERENCES namespace_configs (namespace) ON DELETE RESTRICT`

libsql-server/src/namespace/meta_store.rs
```diff
@@ -322,18 +374,40 @@ impl MetaStore {
...
+                    |row| row.get(0),
+                );
+                match res {
+                    Ok(0) | Err(rusqlite::Error::QueryReturnedNoRows) => { /* ok */ }
+                    Ok(n) => return Err(Error::SharedSchemaDestroyError(namespace.to_string(), n)),
+                    Err(e) => return Err(e.into()),
+                }
+            }
+            tx.execute(
+                "DELETE FROM namespace_configs WHERE namespace = ?",
```
MarinPostma, line 400: shouldn't we also delete the link?

libsql-server/src/namespace/meta_store.rs
```diff
@@ -343,6 +417,46 @@ impl MetaStore {
         self.inner.lock().configs.contains_key(namespace)
     }
 
+    pub async fn update_namespace_schema_and_config(
```
MarinPostma, line 420: I don't think this is necessary, the schema is part of the config, update the config should be sufficient, if updating the link is part of updating the config itself

libsql-server/src/http/admin/mod.rs
```diff
@@ -298,20 +298,22 @@ async fn handle_create_namespace<M: MakeNamespace, C: Connector>(
...
-    };
+    let (shared_schema_name, shared_schema_namespace_name) =
+        if let Some(ns) = req.shared_schema_name {
+            if req.shared_schema {
+                return Err(Error::SharedSchemaCreationError(
+                    "shared schema database cannot reference another shared schema".to_string(),
+                ));
+            }
+            let name_string = ns.clone();
+            let namespace = NamespaceName::from_string(ns)?;
```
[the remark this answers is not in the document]
haaawk: done

libsql-server/src/http/admin/mod.rs
```diff
@@ -298,20 +298,22 @@ async fn handle_create_namespace<M: MakeNamespace, C: Connector>(
...
+            if !app_state.namespaces.exists(&namespace) {
+                return Err(Error::NamespaceDoesntExist(namespace.to_string()));
+            }
```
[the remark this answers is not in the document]
haaawk: Added a TODO.

libsql-server/src/namespace/meta_store.rs
```diff
@@ -128,6 +132,17 @@ impl MetaStoreInner {
             ",
             (),
         )?;
+        conn.execute(
+            "CREATE TABLE IF NOT EXISTS shared_schema_links (
+                shared_schema_name TEXT NOT NULL,
+                db_name TEXT NOT NULL,
```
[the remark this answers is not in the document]
haaawk: I was thinking about naming it namespace but since both shared schema and the other field are namespaces, I found it unclear and misleading. Naming the fields schema + db was meant to be a bit more informative but that's ok. Changed to namespace.

libsql-server/src/namespace/meta_store.rs
```diff
@@ -243,32 +258,69 @@ impl MetaStoreInner {
...
+            let config_encoded = metadata::DatabaseConfig::from(&*config).encode_to_vec();
+
+            let inner = &mut inner.lock();
+
+            inner.conn.execute(
+                "INSERT OR REPLACE INTO namespace_configs (namespace, config) VALUES (?1, ?2)",
+                rusqlite::params![namespace.as_str(), config_encoded],
+            )?;
+
+            let configs = &mut inner.configs;
```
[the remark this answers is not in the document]
haaawk: You're right. I will simplify. Good catch!

libsql-server/src/namespace/meta_store.rs
haaawk: This is true. I thought about removing the check but ended up leaving it just to be able to give user a better error message with info how many existing namespaces still link to the schema. Will remove.

libsql-server/src/namespace/meta_store.rs
haaawk: The link is removed above in lines 381-386.

haaawk:
Thanks for a great review @MarinPostma! I applied all your suggestions - please have another look.

MarinPostma (approved):

PR #1083, opened 2024-02-26
Allow multiple namespaces types to be stored in the same `NamespaceStore`

MarinPostma:
This PR refactors sqld to allow the NamespaceStore to store different kinds of namespaces. For now this means only replica and primaries, but this paves the way to storing schema databases as well.


libsql-server/src/http/admin/mod.rs
```diff
@@ -310,51 +307,31 @@ async fn handle_create_namespace<M: MakeNamespace, C: Connector>(
...
-
-    let store = app_state.namespaces.config_store(namespace).await?;
-    let mut config = (*store.get()).clone();
-
+    config.bottomless_db_id = req.bottomless_db_id;
     config.is_shared_schema = req.shared_schema;
-    config.shared_schema_name = shared_schema_name.as_ref().map(|x| x.to_string());
+    config.heartbeat_url = req.heartbeat_url.as_deref().map(Url::parse).transpose()?;
+    config.txn_timeout = req.txn_timeout_s.map(Duration::from_secs);
+    config.max_row_size = req.max_row_size.unwrap_or(config.max_row_size);
```
Horusiath, line 325: Are you sure that this won't reset settings by accident? Previously we only had to provide fields that were supposed to be changed. Now missing fields would reset config values.

libsql-server/src/namespace/mod.rs
```diff
@@ -190,110 +78,73 @@ pub enum NamespaceBottomlessDbIdInit {
...
-    /// Create a new Namespace instance
-    async fn create(
-        &self,
-        name: NamespaceName,
+impl Namespace {
+    async fn from_config(
+        ns_config: &NamespaceConfig,
+        db_config: MetaStoreHandle,
         restore_option: RestoreOption,
-        bottomless_db_id: NamespaceBottomlessDbId,
```
Horusiath, line 203: What happened to this param?

MarinPostma: let's track this params down.
- it's passes to try [`try_new_primary`](https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/namespace/mod.rs#L1096-L1109), where it's either passed as `Namespace` or read from config. `try_new_primary` is called by [`new_primary`](https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/namespace/mod.rs#L1054), which in turn is called by [`PrimaryNamespaceMaker::create`](https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/namespace/mod.rs#L252-L259)

- `MakeNamespace::create` is called by:
    - [`with`](https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/namespace/mod.rs#L669-L676) (no provided)
    - [`reset`](https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/namespace/mod.rs#L535-L541) (not provided)
    - in [`fork`](https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/namespace/mod.rs#L605-L612) (not provided)
    - in [`try_fork`](https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/namespace/fork.rs#L109-L120), from `self.bottomless_id`, taken from [source namespace config](https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/namespace/mod.rs#L314)
    - from `[NamespaceStore::create`](https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/namespace/mod.rs#L742-L749), only called by [`HTTP::admin::handle_create`](https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/http/admin/mod.rs#L332), with the param from the request. 
    
  So it's only dependent on the value of the if provided during namespace creation. That value was not stored in the meta, and passed directly to namespace creation. If you ask me this whole code was smelly AF. Instead, we set the db id in the config, and create the namespace with a consistent config.

MarinPostma: So we just read it from config.

Horusiath (approved):

PR #1085, opened 2024-02-26
Bottomless savepoint

Horusiath:
This PR introduces a new feature to bottomless - ability to materialise tracker object, that can be awaited for in order to confirm that bottomless asynchronous upload matched the replication status at the moment of method call.

### DIFF
--- a/bottomless/src/completion_progress.rs
+++ b/bottomless/src/completion_progress.rs
@@ -83,12 +83,16 @@ impl CompletionProgress {
     }
 
     pub fn update(&mut self, mut start_frame: u32, mut end_frame: u32) {
-        if start_frame - 1 == self.baseline {
-            while start_frame - 1 == self.baseline {
+        if start_frame == self.baseline + 1 {
+            loop {
                 self.baseline = end_frame;
                 if let Some((s, e)) = self.detached_ranges.pop_first() {
                     start_frame = s;
                     end_frame = e;
+                    if start_frame != self.baseline + 1 {
+                        self.detached_ranges.insert(start_frame, end_frame);
+                        break;
+                    }
                 } else {
                     break;
                 }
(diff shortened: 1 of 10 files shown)


bottomless/src/completion_progress.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub fn update(&mut self, mut start_frame: u32, mut end_frame: u32) {
+        if start_frame == self.baseline + 1 {
+            loop {
+                self.baseline = end_frame;
+                if let Some((s, e)) = self.detached_ranges.pop_first() {
+                    start_frame = s;
+                    end_frame = e;
+                    if start_frame != self.baseline + 1 {
+                        self.detached_ranges.insert(start_frame, end_frame);
+                        break;
```
MarinPostma, line 94: isn't that basically equivalent?
```suggestion
                if let Some((s, e)) = self.detached_ranges.pop_first() {
                    if s != self.baseline + 1 {
                        self.detached_ranges.insert(s, e);
                        break;
```

bottomless/src/replicator.rs
```diff
@@ -317,27 +321,31 @@ impl Replicator {
             })
         };
 
+        let (upload_progress, last_uploaded_frame_no) = CompletionProgress::new(0);
```
MarinPostma, line 324: why is baseline necessarily 0? could we be in the middle of a generation?

MarinPostma (approved):
looks good

bottomless/src/completion_progress.rs
Horusiath: No, since in loop we're progressing `self.baseline = end_frame`.

bottomless/src/replicator.rs
Horusiath: When we initialise it, it's always 0. If we're in a middle of the generation, we know about this as part of restore procedure, which happens after initialisation. Then change below checks out the local status of a frame counts and updates it: 
```rust
{
    let mut guard = self.upload_progress.lock().await;
    guard.update(1, last_frame);
}
```

PR #1086, opened 2024-02-26
Aggregated stats for queries with most elapsed_time

athoscouto:
We need to be able to provide statistics about an individual query.
This is the first step, adding `count`, `elapsed_ms`, `rows_written`, and `rows_read`.

--- 

Those stats are not persisted on purpose, since we want them to be ephemeral and even reset from time to time.

### DIFF
--- a/libsql-server/src/stats.rs
+++ b/libsql-server/src/stats.rs
@@ -2,14 +2,15 @@ use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, RwLock, Weak};
 
+use chrono::{DateTime, DurationRound, Utc};
+use hdrhistogram::Histogram;
 use metrics::{counter, gauge, histogram, increment_counter};
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeSet, HashMap};
 use tokio::io::AsyncWriteExt;
 use tokio::sync::mpsc;
 use tokio::task::JoinSet;
-use tokio::time::Duration;
-use tracing::debug;
+use tokio::time::{Duration, Instant};
 use uuid::Uuid;
 
 use crate::namespace::NamespaceName;
@@ -83,31 +84,46 @@ impl QueryStats {
     }
 }
 
-#[derive(Debug, Default, Serialize, Deserialize)]
+#[derive(Debug)]
 pub struct QueriesStats {
-    #[serde(default)]
-    id: Option<Uuid>,
-    #[serde(default)]
-    stats_threshold: AtomicU64,
-    #[serde(default)]
+    id: Uuid,
     stats: HashMap<String, QueryStats>,
+    elapsed: Duration,
+    stats_threshold: u64,
+    hist: Histogram<u32>,
+    expires_at: Option<Instant>,
+    created_at: DateTime<Utc>,
 }
 
 impl QueriesStats {
-    fn new() -> Arc<RwLock<Self>> {
-        let mut this = QueriesStats::default();
-        this.id = Some(Uuid::new_v4());
-        Arc::new(RwLock::new(this))
+    fn new() -> Self {
+        Self {
+            id: Uuid::new_v4(),
+            stats: HashMap::new(),
+            elapsed: Duration::default(),
+            stats_threshold: 0,
+            hist: Histogram::<u32>::new(3).unwrap(),
+            expires_at: None,
+            created_at: Utc::now(),
+        }
+    }
+
+    fn with_expiration(expires_at: Instant) -> Self {
+        let mut this = Self::new();
+        this.expires_at = Some(expires_at);
+        this
     }
 
     fn register_query(&mut self, sql: &String, stat: QueryStats) {
+        self.elapsed += stat.elapsed;
+        let _ = self.hist.record(stat.elapsed.as_millis() as u64);
+
         let (aggregated, new) = match self.stats.get(sql) {
             Some(aggregated) => (aggregated.merge(&stat), false),
             None => (stat, true),
         };
 
-        debug!("query: {}, elapsed: {:?}", sql, aggregated.elapsed);
-        if (aggregated.elapsed.as_micros() as u64) < self.stats_threshold.load(Ordering::Relaxed) {
+        if (aggregated.elapsed.as_micros() as u64) < self.stats_threshold {
             return;
         }
 
@@ -132,8 +148,7 @@ impl QueriesStats {
 
     fn update_threshold(&mut self) {
         if let Some((_, v)) = self.min() {
-            self.stats_threshold
-                .store(v.elapsed.as_micros() as u64, Ordering::Relaxed);
+            self.stats_threshold = v.elapsed.as_micros() as u64;
         }
     }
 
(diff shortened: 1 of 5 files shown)


athoscouto:
`elapsed_ms` granularity may be too coarse to use as a sorting weight.
It may be a good idea to fall back to `rows_read + rows_written`.

libsql-server/src/stats.rs
```diff
@@ -52,6 +53,100 @@ impl SlowestQuery {
...
+        }
+    }
+    pub fn empty() -> Self {
+        Self {
+            elapsed_ms: 0,
+            count: 0,
+            rows_read: 0,
+            rows_written: 0,
+        }
+    }
```
LucioFranco, line 80: I believe default will do this just fine? no need for a second fn?

libsql-server/src/stats.rs
```diff
@@ -52,6 +53,100 @@ impl SlowestQuery {
...
+            rows_written: self.rows_written + another.rows_written,
+        }
+    }
+}
+
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct QueriesStats {
+    #[serde(default)]
+    id: Option<Uuid>,
+
```
LucioFranco, line 95:
```suggestion
```

libsql-server/src/stats.rs
```diff
@@ -52,6 +53,100 @@ impl SlowestQuery {
...
+}
+
+impl QueriesStats {
+    pub fn new() -> Arc<RwLock<Self>> {
+        let mut this = QueriesStats::default();
+        this.id = Some(Uuid::new_v4());
+        Arc::new(RwLock::new(this))
+    }
+
+    pub fn register_query(&mut self, sql: &String, stat: QueryStats) {
```
LucioFranco, line 109: Instead of using a lock, what about using a background task and a channel. That way the cost of doing this work is offloaded from the request path and we reduce contention to just sending the query over the channel? We need to clone anyways so there is no issue there but we avoid the lock contention. It should also simplify the logic since you don't need to thread this lock everywhere?

LucioFranco: Another option because one challenge with a background task is that the namespace might get deallocated, is to use a concurrent hashmap that should be faster than a mutex<hashmap> https://docs.rs/dashmap/latest/dashmap/ is one option.

libsql-server/src/stats.rs
```diff
@@ -52,6 +53,100 @@ impl SlowestQuery {
...
+        let (aggregated, new) = match self.stats.get(sql) {
+            Some(stat) => (stat.clone(), false),
+            None => (QueryStats::empty(), true),
+        };
+
+        let aggregated = aggregated.merge(&stat);
```
LucioFranco, line 115: I think there is a way here if you embed the merge within the borrow of the get that you don't need to clone the stat coming from the hashmap. You might do something like

```rust
let aggregated = if let Some(aggregated) = self.stats.get(&sql) {
    aggregated.merge(&stat)
} else {
    stat
};
```

libsql-server/src/stats.rs
```diff
@@ -52,6 +53,100 @@ impl SlowestQuery {
...
+        }
+
+        self.stats.insert(sql.clone(), aggregated);
+
+        if !new || self.stats.len() <= 30 {
+            return;
+        }
+
+        let mut vec = self.stats.clone().into_iter().collect_vec();
+        vec.sort_by(|a, b| a.1.partial_cmp(&b.1).unwrap());
```
LucioFranco, line 127: whats the case here for this unwrap? might want to drop a comment why this is okay to do.

libsql-server/src/stats.rs
```diff
@@ -52,6 +53,100 @@ impl SlowestQuery {
...
+            return;
+        }
+
+        self.stats.insert(sql.clone(), aggregated);
+
+        if !new || self.stats.len() <= 30 {
+            return;
+        }
+
+        let mut vec = self.stats.clone().into_iter().collect_vec();
```
LucioFranco, line 126: could we use https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.drain here instead? I think it would be more efficient.

libsql-server/src/stats.rs
```diff
@@ -52,6 +53,100 @@ impl SlowestQuery {
...
+        vec.sort_by(|a, b| a.1.partial_cmp(&b.1).unwrap());
+        let len = vec.len();
+        if len <= 30 {
+            return;
+        }
+
+        for i in 0..len - 30 {
+            self.stats.remove(&vec[i].0);
```
LucioFranco, line 134: drain here might also be more efficient.

libsql-server/src/stats.rs
```diff
@@ -82,6 +177,12 @@ pub struct Stats {
     slowest_queries: Arc<RwLock<BTreeSet<SlowestQuery>>>,
     #[serde(default)]
     embedded_replica_frames_replicated: AtomicU64,
+    #[serde(default)]
+    query_count: AtomicU64,
+    #[serde(default)]
+    query_latency: AtomicU64,
+    #[serde(skip)]
+    queries: Arc<RwLock<QueriesStats>>,
```
LucioFranco, line 185: I would consider using `parking_lot::RwLock` or Mutex here instead of the std one.

libsql-server/src/stats.rs
athoscouto: we don't,  this is a leftover from when I stored the SQL query itself inside the stats struct
back then we needed a default + SQL
I'll remove it!

libsql-server/src/stats.rs
athoscouto: lack of familiarity with the language, so reasoning about every bit gets difficult
we should be able to replace it with `a.1.cmp(&b.1)`
wdyt?

libsql-server/src/stats.rs
athoscouto: I tried replacing all stats `RwLocks` with `parking_lot::RwLock` but it breaks (de)serialization of the stats struct.
Let me know if you want me to change just this single lock, since we skip serialization of the `queries` field.

libsql-server/src/stats.rs
athoscouto: My understanding is that drain removes all values from the map. 
I don't want to do this, since I just want to remove a single entry.
I've removed the vec in favor of just iterating and getting the minimum element.

libsql-server/src/stats.rs
athoscouto: As we talked on Slack, I think the background task should be used to upgrade all stats.
I've implemented an approach that does this, communicating synchronously with a 256 slots buffer.

The downside is that now we have a single thread updating stats.
If for some reason the throughput of that thread is smaller than the query throughput, it will create backpressure.

The upside is that we may consider removing/simplifying some of the internal locks.
We still need to manage the concurrency between the single reader and single writer.
But the atomic ints should be unnecessary at this point.

athoscouto:
@LucioFranco it looks like my background worker is making tests hang.
Any idea why? Probably the blocking thread?

libsql-server/src/stats.rs
LucioFranco: Yeah I think so that should do the trick here, better than a panic lol

LucioFranco (approved):

PR #1100, opened 2024-02-28
Go: Reset connection when it's returned to a pool

haaawk:
Go has a pool of connections.
We need to reset baton in the connection every time it's returned to the pool.
Otherwise, user gets STREAM EXPIRED errors a lot.

This depends on https://github.com/tursodatabase/libsql/pull/1099


haaawk:
@penberg @LucioFranco I'm changing public API in
[Make all methods of Conn trait async](https://github.com/tursodatabase/libsql/pull/1100/commits/c4d053187d1fb2e2926fac96786a6cac7a4350e3)`
Is this ok with you? I hope it does not break other SDKs

PR #1101, opened 2024-02-28
Fix build

haaawk:


libsql-ffi/build.rs
```diff
@@ -10,6 +10,13 @@ const BUNDLED_DIR: &str = "bundled";
 const SQLITE_DIR: &str = "../libsql-sqlite3";
 
 fn main() {
+    let target = env::var("TARGET").unwrap();
+    let host = env::var("HOST").unwrap();
+
+    let is_apple = host.contains("apple") && target.contains("apple");
+    if is_apple {
```
LucioFranco, line 17: do we also want to feature flag this just for encryption since this is only needed for that?

LucioFranco (approved):
LGTM left one imporvement

PR #1104, opened 2024-02-28
libsql: prefix client_wal_index file

LucioFranco:


libsql/src/local/database.rs
```diff
@@ -42,6 +43,10 @@ impl Database {
                 "Unable to open local database {db_path} with Database::open()"
             )))
         } else {
+            if let Some(parent) = std::path::Path::new(&db_path).parent() {
+                std::fs::create_dir_all(&parent).map_err(|e| Error::UnableToCreateDirAll(e))?;
+            }
```
penberg, line 48: This is not needed anymore is it?

LucioFranco: I wasn't sure if the C code will create the directory or not but I was running into issues with my replica test that didn't have the directory created already. So I think this is just safe to have?

penberg: But `sqlite3` doesn't create the directories either:

```
penberg@vonneumann bun-todo-api % sqlite3 foo/x.db
SQLite version 3.44.2 2023-11-24 11:41:44
Enter ".help" for usage hints.
sqlite> CREATE TABLE users (foo);
Error: unable to open database "foo/x.db": unable to open database file
```

Let's just drop that from the database code and make sure in the tests that directories exist.

PR #1103, opened 2024-02-28
register schema migrations

MarinPostma:
Introduce the schema database, and adds the ability to register migration jobs. Jobs are not performed yet.

PR #1109, opened 2024-02-28
Factorize program execution in VM object

MarinPostma:
Factorize program execution in the Vm struct. Schema migration requires more flexibility with program execution, and the `Connection` trait is making it more complicated to work with.

depends on #1103


haaawk (approved):
LGTM

PR #1110, opened 2024-02-29
schema migration

MarinPostma:
implements the happy path of schema migration. do not use this feature yet.

TODO:
- [ ] move all of the migration related stuff from the meta_store to the schema module; let the scheduler use it's own connection to the meta store.
- [ ] move the schema for the meta store to a sqlite file, and `include!` that file for clarity
- [ ] perform error handling, check migration success status, use custom QueryResultHandler to collect results
- [ ] ensure that the temp table `enqueued_tasks` is updated corectly after a task is popped when a tasks status changed (maybe use a trigger?)
- [ ] handle abort; first when a dry run failed, then if a user abort the migration before the dry-run is complete
- [ ] prevent a job from being enqueued if there is still a pending job for that namespaces
- [ ] Add loads of tests, and test load
- [ ] prevent writes while there is are pending jobs in a database
- [ ] sync database before starting migration
- [ ] sync meta store after enqueue job
- [ ] track version schema
- [ ] go through all todos that I left in the code


libsql-server/src/schema/status.rs
```diff
@@ -0,0 +1,92 @@
...
+impl MigrationTaskStatus {
+    pub fn from_int(i: u64) -> Self {
+        match i {
+            0 => Self::Enqueued,
+            1 => Self::DryRunSuccess,
+            2 => Self::DryRunFailure,
+            3 => Self::Run,
+            4 => Self::Success,
+            5 => Self::Failure,
+            _ => panic!(),
```
Horusiath, line 30:
```suggestion
            _ => unreachable!(),
```

libsql-server/src/schema/status.rs
```diff
@@ -0,0 +1,92 @@
...
+impl MigrationJobStatus {
+    pub fn from_int(i: u64) -> Self {
+        match i {
+            0 => Self::WaitingDryRun,
+            1 => Self::DryRunSuccess,
+            2 => Self::DryRunFailure,
+            3 => Self::WaitingRun,
+            4 => Self::RunSuccess,
+            5 => Self::RunFailure,
+            _ => panic!(),
```
Horusiath, line 81:
```suggestion
            _ => unreachable!(),
```

libsql-server/src/schema/migration.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+pub fn perform_migration<B: QueryResultBuilder>(
+    conn: &mut rusqlite::Transaction,
+    migration: &Program,
+    dry_run: bool,
+    builder: B,
+) -> (Result<B, Error>, MigrationTaskStatus) {
+    // todo error handling is sketchy, improve
+    let mut savepoint = conn.savepoint().unwrap();
```
Horusiath, line 63: I guess this is also a place for bottomless savepoint (see #1085) in order to guarantee, that in case of crashes we can restore db to a state where the dry run succeeded. However in a current API it operates on the `Database` type not a `Connection`.

Horusiath (commented):
Since it's still a draft, just few points. The overall design seems pretty solid and aligned with what we were discussing in previous weeks. Please let me know when this is ready to full review.

haaawk (approved):

PR #1117, opened 2024-03-01
server: trim quotes in namespace string in attach statement

avinassh:
This patch comes with regression tests.

### DIFF
--- a/libsql-server/src/query_analysis.rs
+++ b/libsql-server/src/query_analysis.rs
@@ -128,9 +128,16 @@ impl StmtKind {
             Cmd::Stmt(Stmt::Attach {
                 expr: Expr::Id(Id(db_name)),
                 ..
-            }) => Some(Self::Attach(
-                NamespaceName::from_string(db_name.to_string()).ok()?,
-            )),
+            }) => {
+                let db_name = db_name
+                    .strip_prefix('"')
+                    .unwrap_or(db_name)
+                    .strip_suffix('"')
+                    .unwrap_or(db_name);
+                Some(Self::Attach(
+                    NamespaceName::from_string(db_name.to_string()).ok()?,
+                ))
+            }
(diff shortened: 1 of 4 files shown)


libsql-server/tests/standalone/attach.rs
```diff
@@ -214,3 +215,96 @@ fn attach_auth() {
...
+            &token,
+            TurmoilConnector,
+        )?
+        .connect()
+        .unwrap();
+
+        // fails: namespace is uuid, hence needs to be wrapped in quotes
+        assert_debug_snapshot!(main_conn
+            .execute(
+                "ATTACH DATABASE ae308915-caca-480f-a6b4-9f9f9dc84b11 as bar",
```
avinassh, line 291: I included this too, since I wanted to check the parser as well. I have hardcoded the UUID (which doesn't matter) since I am using snapshot tests.

libsql-server/src/query_analysis.rs
```diff
@@ -128,9 +128,16 @@ impl StmtKind {
...
                 ..
-            }) => Some(Self::Attach(
-                NamespaceName::from_string(db_name.to_string()).ok()?,
-            )),
+            }) => {
+                let db_name = db_name
+                    .strip_prefix('"')
+                    .unwrap_or(db_name)
+                    .strip_suffix('"')
+                    .unwrap_or(db_name);
```
MarinPostma, line 136: https://doc.rust-lang.org/std/primitive.str.html#method.trim_matches would that not work?

MarinPostma (approved):

libsql-server/src/query_analysis.rs
avinassh: I wanted to remove only at the only at start and end, not inside of the string. They are not expected to occur anywhere else in the normal flow and parser will fail it anyways. What do you say?

PR #1128, opened 2024-03-04
libsql: Force `AES256CBC` for embedded replica's

LucioFranco:


penberg:
I did https://github.com/tursodatabase/libsql/pull/1141 instead. We should keep the cipher configurability there in the API.

PR #1130, opened 2024-03-04
Jw/changing how jwt is passed around

shopifyski:
Standardized on a structure to pass around auth data instead of expecting downstream code to parse headers
`pub struct UserAuthContext {
    pub scheme: Option<String>,
    pub token: Option<String>,
}`

an immediate follow up cleanup should address the following challenges:
- duplication between code in http/user/mod.rs and http/user/extract.rs
- optionality of UserAuthContext fields - when context is required, its fields should not be optional
- tie userauthcontext to the auth strategy. make the strategy define whether certain edge cases (e.g. lack of token) should be interpreted as errors or dismissed


libsql-server/src/http/user/extract.rs
```diff
@@ -26,19 +29,23 @@ impl FromRequestParts<AppState> for RequestContext {
             .with(namespace.clone(), |ns| ns.jwt_key())
             .await??;
 
-        let auth_header = parts.headers.get(hyper::header::AUTHORIZATION);
+    // todo think how to decide if the particular auth stragegy even needs a context?
```
jeremywrowe, line 32: We had codified this into types in one of our first attempts of the auth strategy work. We may want to revisit that at some point, but it resulted in a large diff. We tied a context to a strategy by implementing a context trait.

jeremywrowe (commented):
Love the change! Let me know if you want help resolving some of the todo items. Other than that, I'd recommend moving this one to a normal PR state.

libsql-server/src/http/user/extract.rs
shopifyski: removing this comment. too vague. will propose a change in a separate pr

libsql-server/src/auth/user_auth_strategies/mod.rs
```diff
@@ -2,17 +2,26 @@ pub mod disabled;
...
 use super::{AuthError, Authenticated};
 
 pub struct UserAuthContext {
-    pub user_credential: Option<HeaderValue>,
+    pub scheme: Option<String>,
+    pub token: Option<String>, // token might not be required in some cases
 }
 
+impl TryFrom<&str> for UserAuthContext {
+    type Error = anyhow::Error;
```
LucioFranco, line 18: I think we should create some errors via `thiserror` and use them in this module, we could do a sub error enum that then embeds in the overall server error enum. That way instead of using context + strings we qualify it via a typed enum.

libsql-server/src/auth/user_auth_strategies/mod.rs
```diff
@@ -2,17 +2,26 @@ pub mod disabled;
...
 pub use disabled::*;
 pub use http_basic::*;
 pub use jwt::*;
 
 use super::{AuthError, Authenticated};
 
 pub struct UserAuthContext {
-    pub user_credential: Option<HeaderValue>,
+    pub scheme: Option<String>,
+    pub token: Option<String>, // token might not be required in some cases
```
LucioFranco, line 14: Since we are converting things via TryFrom now it would be a good time to turn these into private fields and using setters/gettings. It will make refactoring easier in the future. We should in general avoid pub fields, something that we have a lot of now but plan on fixing. Especially, if we are going to release this as a crate.

libsql-server/src/auth/user_auth_strategies/mod.rs
```diff
@@ -2,17 +2,26 @@ pub mod disabled;
 pub mod http_basic;
 pub mod jwt;
 
-use axum::http::HeaderValue;
+use anyhow::Context;
 pub use disabled::*;
 pub use http_basic::*;
 pub use jwt::*;
```
LucioFranco, line 8: Maybe this was missed in the last review but it would be good to remove the glob imports and make this more verbose. Can be done in a separate PR but I want to make sure we don't miss this.

libsql-server/src/auth/user_auth_strategies/mod.rs
```diff
@@ -2,17 +2,26 @@ pub mod disabled;
...
 
 use super::{AuthError, Authenticated};
 
 pub struct UserAuthContext {
-    pub user_credential: Option<HeaderValue>,
+    pub scheme: Option<String>,
+    pub token: Option<String>, // token might not be required in some cases
 }
 
+impl TryFrom<&str> for UserAuthContext {
```
LucioFranco, line 17: Thought, not sure if we should do this or not but we could implement `FromStr` instead of `TryFrom`? It would give us access to https://doc.rust-lang.org/stable/std/primitive.str.html#method.parse which seems like a better way to express what we are doing here?

libsql-server/src/hrana/ws/session.rs
```diff
@@ -77,16 +76,11 @@ pub(super) async fn handle_initial_hello(
...
-    let user_credential = jwt
-        .clone()
-        .and_then(|t| HeaderValue::from_str(&format!("Bearer {t}")).ok());
-
     let auth = namespace_jwt_key
         .map(Jwt::new)
         .map(Auth::new)
         .unwrap_or(server.user_auth_strategy.clone())
-        .authenticate(UserAuthContext { user_credential })
+        .authenticate(UserAuthContext { scheme: Some("Bearer".into()), token: jwt })
```
LucioFranco, line 83: Related to my comment above this would read better if we had a constructor like `UserAuthContext::from_scheme("Bearer", jwt)`.

libsql-server/src/http/user/extract.rs
```diff
@@ -15,6 +16,8 @@ impl FromRequestParts<AppState> for RequestContext {
         parts: &mut axum::http::request::Parts,
         state: &AppState,
     ) -> std::result::Result<Self, Self::Rejection> {
+
+        // start todo this block is same as the one in mod.rs
```
LucioFranco, line 20: is this some todo format Ive never seen before? Does this work with your editor? seems interesting!

libsql-server/src/http/user/extract.rs
```diff
@@ -26,19 +29,21 @@ impl FromRequestParts<AppState> for RequestContext {
             .with(namespace.clone(), |ns| ns.jwt_key())
             .await??;
 
-        let auth_header = parts.headers.get(hyper::header::AUTHORIZATION);
+        let context = parts.headers
+        .get(hyper::header::AUTHORIZATION).context("auth header not found") // todo this context is swallowed for now, gotta fix that but not with panicking
```
LucioFranco, line 33: Agreed, on the removing the panic 👍 I think moving to thiserror based on my comment above would be good here.

libsql-server/src/rpc/replica_proxy.rs
```diff
@@ -42,29 +43,25 @@ impl ReplicaProxyService {
             .with(namespace.clone(), |ns| ns.jwt_key())
             .await;
 
-        let user_credential = parse_grpc_auth_header(req.metadata());
+            //todo julian figure this out
```
LucioFranco, line 46: comment formatting, and would probably be good to add some more information here on what there is to figure out.

libsql-server/src/rpc/replication_log.rs
```diff
@@ -72,13 +71,13 @@ impl ReplicationLogService {
         req: &tonic::Request<T>,
         namespace: NamespaceName,
     ) -> Result<(), Status> {
+
+        // todo duplicate code
```
LucioFranco, line 75: could we add where the duplicate code is like there are in the other todos?

libsql-server/tests/cluster/replication.rs
```diff
@@ -91,7 +91,7 @@ fn apply_partial_snapshot() {
     sim.client("client", async move {
         let primary = libsql::Database::open_remote_with_connector(
             "http://primary:8080",
-            "",
+            "dummy_token",
```
LucioFranco, line 94: is there a problem with using the old `""`? I know we have this like everywhere with examples and our own code so ensuring that we can support that old method I think is quite important.

LucioFranco (changes requested):
Overall seems like a good change 👍 I left some feedback, feel free to ping me on slack if you have any questions. An additional heads up we are working on cutting a release this week so I will time the merging of this with that, likely after that just to be safe. Let me know if there are any timeline concerns and questions! Thanks!

libsql-server/tests/cluster/replication.rs
shopifyski: parser was crashing on string split
I was meaning to ask - what is the actual behavior we expect?
if auth header is present but not required, do we just ignore its content?

libsql-server/src/auth/user_auth_strategies/mod.rs
shopifyski: how about I add more variants into AuthError in auth/errors.rs ?

shopifyski: done

libsql-server/src/http/user/extract.rs
shopifyski: no just my way of tagging possible duplication, removed it

LucioFranco:
@shopifyski looks like some checks CI failures that need to be addressed! I just ran ci.

libsql-server/src/auth/user_auth_strategies/mod.rs
```diff
@@ -2,15 +2,68 @@ pub mod disabled;
...
+    pub fn new(scheme: &str, token: &str) -> UserAuthContext {
+        UserAuthContext {
+            scheme: Some(scheme.into()),
+            token: Some(token.into()),
+        }
+    }
+
+    pub fn from_auth_str(auth_string: &str) -> Result<Self, AuthError> {
+        let (scheme, token) = auth_string
+            .split_once(' ')
```
LucioFranco, line 63: Related to my other comment I see now what the problem is. I think we should support an empty auth string mostly because this is the workflow we have for a ton of tools outside of libsql and in tests. The problem is that now that the parsing of the header is separate from the authentication method we don't know that auth is disabled, because we should throw this error if jwt is enabled and we can't split. 

I think the solution here would be to not return an error here but somehow represent the state within the auth context of an empty auth header and let the auth strategy decide what to do with that. This would allow us to revert all the `dummy_token` stuff and fix the encryption test, as well as ensure that all our other tools continue to work locally.

libsql-server/src/auth/user_auth_strategies/mod.rs
```diff
@@ -2,15 +2,68 @@ pub mod disabled;
...
+        UserAuthContext {
+            scheme: Some("Bearer".into()),
+            token: Some(token.into()),
+        }
+    }
+
+    pub fn bearer_opt(token: Option<String>) -> UserAuthContext {
+        UserAuthContext {
+            scheme: Some("Bearer".into()),
+            token: token,
```
LucioFranco, line 50:
```suggestion
            token,
```

LucioFranco (changes requested):

libsql-server/src/auth/user_auth_strategies/mod.rs
LucioFranco: Related to https://github.com/tursodatabase/libsql/pull/1130#discussion_r1513163735

shopifyski: full agreement - in general terms, we should not require auth context for strategies that don't use it
I figured we could achieve it by lazy access - passing a function that calculates the context into authenticate
it took me down a massive rabbit hole of dynamic dispatch/mutability/ownership issues
decided it is not worth it, we can (and we will) revisit in later PRs
I propose we turn `authenticate(&self, context: UserAuthContext)` into `authenticate(&self, context: Result<UserAuthContext>)`
auth strategies that need the context will unwrap it and map errors accordingly
auth strategies that don't need it will simply ignore
bringing it up as it is not a common pattern in rust to accept Result as input 
it does fit the bill in this case very well
again, this is just a step in evolving auth layer towards more cohesive style

LucioFranco: Yeah, I think this makes a lot of sense. Lets do it.

libsql-server/src/auth/parsers.rs
```diff
@@ -34,11 +36,12 @@ pub fn parse_jwt_key(data: &str) -> Result<jsonwebtoken::DecodingKey> {
...
 
-pub(crate) fn parse_grpc_auth_header(metadata: &MetadataMap) -> Option<HeaderValue> {
+pub(crate) fn parse_grpc_auth_header(metadata: &MetadataMap) -> Result<UserAuthContext, AuthError> {
     metadata
         .get(GRPC_AUTH_HEADER)
-        .map(|v| v.to_bytes().expect("Auth should always be ASCII"))
-        .map(|v| HeaderValue::from_maybe_shared(v).expect("Should already be valid header"))
+        .ok_or(AuthError::AuthHeaderNotFound)
+        .and_then(|h| h.to_str().map_err(|_| AuthError::AuthHeaderNonAscii))
+        .and_then(|t| UserAuthContext::from_auth_str(t))
```
LucioFranco, line 44:
```suggestion
        .and_then(UserAuthContext::from_auth_str)
```

btw if the input type is the same as the closure this works as well!

LucioFranco (approved):
LGTM! Thanks!

PR #1131, opened 2024-03-04
prevent task enqueue when pending task

MarinPostma:
- Move migration related tables and ops to the schema module (`db.rs`)
- Prevent jobs for a schema from being enqueued, if there are already enqueued jobs for that schema
- Few robustness improverments
- more tests
- Some error handling

### DIFF
--- a/libsql-server/src/query_analysis.rs
+++ b/libsql-server/src/query_analysis.rs
@@ -153,6 +153,7 @@ impl StmtKind {
             // special case for `encoding` - it's effectively readonly for connections
             // that already created a database, which is always the case for sqld
             "encoding" => Some(Self::Read),
+            "schema_version" if body.is_none() => Some(Self::Read),
             // always ok to be served by primary
             "defer_foreign_keys" | "foreign_keys" | "foreign_key_list" | "foreign_key_check" | "collation_list"
             | "data_version" | "freelist_count" | "integrity_check" | "legacy_file_format"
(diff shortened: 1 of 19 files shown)


libsql-server/src/query_analysis.rs
```diff
@@ -180,7 +181,6 @@ impl StmtKind {
             | "read_uncommitted"
             | "recursive_triggers"
             | "reverse_unordered_selects"
-            | "schema_version"
```
haaawk, line 183: shouldn't this line be kept? isn't schema version with body still a write?

MarinPostma: schema version with body wasn't accepted before

libsql-server/src/schema/scheduler.rs
```diff
@@ -1,70 +1,104 @@
...
                             let job = self.current_job.take().unwrap();
                             assert_eq!(job.job_id(), job_id);
-                            self.namespace_store.meta_store().update_job_status(job_id, MigrationJobStatus::RunSuccess).await.unwrap();
+                            with_conn_async(self.migration_db.clone(), move |conn| {
+                                update_job_status(conn, job_id, MigrationJobStatus::RunSuccess)
+                            })
+                            .await
+                                .unwrap();
+                            // the current job is finished, try to pop next one out of the queue:
+                            self.has_work = true;
```
haaawk, line 79: I thought that has_work == true means a job is not fininshed?

MarinPostma: has work means that there may be some work to do next, and we should try enqueueing some. When we finish the current job, there may be more work, we raise the flag to try to enqueue the next one.

Issue #1133, opened 2024-03-05
Multi-Database Schema Migrations: add admin API to introspect shared schema


Horusiath:
Introduce an admin endpoints to be able to introspect shared schema:

- One part is to return general info about shared schema migrations.
- Another is to get more detailed information about progress status for a particular schema migration.

PR #1134, opened 2024-03-05
Migrations status admin api

Horusiath:
Covers #1133


MarinPostma (approved):
good for now our debug purposes, but let's revisit later

PR #1146, opened 2024-03-06
use custom query builder to collect migration error

MarinPostma:
Use `SchemaMigrationQueryBuilder` to collect migration error, and check migration task status

### DIFF
--- a/libsql-server/src/schema/migration.rs
+++ b/libsql-server/src/schema/migration.rs
@@ -3,8 +3,9 @@ use once_cell::sync::Lazy;
 use rusqlite::Savepoint;
 
 use crate::connection::program::{Program, Vm};
-use crate::query_result_builder::{IgnoreResult, QueryResultBuilder};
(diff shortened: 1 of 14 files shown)


libsql-server/src/schema/migration.rs
```diff
@@ -144,7 +147,20 @@ pub fn perform_migration<B: QueryResultBuilder>(
...
+        Ok(builder) => {
+            assert!(!builder.is_success());
+            let status = if dry_run {
+                savepoint.rollback().unwrap();
+                drop(savepoint);
+                MigrationTaskStatus::DryRunFailure
+            } else {
+                savepoint.commit().unwrap();
+                MigrationTaskStatus::Failure
+            };
```
haaawk, line 161: Wouldn't it be better to always rollback on failure?

haaawk (approved):

libsql-server/src/schema/migration.rs
MarinPostma: you're right

PR #1154, opened 2024-03-07
Fork a new DB when creating from shared schema

avinassh:


libsql-server/src/namespace/store.rs
```diff
@@ -366,6 +366,18 @@ impl NamespaceStore {
...
+        if let Some(shared_schema_name) = &db_config.shared_schema_name {
+            self.fork(
+                shared_schema_name.clone(),
+                namespace.clone(),
+                db_config.clone(),
+                None,
+            )
+            .await?;
+            let to_store = self.config_store(namespace).await?;
+            return to_store.store(db_config).await;
```
haaawk, line 378: I don't think we need this any more once fork takes the db config already, no?

haaawk (approved):

libsql-server/src/namespace/store.rs
avinassh: you are right. I kept it here thinking we don't flush inside Fork, but we fixed that yesterday. Let me update.

PR #1159, opened 2024-03-07
Add optional flush param when storing in MetaStore

avinassh:
Added a new method `store_and_maybe_flush` in MetaStore, which gives you an option to flush the config changes to disk.

The previous `store` method behaviour is not changed. It internally calls `store_and_maybe_flush` method, stores the config and flushes to disk.

This patch fixes a bug in fork. Earlier, we were flushing the target config before actually creating a fork. If the fork failed, MetaStore would have invalid data. Now, we store the config, without flush, create the fork. Iff successful, flush the target config to disk.


haaawk (approved):

haaawk:
We still have one test to fix it seems

MarinPostma (approved):

PR #1163, opened 2024-03-07
Handle dry run failure, and abort migration

MarinPostma:
abort migration on dry run failure


libsql-server/src/schema/status.rs
```diff
@@ -88,7 +90,9 @@ impl MigrationJob {
                     + self.progress(MigrationTaskStatus::Run)
             }
             MigrationJobStatus::DryRunSuccess => 0,
-            MigrationJobStatus::DryRunFailure => 0,
+            MigrationJobStatus::DryRunFailure => {
+                self.progress.iter().sum::<usize>() - self.progress(MigrationTaskStatus::Failure)
```
haaawk, line 94: Is it possible that one of the tasks failed the dry run but others succeeded so that we have N tasks and 
self.progress(MigrationTaskStatus::Failure) == 0
self.progress(MigrationTaskStatus::DryRunFailure) == 1
self.progress(MigrationTaskStatus::DryRunSuccess) == N - 1
?

MarinPostma: yes, and we're gonna abort them all. The final state when you're in dry run failure, is `Failure`, so any task that is not in that state needs to be failed. Hence, the number of remaining task you need to process is the sum of all the tasks that are not already in a failed state

MarinPostma: this is reflected in how we collect tasks here: https://github.com/tursodatabase/libsql/pull/1163/files#diff-06bbc51ac32833e3710b644fa85345f6631dcb4215149b24ff79d608107d887dR295-R302

haaawk (approved):

PR #1171, opened 2024-03-08
Add a test for a new db in shared schema

avinassh:
This patch just adds the tests. They are breaking and should pass when #1154 is fixed. 

(depends on changes from #1170)


avinassh:
tests moved to #1154

PR #1179, opened 2024-03-08
Namespace passing in http header

shopifyski:
Adding alternative to namespace inference from hostname prefix.
Utilizing NAMESPACE_METADATA_KEY to transfer namespace.
Tested with remote_sync example:

* when namespace injected - use namespace from header
* when namespace not injected - use namespace from hostname
cc: @LucioFranco @jeremywrowe @jameswritescode


libsql-server/src/hrana/ws/session.rs
```diff
@@ -65,60 +88,22 @@ pub enum ResponseError {
...
-        })
-    }
-    // todo dupe #auth
-    let namespace_jwt_key = server
-        .namespaces
-        .with(namespace.clone(), |ns| ns.jwt_key())
-        .await??;
-
-    session.auth = namespace_jwt_key
+    namespace_jwt_key
```
LucioFranco, line 101: nice 👍

libsql-server/src/http/user/db_factory.rs
```diff
@@ -46,18 +47,25 @@ pub fn namespace_from_headers(
...
-    let host = headers
-        .get("host")
-        .ok_or_else(|| Error::InvalidHost("missing host header".into()))?
-        .as_bytes();
-    let host_str = std::str::from_utf8(host)
-        .map_err(|_| Error::InvalidHost("host header is not valid UTF-8".into()))?;
+    headers
+        .get(NAMESPACE_METADATA_KEY)
+        .ok_or(Error::InvalidNamespace)
+        .and_then(|h| h.to_str().map_err(|_| Error::InvalidNamespace))
```
LucioFranco, line 53: would it make sense to add a `&'static str` to this error variant to specify what failed here? Otherwise we consume the string error here and may be hard to figure out why down the line.

libsql/src/replication/client.rs
```diff
@@ -58,7 +59,12 @@ impl Client {
             .try_into()
             .context("Invalid auth token must be ascii")?;
 
-        let ns = split_namespace(origin.host().unwrap()).unwrap_or_else(|_| "default".to_string());
+
+        let ns = maybe_namespace.unwrap_or_else(||
+            split_namespace(origin.host().unwrap())
+            .unwrap_or_else(|_| "default".to_string())
+        );
```
LucioFranco, line 66: this code is slightly hard to read, could we use `else {}` maybe?

LucioFranco (changes requested):
two small nits then lets merge

libsql-server/src/hrana/ws/session.rs
```diff
@@ -22,6 +22,29 @@ pub struct Session {
...
+            bail!(ProtocolError::NotSupported {
+                what: "Repeated hello message",
+                min_version: Version::Hrana2,
+            })
+        }
+        self.auth = auth;
+        Ok(())
+    }
+}
+
```
haaawk, line 47: Why is this here? This is already in main -> https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/hrana/ws/session.rs#L26

Could you rebase the change @shopifyski please?

libsql-server/src/hrana/ws/conn.rs
```diff
@@ -206,16 +206,18 @@ async fn handle_client_msg(conn: &mut Conn, client_msg: proto::ClientMsg) -> Res
...
-        None => {
-            session::handle_initial_hello(&conn.server, conn.version, jwt, conn.namespace.clone())
-                .await
-                .map(|session| conn.session = Some(session))
-        }
-        Some(session) => {
-            session::handle_repeated_hello(&conn.server, session, jwt, conn.namespace.clone()).await
-        }
-    };
+    let auth = session::handle_hello(&conn.server, jwt, conn.namespace.clone()).await;
```
haaawk, line 209: This is already in main as well, no? https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/hrana/ws/conn.rs#L209

It generally makes maintainer's work much easier if PRs are rebased and contain only new changes. I would be grateful if you could try to make sure that your PRs are rebased and clean. Thank you!

libsql-server/src/hrana/ws/session.rs
haaawk: This is already in main @LucioFranco -> https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/hrana/ws/session.rs#L101

libsql-server/src/http/user/db_factory.rs
haaawk, line 53: Wouldn't it be easier/cleaner to do:
```
headers
    .get(NAMESPACE_METADATA_KEY)
    .and_then(|h| h.to_str())
    .ok_or(Error::InvalidNamespace)
```
?

haaawk: I don't understand this whole block of code though. It contains errors but at the end it has `.or_else` which ignores those errors:

```
headers
        .get(NAMESPACE_METADATA_KEY)
        .ok_or(Error::InvalidNamespace) // <=== 1. translating any error to InvalidNamespace
        .and_then(|h| h.to_str().map_err(|_| Error::InvalidNamespace)) // <=== 2. possibly InvalidNamespace if header value can't be translated to string
        .and_then(|n| NamespaceName::from_string(n.into()))
        .or_else(|_| { // <=== ignoring both error from (1) and from (2)
            let host = headers
                .get("host")
                .ok_or_else(|| Error::InvalidHost("missing host header".into()))?
                .as_bytes();
            let host_str = std::str::from_utf8(host)
                .map_err(|_| Error::InvalidHost("host header is not valid UTF-8".into()))?;

            match split_namespace(host_str) {
                Ok(ns) => Ok(ns),
                Err(_) if !disable_default_namespace => Ok(NamespaceName::default()),
                Err(e) => Err(e),
            }
        })
```
What do I miss? How are `.ok_or(Error::InvalidNamespace)` and `.map_err(|_| Error::InvalidNamespace)` changing anything if the error gets ignore at the end by `.or_else(|_| {` step?

libsql/src/replication/client.rs
shopifyski: you're right, rewritten to more expressive form

libsql/src/database/builder.rs
```diff
@@ -226,6 +228,11 @@ cfg_replication! {
 
         }
 
+        pub fn namespace(mut self, namespace: &str) -> Builder<RemoteReplica> {
```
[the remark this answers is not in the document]
shopifyski: lemme know if that's better

libsql-server/src/http/user/db_factory.rs
shopifyski: yea, this code needs to change
need to define explicit rule of when to use namespace from header and when from hostname
@LucioFranco @haaawk 
what do you think about this?
* If get(NAMESPACE_METADATA_KEY) gives Some, use the header.
* Do not fallback to namespace from hostname in case of malformed header, just return Error
* If get(NAMESPACE_METADATA_KEY) gives None, use namespace from hostname
* If disable_default_namespace not set and any of the above fails, return default namespace
* if disable_default_namespace is set, any error produced by the above should be returned to caller

shopifyski: committed a new proposal

libsql/src/database/builder.rs
```diff
@@ -226,6 +228,12 @@ cfg_replication! {
 
         }
 
+        pub fn namespace<T>(mut self, namespace: T) -> Builder<RemoteReplica>
```
haaawk, line 231: Would `pub fn namespace(mut self, namespace: impl Into<String>) -> Builder<RemoteReplica>` do the same?

libsql/src/replication/client.rs
```diff
@@ -58,7 +59,15 @@ impl Client {
             .try_into()
             .context("Invalid auth token must be ascii")?;
 
-        let ns = split_namespace(origin.host().unwrap()).unwrap_or_else(|_| "default".to_string());
+        let ns = if let Some(ns_from_arg) = maybe_namespace
+        {
```
haaawk, line 63: Nit: this `{` should probably be at the end of the previous line

libsql-server/src/http/user/db_factory.rs
haaawk: The proposal sounds good to me. You said that you committed it but I don't see code in `namespace_from_headers` being changed, is it?

libsql/src/replication/client.rs
shopifyski: could that be configured so cargo fmt fixes it automatically?

haaawk: I would hope cargo fmt to fix it by default.

libsql/src/database/builder.rs
shopifyski: it wouldn't compile afaik

shopifyski: ur right, fixed

libsql-server/src/http/user/db_factory.rs
shopifyski: pushed

libsql-server/src/http/user/db_factory.rs
```diff
@@ -46,17 +47,27 @@ pub fn namespace_from_headers(
...
+        headers
+            .get("host")
+            .ok_or_else(|| Error::InvalidHost("missing host header".into()))
+            .and_then(|res| Ok(res.as_bytes()))
+            .and_then(|host| {
+                std::str::from_utf8(host)
+                    .map_err(|_| Error::InvalidHost("host header is not valid UTF-8".into()))
+            })
+            .and_then(|host_str| split_namespace(host_str))
+    };
```
haaawk, line 65: The behaviour here changed in two ways:
1. If host header is missing then previously we were always returning Err. New version returns Err only if disbale_default_namespace is true
2. If std::str::from_utf8(host) fails then previously we were always returning Err. New version returns Err only if disable_default_namespace is true

I think it is safer to keep the old behaviour. New behaviour is covering error cases like host error being set to wrong value.

shopifyski: I agree with it being safer not to fall back to default so aggressively
but we now get into an "exception driven" logic territory
e.g. error out when host header non utf, but fallback to default, when host header malformed
is there a reason for this special treatment of that particular error?

haaawk: Let's then fallback to default only when the header is missing and otherwise just error out with appropriate error. It seems safer and is more in line with the logic that was there before.

shopifyski: @haaawk, I restructured that section for backwards compatibility, let me know if you have further comments

libsql/src/database/builder.rs
```diff
@@ -226,6 +228,12 @@ cfg_replication! {
 
         }
 
+        pub fn namespace(mut self, namespace: impl Into<String>) -> Builder<RemoteReplica>
```
LucioFranco, line 231: This needs docs like the other constructors

LucioFranco (changes requested):
LGTM left one thing that needs to be added then we can merge

LucioFranco (approved):

MarinPostma:
This is a breaking change for the platform. The `x-namespace-bin` should be base64 encoded. We just hit this issue deploying a new canary image.

FYI, I'll patch that to base64 decode the header value.

PR #1183, opened 2024-03-09
block migration during db operation

MarinPostma:
- introduce schema lock
- synchronize migration with db creation

### DIFF
--- a/libsql-server/src/namespace/schema_lock.rs
+++ b/libsql-server/src/namespace/schema_lock.rs
@@ -1,0 +1,95 @@
+use std::sync::Arc;
+
+use hashbrown::{hash_map::Entry, HashMap};
+use parking_lot::Mutex;
+
+use super::NamespaceName;
+
+#[derive(Default)]
+pub struct SchemaLocksRegistry {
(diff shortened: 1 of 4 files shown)


libsql-server/src/namespace/schema_lock.rs
```diff
@@ -0,0 +1,95 @@
...
+    locks: Arc<Mutex<HashMap<NamespaceName, Arc<tokio::sync::RwLock<()>>>>>,
+}
+
+impl Drop for SchemaLock {
+    fn drop(&mut self) {
+        let mut locks = self.locks.lock();
+        match locks.entry(self.schema.clone()) {
+            Entry::Occupied(entry) => {
+                // there's only two ref left: the maps, and ours
+                if Arc::strong_count(entry.get()) == 2 {
```
Horusiath, line 30: If we cannot make this safe by compiler type checks, at least we should write comments in both the places pointing to this line, so it won't be forgotten during refactoring.

Horusiath (commented):
In general good, but SchemaLock looks a bit fragile in case when refactoring will be done.

PR #1185, opened 2024-03-11
Hrana: don't close the stream at the end of prepared statement execution

Horusiath:
This PR fixes the issue @haaawk and me originally noticed in libsql Go tests. It turns out, that the prepared statements - when calling `stmt.execute` - were always closing the stream, even if there was a transaction pending (created before prepare stmt).

I've also added standalone test that confirms the original issue.

### DIFF
--- a/libsql/src/hrana/hyper.rs
+++ b/libsql/src/hrana/hyper.rs
@@ -89,7 +89,7 @@ impl HttpSend for HttpSender {
         if let Ok(rt) = tokio::runtime::Handle::try_current() {
             rt.spawn(self.send(url, auth, body));
         } else {
-            log::warn!("tried to send request to `{url}` while no runtime was available");
+            tracing::warn!("tried to send request to `{url}` while no runtime was available");
         }
     }
 }
--- a/libsql/src/hrana/hyper.rs
+++ b/libsql/src/hrana/hyper.rs
@@ -16,6 +16,7 @@ use http::{HeaderValue, StatusCode};
 use hyper::body::HttpBody;
 use std::io::ErrorKind;
 use std::sync::Arc;
+use tracing::log;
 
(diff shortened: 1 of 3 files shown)


haaawk (approved):

haaawk:
Could you have a look please @LucioFranco?

libsql/src/hrana/mod.rs
```diff
@@ -134,7 +134,9 @@ where
             )),
             Some(Err(e)) => Err(e),
             Some(Ok(stmt)) => {
-                let close_stream = !matches!(stmt.kind, StmtKind::TxnBegin | StmtKind::TxnBeginReadOnly);
+                let in_tx_scope = !stream.is_autocommit()
```
LucioFranco, line 137: can we add a comment here explaining these conditionals?

LucioFranco (changes requested):

LucioFranco (approved):

libsql/src/hrana/hyper.rs
```diff
@@ -85,7 +85,11 @@ impl HttpSend for HttpSender {
     }
 
     fn oneshot(self, url: Arc<str>, auth: Arc<str>, body: String) {
-        tokio::spawn(self.send(url, auth, body));
+        if let Ok(rt) = tokio::runtime::Handle::try_current() {
+            rt.spawn(self.send(url, auth, body));
+        } else {
+            tracing::warn!("tried to send request to `{url}` while no runtime was available");
+        }
```
MarinPostma, line 92: in what situation wouldn't there be a runtime?

Horusiath: At some point `execute_transaction` suggested that by panicking from inside of that method.

PR #1187, opened 2024-03-11
allow ddl override

MarinPostma:
This pr allows overriding DLL block on shared schema database. This is necessary to provide an escape hatch if a migration fails for some reason, and individual database need manual intervention.

a new `ddl` claim is added to override the default ddl block on the given scopes. See test for example.


libsql-server/src/connection/libsql.rs
```diff
@@ -662,7 +663,7 @@ impl<W: Wal> Connection<W> {
                             || block_writes.load(Ordering::SeqCst)
                     }
                     StmtKind::DDL => {
-                        config.block_reads || config.block_writes || config.block_ddl()
```
haaawk, line 665: Can we now remove config.block_ddl()?

haaawk (approved):

PR #1198, opened 2024-03-12
move migration api to user api

MarinPostma:
move the job status endpoints to the user API. adds two routes:
- `GET /v1/jobs`: list the job for the schema namespace that the `host` header key is pointing to
- `GET /v1/jobs/:id`: returns details for the specific job id

requires read permission to the schema


libsql-server/tests/namespaces/shared_schema.rs
```diff
@@ -69,7 +69,9 @@ async fn check_data(ns: &str) -> Vec<String> {
 
 async fn http_get(url: &str) -> String {
     let client = Client::new();
-    client.get(url).await.unwrap().body_string().await.unwrap()
+    let resp = client.get(url).await.unwrap();
+    dbg!(resp.status());
```
haaawk, line 73: debug leftover?

haaawk (approved):
LGTM what header do I have to set to point to a specific namespace?

PR #1204, opened 2024-03-13
Namespace::destroy: add parameter to perform soft delete (preserve backup)

Horusiath:
Accordingly to @haaawk request: this PR extends DELETE `v1/namespace/:ns` API with optional `{"keep_backup": bool}` option. If it's set, the database will be deleted locally, but we make sure that its state is preserved in the backup.

~~Question: _while we don't delete the backup, we still have pretty much no out-of-the-box way of restoring the db using it (other than using bottomless-cli which will be troublesome) after it was deleted this way._~~

### DIFF
--- a/bottomless/src/completion_progress.rs
+++ b/bottomless/src/completion_progress.rs
@@ -60,10 +60,16 @@ impl SavepointTracker {
         let last_frame_no = self.next_frame_no.load(Ordering::SeqCst) - 1;
         let frame_no = *self.receiver.wait_for(|fno| *fno >= last_frame_no).await?;
         let generation = self.confirm_snapshotted().await?;
-        Ok(BackupThreshold {
+        let t = BackupThreshold {
             generation: generation.as_ref().map(Uuid::to_string),
             frame_no,
-        })
+        };
+        tracing::debug!(
+            "confirmed backup savepoint for generation `{:?}`, frame no.: {}",
+            t.generation,
+            t.frame_no
+        );
+        Ok(t)
     }
 }
 
--- a/bottomless/src/replicator.rs
+++ b/bottomless/src/replicator.rs
@@ -489,7 +489,10 @@ impl Replicator {
         tracing::debug!(
             "calling for backup savepoint for `{}` on generation `{}`, frame no.: {}",
             self.db_name,
-            self.generation().unwrap(),
+            match self.generation() {
+                Ok(gen) => gen.to_string(),
+                _ => "".to_string(),
+            },
             self.next_frame_no() - 1
         );
         if let Some(tx) = &self.flush_trigger {
--- a/bottomless/src/replicator.rs
+++ b/bottomless/src/replicator.rs
@@ -486,6 +486,12 @@ impl Replicator {
     }
 
     pub fn savepoint(&self) -> SavepointTracker {
+        tracing::debug!(
+            "calling for backup savepoint for `{}` on generation `{}`, frame no.: {}",
+            self.db_name,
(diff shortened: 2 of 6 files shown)


MarinPostma:
Can you try recreating a db for which you have a backup? I'm pretty sure it will restore

Horusiath:
You're right @MarinPostma. I've just checked it locally - it can be recreated and confirmed via:

1. Server:
```bash
export LIBSQL_BOTTOMLESS_DATABASE_ID=test-db
export LIBSQL_BOTTOMLESS_ENDPOINT=<aws-endpoint>
export LIBSQL_BOTTOMLESS_AWS_ACCESS_KEY_ID=<access_key_id>
export LIBSQL_BOTTOMLESS_AWS_SECRET_ACCESS_KEY=<secret_access_key>
export LIBSQL_BOTTOMLESS_AWS_DEFAULT_REGION="eu-central-2"

cargo r -- --meta-store-bucket-endpoint "http://localhost:9000" --meta-store-backup-interval-s 10 --meta-store-bucket-name metast --meta-store-backup-id mybackup --meta-store-region eu-central-2 --meta-store-secret-access-key minioadmin --meta-store-access-key-id minioadmin --backup-meta-store --enable-namespaces --enable-bottomless-replication --disable-default-namespace --admin-listen-addr 0.0.0.0:9091
```

2. Client
```bash
# create namespace with bottomless db id
curl -X POST http://localhost:9091/v1/namespaces/ns1/create -H 'Content-Type: application/json' -d '{"bottomless_db_id":"ns1"}'

# initialize some data on the db
curl -X POST http://ns1.localhost:8080/ -H 'Content-Type: application/json' -d '{"statements":[{"q":"create table t(a int);"},{"q":"insert into t(a) values(1);"}]}'

# confirm data is there
curl -X POST http://ns1.localhost:8080/ -H 'Content-Type: application/json' -d '{"statements":[{"q":"select * from t"}]}'

# soft delete namespace
curl -X DELETE http://localhost:9091/v1/namespaces/ns1 -H 'Content-Type: application/json' -d '{"keep_backup":true}'
```

3. Server:
```
# kill cargo process then remove local db to force restore from  bottomless
rm -rf ./data.sqld/

cargo r -- --meta-store-bucket-endpoint "http://localhost:9000" --meta-store-backup-interval-s 10 --meta-store-bucket-name metast --meta-store-backup-id mybackup --meta-store-region eu-central-2 --meta-store-secret-access-key minioadmin --meta-store-access-key-id minioadmin --backup-meta-store --enable-namespaces --enable-bottomless-replication --disable-default-namespace --admin-listen-addr 0.0.0.0:9091
```

4. Client:
```
# recreate the namespace with the same bottomless db id
curl -X POST http://localhost:9091/v1/namespaces/ns1/create -H 'Content-Type: application/json' -d '{"bottomless_db_id":"ns1"}'

# confirm that the database has been restored (it already contains backed up data)
curl -X POST http://ns1.localhost:8080/ -H 'Content-Type: application/json' -d '{"statements":[{"q":"select * from t"}]}'
```

haaawk:
So will I be able to just create this namespace in another sqld instance (providing the config) and that new instance will restore the namespace correctly?

Horusiath:
@haaawk yes, assuming that you'll provide correct config options for the bottomless to figure out where the backup is located. Bottomless uses `{db-id}` as an individual database backup identifier: it's given by the host and usually it looks like `ns-{namespace}:{group}`.

haaawk (approved):

PR #1207, opened 2024-03-13
bottomless: update S3 driver

Horusiath:
This PR updates S3 drivers to a stable version. For basically a year, we were using beta version and never updated it since.


MarinPostma:
@Horusiath how confident are you with this bump? anything that could bite us? if so I'd merge that after 0.24

Horusiath:
Feel free to wait with it. My main reason was that I wanted to add Turmoil connector to bottomless, but since current S3 API is outdated, there are version clashes, need for more refactorings in the future etc. It doesn't have to be a part of the next release.

haaawk:
@MarinPostma @LucioFranco do we want to merge this?

PR #1208, opened 2024-03-13
catch parser panic

MarinPostma:
catch parser panic until we can fix it

### DIFF
--- a/libsql-server/src/query_analysis.rs
+++ b/libsql-server/src/query_analysis.rs
@@ -332,15 +332,41 @@ impl Statement {
         // on the heap:
         // - https://github.com/gwenn/lemon-rs/issues/8
         // - https://github.com/gwenn/lemon-rs/pull/19
-        let mut parser = Box::new(Parser::new(s.as_bytes()).peekable());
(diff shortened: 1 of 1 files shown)


libsql-server/src/query_analysis.rs
```diff
@@ -332,15 +332,41 @@ impl Statement {
         // on the heap:
         // - https://github.com/gwenn/lemon-rs/issues/8
         // - https://github.com/gwenn/lemon-rs/pull/19
-        let mut parser = Box::new(Parser::new(s.as_bytes()).peekable());
+        let mut parser = Some(Box::new(Parser::new(s.as_bytes()).peekable()));
         let mut stmt_count = 0;
         std::iter::from_fn(move || {
+            // temporary macro to catch panic from the parser, until we fix it.
+            macro_rules! parse {
```
LucioFranco, line 339: do we really need a match for two uses 😂

LucioFranco (approved):

libsql-server/src/query_analysis.rs
MarinPostma: it gets very clumsy, ideally we get rid of that asap

PR #1211, opened 2024-03-14
Fix namespace deletion

MarinPostma:
- fix schema deletion
- optional ns delete payload

### DIFF
--- a/libsql-server/src/schema/db.rs
+++ b/libsql-server/src/schema/db.rs
@@ -95,6 +95,19 @@ pub(crate) fn has_pending_migration_jobs(
     Ok(has_pending)
 }
 
+pub(crate) fn schema_has_linked_dbs(
+    conn: &rusqlite::Connection,
+    schema: &NamespaceName,
+) -> Result<bool, Error> {
+    let has_linked = conn.query_row(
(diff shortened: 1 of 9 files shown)


libsql-server/src/schema/db.rs
```diff
@@ -95,6 +95,19 @@ pub(crate) fn has_pending_migration_jobs(
     Ok(has_pending)
 }
 
+pub(crate) fn schema_has_linked_dbs(
+    conn: &rusqlite::Connection,
+    schema: &NamespaceName,
+) -> Result<bool, Error> {
+    let has_linked = conn.query_row(
+        "SELECT count(1) FROM (SELECT 0 FROM shared_schema_links WHERE shared_schema_name = ? LIMIT 1)",
```
haaawk, line 103: Why not just `SELECT count(1) FROM shared_schema_links WHERE shared_schema_name = ? LIMIT 1`?

haaawk (approved):

libsql-server/src/schema/db.rs
MarinPostma: count always return a single row, while that would still work, it would perform a full db scan, whereas this query stops when it hits the first matching row, and returns 1 or 0

avinassh (approved):

PR #1217, opened 2024-03-14
libsql: fix malformed db and add test

LucioFranco:
Our current injection logic is hacked on top of the sqlite wal and connection. Due to some internal state of the connection and how we inject stuff, the connection ends up in a corrupt state reporting that the db is malformed even though it is not. The real solution is to rewrite the WAL in our custom way rather than layer on top of the sqlite wal implementation. This is a lot of work that we are working on but in the mean time we can just recreate the connection to avoid these issues.


libsql-replication/src/injector/mod.rs
```diff
@@ -33,6 +33,10 @@ pub struct Injector {
     // connection must be dropped before the hook context
     connection: Arc<Mutex<libsql_sys::Connection<InjectorWal>>>,
     biggest_uncommitted_seen: FrameNo,
+
+    path: PathBuf,
+    encryption_config: Option<libsql_sys::EncryptionConfig>,
+    auto_checkpoint: u32,
```
haaawk, line 39: It would be nice to have comments describing what the new fields are

libsql-replication/src/injector/mod.rs
```diff
@@ -138,7 +148,7 @@ impl Injector {
                 if let Some(e) = e.sqlite_error() {
                     if e.extended_code == LIBSQL_INJECT_OK {
                         // refresh schema
-                        connection.pragma_update(None, "writable_schema", "reset")?;
+                        // connection.pragma_update(None, "writable_schema", "reset")?;
```
haaawk, line 151: Why is that change needed?

libsql-replication/src/injector/mod.rs
```diff
@@ -161,7 +171,24 @@ impl Injector {
     }
 
     fn begin_txn(&mut self) -> Result<(), Error> {
-        let conn = self.connection.lock();
+        let mut conn = self.connection.lock();
```
haaawk, line 174: Could you add a description to the PR to explain why we need to replace the connection?

haaawk (approved):

libsql-replication/src/injector/mod.rs
LucioFranco: not needed backing this out

PR #1228, opened 2024-03-18
server: fix migration error hrana

LucioFranco:


glommer (commented):
I suspect this will not be enough. I tried attach (which is not a migration), and got a dogshit error as well.

@avinassh  - this is the issue with ws.

haaawk:
@MarinPostma please have a look at this PR

MarinPostma (approved):
you gotta love that fancy error handling 😄

MarinPostma:
@glommer says it's not enough though

haaawk:
Probably not enough but won't hurt

PR #1233, opened 2024-03-19
server: fix conflict on attach update

LucioFranco:


MarinPostma (approved):

athoscouto (approved):
We should probably update the one on the else branch as well

⛏️: I don't think you need the replace anymore
⛏️: we can use the same strategy to update shared schema links, we don't need the delete there

athoscouto (approved):

haaawk:
Anything preventing this from merge?

PR #1237, opened 2024-03-20
skipp inject_replication_index on checkpoint

Horusiath:
Apparently this helps with malformed db errors in databases under high load.


MarinPostma:
already in main

PR #1242, opened 2024-03-21
write queue cleanup

MarinPostma:


libsql-server/src/connection/connection_manager.rs
```diff
@@ -302,24 +313,12 @@ impl WrapWal<Sqlite3Wal> for ManagedConnectionWalWrapper {
...
-
-                    if let Some((id, unparker)) = next {
-                        tracing::debug!(line = line!(), "unparking id={id}");
-                        *current = Some((id, Instant::now(), false));
-                        unparker.unpark()
-                    } else {
-                        *current = None;
-                    }
+                    drop(current);
+                    self.release();
```
athoscouto, line 321: Is it a problem if a race condition happens between releasing and getting the current lock?

athoscouto (approved):
Nice! Take a look at my question though, not sure if something can go wrong between dropping the lock and reacquiring it on the release function.

libsql-server/src/connection/connection_manager.rs
MarinPostma: unless the slot is empty, it's only replaced on release, and only a connection is allowed to release itself, so it's safe :)

Issue #1224, opened 2024-03-18
Add timeout to embedded replica sync


penberg:
Let's add a timeout to embedded replica sync so that we don't hang a `sync()` call if there's a network issue.

PR #1248, opened 2024-03-22
Added timeout to embedded replica sync.

xonx4l:
closes issue-:#1224

Let's add a timeout to embedded replica sync so that we don't hang a sync() call if there's a network issue.


haaawk:
Thanks you for your contribution @xonx4l but this adds a timeout only to a test. It won't help at all with #1224.

xonx4l:
@haaawk thanks for the feedback could you please provide me more info on the issue  ??

haaawk:
Closing the PR due to inactivity

PR #1249, opened 2024-03-22
reenqueue stale reads on checkpoint

MarinPostma:
This PR force the re-enqueueing of waiting connection when a checkpoint request > RESTART is received. This causes all the waiting connections to refresh their read marks, and not block the checkpoint anymore.

### DIFF
--- a/libsql-server/src/connection/connection_manager.rs
+++ b/libsql-server/src/connection/connection_manager.rs
@@ -1,4 +1,5 @@
 use std::ops::Deref;
+use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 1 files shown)


libsql-server/src/connection/connection_manager.rs
```diff
@@ -126,12 +127,20 @@ impl ManagedConnectionWalWrapper {
...
             // function.
+            let new_token = self.manager.sync_token.load(Ordering::SeqCst);
+            if new_token != sync_token {
+                return Err(rusqlite::ffi::Error {
+                    code: ErrorCode::DatabaseBusy,
+                    extended_code: 517, // stale read
+                });
+            }
             if current.map_or(true, |(id, _, _)| id != self.id) && !enqueued {
+                sync_token = new_token;
```
haaawk, line 138: Is it ok not to do CompareAndSet here? There's never a race in which:
1. new_token == sync_token
2. sync_token is changed by someone else
3. we override sync_token with new_token
?

haaawk (approved):

PR #1251, opened 2024-03-22
libsql: add 5s keepalive for clients

LucioFranco:


haaawk:
Let's discuss this some more. I'm not sure it will help and I'm a bit afraid that it might cover up the real issue.

haaawk (approved):

PR #1262, opened 2024-03-26
bottomless: ability to replicate using temporary credentials

ryands17:
This PR adds the functionality to use temporary credentials to backup the database to S3. The `AWS_SESSION_TOKEN` can be optionally passed in to use temporary credentials locally or in an environment like EC2/EKS where roles are attached to the instance.


avinassh (approved):

avinassh:
@ryands17 could you please squash all commits into one? (we use merge queue, so I cannot do that from the PR)

PR #1267, opened 2024-03-27
server: add query stats to hrana responses

LucioFranco:


invisal:
Wow. This is game changing feature.

docs/HRANA_3_SPEC.md
```diff
@@ -1026,6 +1026,8 @@ type StmtResult = {
     "rows": Array<Array<Value>>,
     "affected_row_count": uint32,
     "last_insert_rowid": string | null,
+    "rows_read": uint64,
```
haaawk, line 1029: Since `affected_row_count` is uint32, shouldn't `rows_read` and `rows_written` be uint32 as well?

libsql/src/hrana/cursor.rs
```diff
@@ -134,6 +134,8 @@ where
                 affected_row_count,
                 last_insert_rowid,
                 replication_index: None,
+                rows_read: 0,
+                rows_written: 0,
```
haaawk, line 138: Why are those always 0 here?

libsql-hrana/src/proto.rs
```diff
@@ -22,6 +22,9 @@ pub struct PipelineRespBody {
     pub results: Vec<StreamResult>,
 }
 
+#[derive(Serialize, Deserialize, prost::Message)]
+pub struct Stats {}
```
haaawk, line 26: Where is this used?

libsql-server/src/connection/program.rs
```diff
@@ -105,7 +106,7 @@ impl<'a, B, F, S> Vm<'a, B, F, S>
 where
     B: QueryResultBuilder,
     F: Fn(&StmtKind) -> (bool, Option<String>),
-    S: Fn(String, &rusqlite::Statement, Duration),
+    S: Fn(String, u64, u64, u64, Duration),
```
haaawk, line 109: Why is this change needed? It seems that the whole PR would be much simpler if we kept accepting Statement here, no?

haaawk (approved):

docs/HRANA_3_SPEC.md
LucioFranco: I am not following for why it should match the uint32 of affected row count? Is it just to save the 4 extra bytes?

libsql/src/hrana/cursor.rs
LucioFranco: Cursors for now don't implement stats because its too complicated and not needed for the requirements set out by the requested users.

libsql-hrana/src/proto.rs
LucioFranco: I believe this was extra, removing it. :)

libsql-server/src/connection/program.rs
LucioFranco: I did this change because it made the most sense, we only want to request these stats once and then pass them to the callback and to the add_stats of the build. In addition, we want these values before we hit the callback. This simplification made the most sense to me rather than leaking a detail like this.

haaawk: I have nothing against this change but it is completely irrelevant to the goal of this PR. It's better to do it in a separate PR as a follow up. Otherwise I needed to spend 50% of time reviewing non-essential change. No need to do anything now but in the future having independent changes in independent PRs or at least commits is a desirable thing.

docs/HRANA_3_SPEC.md
haaawk: Just for consistency. if affected_row_count is uint32 and rows_written is uint64 then it is confusing. Does it mean max affected_row_count is 2^32 and max rows_written is 2^64? Maybe it's fine to have different types for those fields but it is confusing.

haaawk (approved):

docs/HRANA_3_SPEC.md
LucioFranco: lets solve this in a follow up. This feature is still hidden for now. Looks like there is both a uint64 and uint32 api for affected rows.

PR #1271, opened 2024-03-29
Jw/custom headers in auth

shopifyski:
every UserAuthStrategy now needs to also implement required_fields()
the caller is expected to make best effort to provide these headers, but it is not guaranteed
it is up to the UserAuthStrategy how to act in case of missing or malformed fields


haaawk:
Are you still working on this @shopifyski ? If not, please close the PR.

PR #1294, opened 2024-04-09
Add `enum` deserialization for sqlite Value

sveltespot:
Currently, if a sql row read needs to be deserialized into a struct, and the struct has one of the fields as a Rust enum type, then the deserialization fails. For example:

```rust
#[derive(Deserialize)]
struct MyRow {
  id: i32,
  status: Status
}

#[derive(Deserialize)]
enum Status {
  Pass,
  Fail
}

```
The deserialization of a row: `1, 'Pass'` will fail.

This commit adds the ability to deserialize a RowValue into an Enum type.


aminya (approved):

sveltespot:
@haaawk @LucioFranco Apologies for tagging. Could you help review and possibly merge this PR. It adds deserialization of a sqlite row into a struct that contains one of the fields as a unit Rust enum, which currently fails.

aminya:
@sveltespot Why did you close this PR? Could you reopen it so it is not lost?

LucioFranco (approved):
Thanks!

PR #1309, opened 2024-04-11
server: add `query_duration_ms` stats to query res

LucioFranco:


haaawk:
It seems that rust client library expects the new field and can't handle when old server does not send the new field in the response.

PR #1317, opened 2024-04-15
Prefetch frames during handshake

haaawk:
When the client already has a session token we can try to prefetch next_frames during handshake.

For a replica in Sydney and primary in US it cuts the sync time from 400ms to 200ms.

### DIFF
--- a/libsql/src/local/database.rs
+++ b/libsql/src/local/database.rs
@@ -90,6 +90,14 @@ impl Database {
 
         let endpoint = coerce_url_scheme(endpoint);
         let remote = crate::replication::client::Client::new(
+            connector.clone(),
+            endpoint.as_str().try_into().unwrap(),
+            auth_token.clone(),
+            version.as_deref(),
+            http_request_callback.clone(),
+        )
+        .unwrap();
+        let secondary_remote = crate::replication::client::Client::new(
             connector,
             endpoint.as_str().try_into().unwrap(),
             auth_token,
@@ -98,7 +106,7 @@ impl Database {
         )
         .unwrap();
         let path = PathBuf::from(db_path);
-        let client = RemoteClient::new(remote.clone(), &path)
+        let client = RemoteClient::new(remote.clone(), secondary_remote, &path)
             .await
             .map_err(|e| crate::errors::Error::ConnectionFailed(e.to_string()))?;
 
--- a/libsql/src/local/database.rs
+++ b/libsql/src/local/database.rs
@@ -92,20 +92,17 @@ impl Database {
         let endpoint = coerce_url_scheme(endpoint);
         let remote = crate::replication::client::Client::new(
             connector.clone(),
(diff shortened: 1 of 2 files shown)


penberg:
@haaawk Wouldn't extending the protocol to only require one message for this use case be more maintainable? That is, make it possible to only send a `LogOffset` message for sync?

haaawk:
> @haaawk Wouldn't extending the protocol to only require one message for this use case be more maintainable? That is, make it possible to only send a `LogOffset` message for sync?

It would but this way we can give users improvement without forcing them to update their production dbs. We should work on extending the protocol to allow only a single message but in the meantime I have this to give users improvement straight away.

libsql/src/local/database.rs
```diff
@@ -90,6 +90,14 @@ impl Database {
 
         let endpoint = coerce_url_scheme(endpoint);
         let remote = crate::replication::client::Client::new(
+            connector.clone(),
+            endpoint.as_str().try_into().unwrap(),
+            auth_token.clone(),
+            version.as_deref(),
+            http_request_callback.clone(),
+        )
+        .unwrap();
```
LucioFranco, line 99: You will probably need to merge to get these fixes as well for the unwraps https://github.com/tursodatabase/libsql/commit/d24435e564db895dc9578eada7b8bbcabdc88df5

libsql/src/replication/remote_client.rs
```diff
@@ -7,33 +7,38 @@ use libsql_replication::frame::{Frame, FrameHeader, FrameNo};
...
 };
 use tokio_stream::Stream;
 use tonic::metadata::AsciiMetadataValue;
+use tonic::{Response, Status};
 use zerocopy::FromBytes;
 
 /// A remote replicator client, that pulls frames over RPC
 pub struct RemoteClient {
     remote: super::client::Client,
+    secondary_remote: super::client::Client,
```
LucioFranco, line 20: Why does this need a secondary client, why can't we use the other client?

LucioFranco (changes requested):

libsql/src/local/database.rs
haaawk: The change is based on main. It seems that you missed this place. I will add a commit that fixes it.

libsql/src/replication/remote_client.rs
haaawk: because both `self.remote.replication.hello` and `self.remote.replication.batch_log_entries` borrow mut self and rust screems at me when I try to do:
```
let (hello, frames) = tokio::join!(
                self.remote.replication.hello(hello_req),
                self.remote.replication.batch_log_entries(log_offset_req)
            );
```

libsql/src/local/database.rs
haaawk: Added a commit that removes the unwraps

MarinPostma (approved):

Issue #1329, opened 2024-04-19
column names differ from what sqlite would produce, confusing clients.


glommer:
Ref: https://github.com/fermyon/feedback/issues/52

the issue is that the column names we produce are different than what sqlite produces. This can lead to confusing behavior if the user tries to get the columns by name.

For the query "select count(*) from foo", we receive the query as-is:

```
2024-04-19T00:02:07.626840Z DEBUG request{method=POST uri=/v2/pipeline version=HTTP/1.1}: libsql_server::hrana::http: pipeline:{ Hrana2, Execute(ExecuteStreamReq { stmt: Stmt { sql: Some("select count(*) from foo"), sql_id: None, args: [], named_args: [], want_rows: Some(true), replication_index: Some(1) } }) }
```

when it is time to `execute_query`, we produce: 
```
2024-04-19T00:02:07.627181Z DEBUG libsql_server::connection::program: executing query: SELECT count (*) FROM foo;
```

Somewhere in the query parsing we end up adding a space.

PR #1330, opened 2024-04-19
use user query string for queries rather than generated

MarinPostma:
rather than using a query string reconstructed from the parsed AST, use the user query string.

Fixes https://github.com/tursodatabase/libsql/issues/1329


glommer:
Marin, please use the "Fixes #1329" syntax in the description so it can auto close the issue

haaawk (approved):

PR #1331, opened 2024-04-20
Fix bottomless replication subproject link in libsql-server readme

marcob:
I was reading the doc and thought I'd fix these.


avinassh:
Thank you! @marcob

PR #1336, opened 2024-04-22
server: parallelize namespace shutdown

LucioFranco:


MarinPostma (approved):

haaawk:
I'm wondering if this unbounded concurrency will work for 100k or 1M namespaces?

LucioFranco:
@haaawk should be limited by the amount of threads available on the host and the runtime. If anything most of them will be sleeping waiting for io etc, I think it should be fine.

haaawk:
> @haaawk should be limited by the amount of threads available on the host and the runtime. If anything most of them will be sleeping waiting for io etc, I think it should be fine.

Ok. When you will be testing please make sure you use at least 1k or 10k namespaces and see that it works

PR #1343, opened 2024-04-25
libsql-wal: SealedSegment

MarinPostma:
This PR introduces `SealedSegment`. A sealed segment is a WAL segment that is immutable. It contains its own index and a set of potentially duplicated frames.

note: I removed all terminology referring to `log`, for better clarity. The `log` is now synonymous with `wal`, and it's made of segments of different types.


libsql-wal/src/segment/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+
+pub mod list;
+pub mod sealed;
+
+#[repr(C)]
+#[derive(Debug, zerocopy::AsBytes, zerocopy::FromBytes, zerocopy::FromZeroes, Clone, Copy)]
+pub struct SegmentHeader {
+    start_frame_no: U64,
+    last_commited_frame_no: U64,
+    db_size: U32,
```
haaawk, line 19: Why is db_size only 32-bit number? Isn't 2GB too small?

libsql-wal/src/segment/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+            .checked_sub(self.start_frame_no.get() - 1)
+            .unwrap_or(0) as usize
+    }
+
+    pub fn last_committed(&self) -> u64 {
+        // either the current segment is empty, and the start frame_no is the last frame_no commited on
+        // the previous segment (start_frame_no - 1), or it's the last committed frame_no from this
+        // segment.
+        if self.is_empty() {
+            self.start_frame_no.get() - 1
```
haaawk, line 77: Can self.start_frame_no be 0?

libsql-wal/src/segment/mod.rs
MarinPostma: Db size in SQLite terminology is in pages

LucioFranco (approved):

libsql-wal/src/segment/mod.rs
```diff
@@ -0,0 +1,160 @@
+//! Libsql-wal is organized as a linked list of segments. Frames are appended to the head segments,
```
avinassh:
```suggestion
//! Libsql-wal is organized as a linked list of segments. Frames are appended to the head segment,
```

libsql-wal/src/segment/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+//! and eventually, the head segment is swapped for a new empty one. The previous head segment is
+//! sealed and becomes immutable. The head segment is represented by the `CurrentSegment` type, and
+//! the sealed segments by the `SealedSegment` type.
+//!
+//! When a reader starts a transaction, it record the head segment current frame_no. This is the
```
avinassh:
```suggestion
//! When a reader starts a transaction, it records the head segment current frame_no. This is the
```

avinassh (approved):

libsql-wal/src/segment/sealed.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub fn read_frame_offset(&self, offset: u32, frame: &mut Frame) -> Result<()> {
+        let offset = frame_offset(offset);
+        self.file.read_exact_at(frame.as_bytes_mut(), offset as _)?;
+        Ok(())
+    }
+
+    pub fn read_page(&self, page_no: u32, max_frame_no: u64, buf: &mut [u8]) -> Result<bool> {
+        if self.header().last_commited_frame_no.get() > max_frame_no {
+            return Ok(false);
+        }
```
avinassh, line 130: you should check if `self.header().start_frame_no > max_frame_no` no?

avinassh (approved):

libsql-wal/src/segment/list.rs
```diff
@@ -0,0 +1,173 @@
...
+    segment: SealedSegment<F>,
+    next: ArcSwapOption<SegmentLink<F>>,
+}
+
+pub struct SegmentList<F> {
+    head: ArcSwapOption<SegmentLink<F>>,
+    len: AtomicUsize,
+    /// Whether the segment list is already being checkpointed
+    checkpointing: AtomicBool,
+}
```
avinassh, line 24: may be its just me, but I find `SegmentLink` and `SegmentList` confusing. I would probably call `SegmentLink` as `SegmentNode` may be, cos most linked list implementations use the term `node`

haaawk (approved):

libsql-wal/src/segment/list.rs
```diff
@@ -0,0 +1,173 @@
...
+
+impl<F> SegmentList<F> {
+    /// Prepend the list with the passed sealed segment
+    pub fn push_log(&self, segment: SealedSegment<F>) {
+        let segment = Arc::new(SegmentLink {
+            segment,
+            next: self.head.load().clone().into(),
+        });
+
+        self.head.swap(Some(segment));
```
avinassh, line 44: these two lines are not atomic, that is okay? i was expecting to see operation similar to CAS

avinassh (approved):

libsql-wal/src/segment/list.rs
MarinPostma: if that can unconfuse you, np

libsql-wal/src/segment/list.rs
MarinPostma: I don't understand your concern. What do you expect and why?

avinassh: Is following possible:

1. t1: Call to `push_log`, the first line gets executed `let segment = Arc::new...`. This `segment` points to the old head `h1`
2. t2: (Another thread), same call to `push_log`, this thread runs both the instructions, now the head points to the new head
3. t3: the first thread gets resumed, inserts the old `segment` object

(my rust is weak, so I was wondering if this can happen)

PR #1356, opened 2024-04-29
server: return 429 when dbcreatetimeout is hit

LucioFranco:


avinassh (approved):

athoscouto:
@LucioFranco we should probably use 408 instead

PR #1359, opened 2024-04-30
server: add config update txn_timeout

LucioFranco:


haaawk (approved):
Probably will only affect new connections but it's better than what we have now.

PR #1364, opened 2024-05-01
implement wal interface

MarinPostma:
implement the sqlite Wal interface using the `SharedWal`

### DIFF
--- a/libsql-wal/src/wal.rs
+++ b/libsql-wal/src/wal.rs
@@ -1,0 +1,303 @@
+use std::ffi::OsStr;
+use std::os::unix::prelude::OsStrExt;
+use std::sync::atomic::AtomicU64;
+use std::sync::Arc;
+
+use libsql_sys::wal::{Wal, WalManager};
+
+use crate::fs::FileSystem;
+use crate::registry::WalRegistry;
+use crate::shared_wal::SharedWal;
+use crate::transaction::Transaction;
+
+#[derive(Clone)]
+pub struct LibsqlWalManager<FS: FileSystem> {
+    pub registry: Arc<WalRegistry<FS>>,
+    pub next_conn_id: Arc<AtomicU64>,
+}
+
+pub struct LibsqlWal<FS: FileSystem> {
+    last_read_frame_no: Option<u64>,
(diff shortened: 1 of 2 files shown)


libsql-wal/src/wal.rs
```diff
@@ -0,0 +1,303 @@
...
+
+        tracing::debug!(invalidate_cache, "read started");
+        Ok(invalidate_cache)
+    }
+
+    #[tracing::instrument(skip_all, fields(id = self.conn_id))]
+    fn end_read_txn(&mut self) {
+        match self.tx.take() {
+            Some(Transaction::Read(tx)) => tx,
+            Some(Transaction::Write(tx)) => tx.downgrade(),
```
avinassh, line 115: in which situation a write transaction can call `end_read_txn`?

libsql-wal/src/wal.rs
```diff
@@ -0,0 +1,303 @@
...
+    #[tracing::instrument(skip_all, fields(id = self.conn_id))]
+    fn undo<U: libsql_sys::wal::UndoHandler>(
+        &mut self,
+        handler: Option<&mut U>,
+    ) -> libsql_sys::wal::Result<()> {
+        match self.tx {
+            Some(Transaction::Write(ref mut tx)) => {
+                if tx.is_commited() {
+                    return Ok(());
+                }
```
avinassh, line 196: shouldn't this return an error since you are calling undo on a committed transaction

avinassh (approved):

libsql-wal/src/wal.rs
MarinPostma: Usually, SQLite will end the write and then end the read, but in OG SQLite WAL implementation, ending a read transaction also ends the write

libsql-wal/src/wal.rs
MarinPostma: That's what I initially thought, but SQLite can call undo after a commit, and that's noop. I found that with fault injection.

libsql-wal/src/wal.rs
MarinPostma: Although ending a write txn does not end a read txn

PR #1365, opened 2024-05-01
server: flag to destroy metastore on restore err

LucioFranco:
This adds a new flag that when set to true will destroy the metastore and start with a fresh db file when restore runs into issues. This is a temp fix to for larger metastore restore issues that include disk malformed issues that we have seen.

### DIFF
--- a/libsql-server/src/namespace/meta_store.rs
+++ b/libsql-server/src/namespace/meta_store.rs
@@ -291,6 +291,7 @@ impl MetaStoreInner {
 
                 Err(e) => {
                     tracing::error!("meta store restore failed: {}", e);
+
                     return Err(Error::from(e));
                 }
             }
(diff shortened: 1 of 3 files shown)


athoscouto (approved):

haaawk:
Why is this not merged @LucioFranco ?

haaawk:
I took the liberty to resolve conflicts and push the updated change. @LucioFranco please have a look if I didn't make any mistakes and let's get this merged if we still want this

haaawk:
@athoscouto Lucio told me that this was done on your request. Do we still want/need this?

libsql-server/src/namespace/meta_store.rs
```diff
@@ -399,7 +400,58 @@ impl MetaStore {
...
+
+        let destroy_on_error = config.destroy_on_error;
+
+        let inner = match MetaStoreInner::new(base_path, conn, wal_manager, config.clone()).await {
+            Ok(inner) => inner,
+            Err(e) => {
+                if destroy_on_error {
+                    let db_path = base_path.join("metastore");
+
+                    tracing::info!(
```
athoscouto, line 412: We should probably have a warning-level log telling us this operation will happen or is done, so we can detect it easily.

athoscouto (commented):
We want this to be able to prevent a crash loop when metastore fails to restore from S3
If we don't have this, then when restore fails sqld will keep on restarting and we have to change images to fix it

PR #1370, opened 2024-05-03
add query timing

ajayvish0:
This is my first time contributing to the project, if i made some mistake please guide me.

### DIFF
--- a/bindings/c/src/lib.rs
+++ b/bindings/c/src/lib.rs
@@ -446,7 +446,7 @@ pub unsafe extern "C" fn libsql_execute_stmt(
             set_err_msg(format!("Error executing statement: {}", e), out_err_msg);
             2
         }
-    }
+    }let end_time = start_time.elapsed();
 }
 
 #[no_mangle]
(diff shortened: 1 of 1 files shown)


bindings/c/src/lib.rs
```diff
@@ -446,7 +446,7 @@ pub unsafe extern "C" fn libsql_execute_stmt(
             set_err_msg(format!("Error executing statement: {}", e), out_err_msg);
             2
         }
-    }
+    }let end_time = start_time.elapsed();
```
haaawk, line 449: I don't think this will compile.

bindings/c/src/lib.rs
```diff
@@ -472,13 +472,19 @@ pub unsafe extern "C" fn libsql_query(
...
         Ok(rows) => {
+            let end_time = start_time.elapsed(); // Calculate elapsed time
+            println!("Query execution time: {:?}", end_time);
+            
             let rows = Box::leak(Box::new(libsql_rows { result: rows }));
             *out_rows = libsql_rows_t::from(rows);
         }
         Err(e) => {
+            let end_time = start_time.elapsed(); // Calculate elapsed time
+            println!("Query execution failed. Time elapsed: {:?}", end_time);
```
haaawk, line 487: It is a bad practise to print like that in libraries. Please use tracing instead.

haaawk:
Closing this PR due to inactivity

PR #1371, opened 2024-05-05
Updated to Worker 0.2.0, Fixed an example and a minor issue

spigaz:


haaawk:
Thanks for your contribution @spigaz! Could you please say some more about what was wrong with the path?

image

haaawk:
I extracted non-controversial part to https://github.com/tursodatabase/libsql/pull/1604
and controversial part to https://github.com/tursodatabase/libsql/pull/1605
I will merge the non-controversial part and let's keep discussing the controversial one

spigaz:
> Could you please say some more about what was wrong with the path?

To be honest I don't remember, I was trying to get this example running:
https://github.com/spigaz/workers-rs-axum-tursodb

Normally it should be something obvious, otherwise I would have commented it.

I can try to figure that out latter today with the example, but if you try and you just can't, I believe you can simply ignore it.

PR #1374, opened 2024-05-06
Support multiple jwt keys for authentication

david-mccullars:
Modify the auth strategy for libsql-server to support multiple JWT keys. This will allow server administrators to more gracefully rotate keys by adding new keys without having to remove old ones simultaneously. Once all clients have been migrated to using new keys, the old one can be safely removed.

The modified auth strategy will accept a JWT if any of the public keys can successfully decrypt it. However, if none can (or all error), the last error will be returned.

As defined in [RFC 1422](https://datatracker.ietf.org/doc/html/rfc1422) a *.PEM file by definition can contain multiple items, so to provide multiple JWT keys one needs only to concatenate them together. For example:

```
-----BEGIN PUBLIC KEY-----
MCowBQYDK2VwAyEAz78yKnNWXkSQJUdQW2TU3WdelH2KtifEzg27BdtIL7c=
-----END PUBLIC KEY-----
-----BEGIN PUBLIC KEY-----
MCowBQYDK2VwAyEAKg/aT1QW16CtOKFqEvj2kvxODjWOBA6iYPDRnzrbLJ0=
-----END PUBLIC KEY-----
-----BEGIN PUBLIC KEY-----
MCowBQYDK2VwAyEARi1RRwfK5qpUf1leN9Qtjt9lRIBuNDAz6AE8PdnoSCc=
-----END PUBLIC KEY-----
```

This should work whether the multiple certificates are passed via the `SQLD_AUTH_JWT_KEY` environment variable or the `--auth-jwt-key-file` argument.


libsql-server/src/auth/user_auth_strategies/jwt.rs
```diff
@@ -80,6 +88,20 @@ mod jwt_time {
     }
 }
 
+fn validate_any_jwt(
+    jwt_keys: &Vec<jsonwebtoken::DecodingKey>,
+    jwt: &str,
+) -> Result<Authenticated, AuthError> {
+    for (index, jwt_key) in jwt_keys.iter().enumerate() {
```
jeremywrowe, line 95: Should this be limited to two keys so that no iteration needs to be done and encapsulate the concept some how. 

EG:
validate_jwt(&primary_jwt_key, jwt) || validate_jwt(&secondary_jwt_key, jwt);

david-mccullars: It's a good question. Here was my thinking:

1) For the (99%) key rotation use case it should only be in this state for a relatively short period of time (maybe a few hours or a day tops). Otherwise it'll be in the singleton case where we use the `DecodingKeyContainer::Single` to avoid iteration per Turso's suggestion.
2) There may be use cases that I'm not thinking about where 3 or more keys are actually desired. I couldn't come up with a good reason to artificially block this.

libsql-server/src/auth/user_auth_strategies/jwt.rs
```diff
@@ -7,8 +7,13 @@ use crate::{
 
 use super::{UserAuthContext, UserAuthStrategy};
 
+pub enum DecodingKeyContainer {
```
shopifyski, line 10: It won't be long until we forget why we implemented this single/multiple distinction
it is a seed of complexity -> it requires forked logic to handle
any chance we could have one more round of conversations about how much real overhead a vec! introduces?

libsql-server/src/auth/user_auth_strategies/jwt.rs
```diff
@@ -80,6 +88,20 @@ mod jwt_time {
     }
 }
 
+fn validate_any_jwt(
+    jwt_keys: &Vec<jsonwebtoken::DecodingKey>,
+    jwt: &str,
+) -> Result<Authenticated, AuthError> {
+    for (index, jwt_key) in jwt_keys.iter().enumerate() {
+        let result = validate_jwt(&jwt_key, jwt);
+        if result.is_ok() || index == jwt_keys.len() - 1 {
```
shopifyski, line 97: we're swallowing individual errors
we should at least log a warning if result is Err

libsql-server/src/auth/user_auth_strategies/jwt.rs
```diff
@@ -80,6 +88,20 @@ mod jwt_time {
...
+            return result;
+        }
+    }
+
+    Err(AuthError::Other)
```
shopifyski, line 102: in case validation fails, we might be interested in individual errors for each jwt
I suggest that we collect all errors an fold them into a single one, something like https://stackoverflow.com/a/61171459

libsql-server/src/auth/user_auth_strategies/jwt.rs
```diff
@@ -138,6 +164,23 @@ mod tests {
         (encoding_key, decoding_key)
     }
 
+    fn key_pairs(
```
shopifyski, line 167: maybe rename to generate_key_pairs
<verb>_<object> naming convention is preferred for functions
key_pair() function should also be renamed
<object> only names are best left for getters

libsql-server/src/auth/parsers.rs
```diff
@@ -160,4 +173,117 @@ mod tests {
...
+                        header: _,
+                        claims: _
+                    })
+                ));
+
+                let jsonwebtoken::TokenData { header, claims } = decoded.unwrap();
+                assert_eq!(header.alg, jsonwebtoken::Algorithm::EdDSA);
+
+                let claims = format!("{:?}", claims);
+                assert_eq!(claims, "Token { id: None, a: None, p: Some(Authorized { read_only: Some(Scopes { namespaces: Some({example1c, example1a, example1b}), tags: None }), read_write: None, read_only_attach: None, read_write_attach: None, ddl_override: None }), exp: Some(+33713-01-29T21:47:33Z) }");
```
shopifyski, line 205: std::fmt::Debug does not guarantee the ordering of fields
maybe better to make assertions on individual claims fields
that way we can ignore fields that we're not concerned about
and such individual assertions could be easily reusable

libsql-server/src/auth/user_auth_strategies/jwt.rs
david-mccullars: The problem is that those errors are expected. If I have to different keys, only one of them can decrypt the JWT, so the other one will error out. If we log them it will get very noisy without adding value.

libsql-server/src/auth/user_auth_strategies/jwt.rs
david-mccullars: When validation fails, we'll return the last error. The only reason we would ever get this generic one is when there are no JWT keys - which really shouldn't happen.

libsql-server/src/auth/user_auth_strategies/jwt.rs
david-mccullars: I'm okay with renaming both, but we should keep them consistent. It would be weird to have one be `key_pair` and the other `generate_key_pairs`.

libsql-server/src/auth/user_auth_strategies/jwt.rs
david-mccullars: I'm okay removing the complexity if everyone prefers that after looking at the enum version. Could be a classic case of premature optimization. Fwiw, I do think it should remaining somewhat intuitive that the vast majority of the time there should only be a single key and hence a motivation for this bit of weirdness.

libsql-server/src/auth/user_auth_strategies/jwt.rs
david-mccullars: Another thought here is do we want to expose the fact that there are multiple keys to clients? That feels like an internal piece of data that really shouldn't be leaked.

libsql-server/src/auth/parsers.rs
david-mccullars: I opened up (crate pub) the `p` field from `Token` in order to validate that. Not sure if that's going to be a point of concern or not.

libsql-server/src/auth/user_auth_strategies/jwt.rs
shopifyski: let's let the others chime in then

libsql-server/src/auth/user_auth_strategies/jwt.rs
shopifyski: maybe debug log then? to turn it on only when troubleshooting?

libsql-server/src/auth/user_auth_strategies/jwt.rs
MarinPostma: I second @shopifyski here, I'd rather have a single code path, the overhead of allocating a vector once in a while is negligible

david-mccullars: As an extra data point, I was exploring today what it would look like with just `Vec<jsonwebtoken::DecodingKey>` everywhere except for the `DecodingKeyContainer` enum. The immediate instinct was to rename all `jwt_key` to `jwt_keys` since it felt weird to do things like `pub fn parse_jwt_key(data: &str) -> Result<Vec<DecodingKey>>` or `let key: Vec<DecodingKey> = ...`. However, as the thread kept on pulling I started to run into some backwards-compatibility issues, specifically with the `jwt_key` (singular) request variable in libsql-server/src/http/admin/mod.rs. We can certainly halt the cascade of singular-to-plural changes anywhere we like, but at some point we may have something gross like `config.jwt_keys = req.jwt_key`.

MarinPostma: We could extend the config with `jwt_keys`, and merge `jwt_key` and `jwt_keys` as late as possible when building the decoding object. We would mark the `jwt_key` field as deprecated, and cleanup sometime in the future, would that contain the changes?

libsql-server/src/main.rs
```diff
@@ -63,8 +64,11 @@ struct Cli {
     /// APIs. The key is either a PKCS#8-encoded Ed25519 public key in PEM, or just plain bytes of
     /// the Ed25519 public key in URL-safe base64.
     ///
+    /// It is possible to provide multiple JWT decoding keys in a single file by concatenating them
+    /// together. All decoding keys will be tried when parsing incoming JWT's.
+    ///
     /// You can also pass the key directly in the env variable SQLD_AUTH_JWT_KEY.
-    #[clap(long, env = "SQLD_AUTH_JWT_KEY_FILE")]
+    #[clap(long, env = "SQLD_AUTH_JWT_KEY_FILE", num_args = 1..)]
```
MarinPostma, line 71: what's that for?

MarinPostma (commented):
Hey folks, very sorry it took me so much time to get to it! I forgot about it after brazil. Mostly looks good to me, left a few comments. Thanks for all the tests

libsql-server/src/main.rs
david-mccullars: Good catch! This was done before implementing support for multi key strings and then forgotten. I'll remove.

libsql-server/src/auth/user_auth_strategies/jwt.rs
david-mccullars: I believe it's reachable if the `jwt_keys` Vec is empty.

libsql-server/src/auth/user_auth_strategies/jwt.rs
david-mccullars: After giving this more thought I believe the intent - and desired behavior - is to do our very best to communicate to a user why their request is failing. As such, would should try and avoid obfuscating the real error with a generic one if possible. My thought process is that if we receive any error other than `InvalidJwt` then it means that we were able to parse the incoming JWT with the current key but that it is either expired or immature. There shouldn't be any reason to keep trying other keys in this scenario and we should just communicate "hey, your jwt is expired" right away. Otherwise, if we go through all of the keys but all return back `InvalidJwt` then we can just return that.

libsql-server/src/auth/user_auth_strategies/jwt.rs
david-mccullars: I opted to keep the stuff outside of libsql-server as is. It didn't feel as bizarre as I had feared to draw the line here.

libsql-server/scripts/gen_jwt_test_assets.py
MarinPostma, line 1: Sorry, I wasn't clear, I mean generate to keys and tokens from rust so that we don't need assets 🤦 But that ok, if we need to edit those tests in the future, we'll port that to rust 👍

MarinPostma (approved):

PR #1386, opened 2024-05-13
bottomless2 boilerplate

MarinPostma:


bottomless2/'
```diff
@@ -0,0 +1,31 @@
+use crate::{NamespaceName, StoreSegmentRequest};
```
LucioFranco, line 1: You committed a file called `'` 😂

bottomless2/src/lib.rs
```diff
@@ -0,0 +1,179 @@
+#![allow(dead_code, unused_variables, async_fn_in_trait)]
```
LucioFranco, line 1: whats up with the async fn in trait stuff? Probably complaining about the send issues?

bottomless2/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+pub struct BottomlessLoop<S: Storage> {
+    receiver: mpsc::Receiver<StoreSegmentRequest<S::Config>>,
+    scheduler: Scheduler<S>,
+    max_in_flight: usize,
+    in_flight_futs: JoinSet<JobResult<S>>,
+    force_shutdown: oneshot::Receiver<()>,
+}
+
+impl<S: Storage + 'static> BottomlessLoop<S> {
+    /// Schedules durability jobs. This loop is not allowed to fail, or loose jobs.
```
LucioFranco, line 33:
```suggestion
    /// Schedules durability jobs. This loop is not allowed to fail, or lose jobs.
```

bottomless2/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+    /// A job is prepared by calling `Scheduler::prepare(..)`. The job is spawned, and it returns a
+    /// `JobResult`, which is then returned to the scheduler by calling `Scheduuler::report(..)`.
```
LucioFranco, line 35:
```suggestion
    /// `JobResult`, which is then returned to the scheduler by calling `Scheduler::report(..)`.
```

bottomless2/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+                    match msg {
+                        Some(req) => {
+                            self.scheduler.register(req);
+                        }
+                        None => {
+                            shutting_down = true;
+                        }
+                    }
+                }
+                Some(join_result) = self.in_flight_futs.join_next(), if !self.in_flight_futs.is_empty() => {
```
LucioFranco, line 70: I think we should consider making this a `biased` (info in https://docs.rs/tokio/latest/tokio/macro.select.html) so that these always get processed first instead of enqueuing things to prevent starvation. If the mpsc queue is always full it can be trying to enqueue more where I think we should always prioritize getting the results out first.

bottomless2/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+                            if shutting_down {
+                                tracing::info!("processed job, {} jobs remaining", self.in_flight_futs.len());
+                            }
+                            self.scheduler.report(job_result);
+                        }
+                        Err(e) => {
+                            // job panicked. report and exit process. The program is crippled, from
+                            // now on, so we just exit, and hope to restart on a fresh state.
+                            tracing::error!("fatal error: bottomless job panicked: {e}");
+                            std::process::exit(1);
```
LucioFranco, line 83: should we make this choice at a higher layer? Seems a bit extreme here to abort the whole process at a layer this deep.

bottomless2/src/storage/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+    namespace: NamespaceName,
+    /// If provided, will restore up to the most recent segment lesser or equal to `before`
+    before: Option<DateTime<Utc>>,
+}
+
+pub struct DbMeta {
+    max_frame_no: u64,
+}
+
+pub trait Storage: Send + Sync {
```
LucioFranco, line 30: I think we can use https://docs.rs/trait-variant/0.1.2/trait_variant/ to make the async fn sendable.

LucioFranco (approved):
Left some comments but overall LGTM

bottomless2/src/lib.rs
MarinPostma: linter complains because the trait is public API

bottomless2/src/storage/mod.rs
MarinPostma: we'll see it's that's necessary later

PR #1388, opened 2024-05-13
Proof of concept of CDC on libsql-server

athoscouto:
Uses [SQLite's update hooks](https://sqlite.org/c3ref/update_hook.html) to allow users to listen for database changes.
Listening is pull-based, based on SSE.

This aims to offer a simple (implementation and API) and efficient (stateless, no need to query the table, nor store multiple versions of data) solution.
Users who want stricter guarantees should be able to combine this with other features to achieve their goals.
For instance, to get at least once delivery you can create a trigger that appends to an event table and listen to inserts to it.
To recover from crashes, clients can store the last processed entry somewhere (e.g. another libsql-server table).

---

Quirks to document and/or address:
- This will only work for rowid tables. Which excludes any virtual tables.
- This will not notify users of whole table deletions. From the link above:
> In the current implementation, the update hook is not invoked when conflicting rows are deleted because of an [ON CONFLICT REPLACE](https://sqlite.org/lang_conflict.html) clause. Nor is the update hook invoked when rows are deleted using the [truncate optimization](https://sqlite.org/lang_delete.html#truncateopt). The exceptions defined in this paragraph might change in a future release of SQLite.

- Updates are broadcasted when they are run, not when they are committed.

---

Before merging this we should discuss the ideal API.
My plan is to use [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) to notify listeners of any changes.
This is kind of implemented, I need to add the proper headers to make it 100% though.
Sending single-line JSON objects to notify changes seems like a good streaming strategy, it allows us to expand the API in the future and add arbitrary data.

Example of request and response:
```
➜  libsql git:(main) ✗ curl http://localhost:6060/beta/listen -H "Host: main.turso" -v
* Host localhost:6060 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:6060...
* Connected to localhost (::1) port 6060
> GET /beta/listen HTTP/1.1
> Host: main.turso
> User-Agent: curl/8.5.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< access-control-allow-origin: *
< vary: origin
< vary: access-control-request-method
< vary: access-control-request-headers
< transfer-encoding: chunked
< date: Sun, 12 May 2024 17:54:07 GMT
< 
{"db":"main","table":"t3","rowid":1}
{"db":"main","table":"t3","rowid":1}
{"db":"main","table":"t3","rowid":1}
{"db":"main","table":"t3","rowid":1}
{"db":"main","table":"t3","rowid":1}
```

### DIFF
--- a/libsql-server/src/broadcaster.rs
+++ b/libsql-server/src/broadcaster.rs
@@ -1,10 +1,7 @@
 use std::{
     collections::{hash_map::Entry, HashMap},
     mem,
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        Arc,
-    },
+    sync::Arc,
 };
 
 use parking_lot::Mutex;
@@ -32,7 +29,6 @@ fn is_zero(num: &u64) -> bool {
 pub struct BroadcasterInner {
     state: Mutex<HashMap<String, BroadcastMsg>>,
     senders: Mutex<HashMap<String, broadcast::Sender<BroadcastMsg>>>,
-    active: AtomicBool,
 }
 
 #[derive(Debug, Default, Clone)]
@@ -43,14 +39,7 @@ pub struct Broadcaster {
 impl Broadcaster {
     const BROADCAST_CAP: usize = 1024;
 
-    pub fn active(&self) -> bool {
-        self.inner.active.load(Ordering::Relaxed)
-    }
-
     pub fn notify(&self, table: &str, action: rusqlite::hooks::Action) {
-        if !self.active() {
-            return;
-        }
         let mut state = self.inner.state.lock();
         if let Some(entry) = state.get_mut(table) {
             Self::increment(entry, action);
@@ -71,9 +60,6 @@ impl Broadcaster {
     }
 
     pub fn commit(&self) {
-        if !self.active() {
-            return;
-        }
         let senders = self.inner.senders.lock();
         for (table, entry) in self.flush_changes() {
             if let Some(sender) = senders.get(&table) {
@@ -89,9 +75,6 @@ impl Broadcaster {
     }
 
     pub fn rollback(&self) {
-        if !self.active() {
-            return;
-        }
         self.flush_changes();
     }
 
@@ -101,7 +84,6 @@ impl Broadcaster {
             Entry::Vacant(entry) => {
                 let (sender, receiver) = broadcast::channel(Self::BROADCAST_CAP);
                 entry.insert(sender);
-                self.inner.active.store(true, Ordering::Relaxed);
                 receiver
             }
         };
@@ -109,15 +91,16 @@ impl Broadcaster {
         BroadcastStream::new(receiver)
     }
 
-    pub fn unsubscribe(&self, table: &String) {
+    pub fn unsubscribe(&self, table: &String) -> bool {
         let mut tables = self.inner.senders.lock();
         if let Some(sender) = tables.get(table) {
             if sender.receiver_count() == 0 {
                 tables.remove(table);
                 if tables.is_empty() {
-                    self.inner.active.store(false, Ordering::Relaxed);
+                    return false;
                 }
             }
         }
+        return true;
     }
 }
--- a/libsql-server/src/broadcaster.rs
+++ b/libsql-server/src/broadcaster.rs
@@ -1,5 +1,6 @@
 use std::{
-    collections::{hash_map::Entry, HashMap},
+    collections::{hash_map::Entry, BTreeMap, HashMap},
+    mem,
     sync::Arc,
 };
 
@@ -27,7 +28,7 @@ fn is_zero(num: &u64) -> bool {
 
 #[derive(Debug, Default)]
 pub struct BroadcasterInner {
-    state: Mutex<HashMap<String, BroadcastMsg>>,
+    state: Mutex<BTreeMap<String, BroadcastMsg>>,
     senders: Mutex<HashMap<String, broadcast::Sender<BroadcastMsg>>>,
 }
 
@@ -61,7 +62,10 @@ impl Broadcaster {
 
     pub fn commit(&self) {
         let senders = self.inner.senders.lock();
-        self.inner.state.lock().drain().for_each(|(table, entry)| {
+        let mut state = self.inner.state.lock();
+        let state = mem::take(&mut *state);
+
+        state.into_iter().for_each(|(table, entry)| {
             if let Some(sender) = senders.get(&table) {
                 let _ = sender.send(entry);
             }
--- a/libsql-server/src/broadcaster.rs
+++ b/libsql-server/src/broadcaster.rs
@@ -1,0 +1,126 @@
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+};
+
+use parking_lot::Mutex;
+use serde::{Deserialize, Serialize};
+use tokio::sync::broadcast::{self};
+use tokio_stream::wrappers::BroadcastStream;
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(tag = "action", rename_all = "snake_case")]
+pub enum BroadcastMsg {
+    Commit,
+    Rollback,
+    #[serde(untagged)]
+    Change {
+        action: Action,
+        rowid: i64,
+    },
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Hash, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+pub enum Action {
+    UNKNOWN,
+    DELETE,
+    INSERT,
+    UPDATE,
+}
+
+impl From<rusqlite::hooks::Action> for Action {
+    fn from(value: rusqlite::hooks::Action) -> Self {
+        match value {
+            rusqlite::hooks::Action::SQLITE_DELETE => Action::DELETE,
+            rusqlite::hooks::Action::SQLITE_INSERT => Action::INSERT,
+            rusqlite::hooks::Action::SQLITE_UPDATE => Action::UPDATE,
+            _ => Action::UNKNOWN,
+        }
+    }
+}
+
+impl From<&str> for Action {
+    fn from(value: &str) -> Self {
+        match value {
+            "delete" => Action::DELETE,
+            "insert" => Action::INSERT,
+            "update" => Action::UPDATE,
+            _ => Action::UNKNOWN,
+        }
+    }
+}
+
+pub struct UpdateSubscription {
+    pub inner: BroadcastStream<BroadcastMsg>,
+}
+
+const BROADCAST_CAP: usize = 1024;
+
+#[derive(Debug, Default)]
+pub struct BroadcasterInner {
+    senders: Mutex<HashMap<String, broadcast::Sender<BroadcastMsg>>>,
+    active: AtomicBool,
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct Broadcaster {
+    inner: Arc<BroadcasterInner>,
+}
+
+impl Broadcaster {
+    pub fn active(&self) -> bool {
+        self.inner.active.load(Ordering::Relaxed)
+    }
+
+    pub fn notify(&self, table: &str, msg: BroadcastMsg) {
+        if !self.active() {
+            return;
+        }
+        self.inner
+            .senders
+            .lock()
+            .get(table)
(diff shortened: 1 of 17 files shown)


glommer:
My main question is whether the restrictions make this useful enough. My intuition is that the rowid restriction is ok, but the deletion one, will make the API not that useful. @notrab very curious to hear your thoughts.

notrab:
I really like the direction of this! Super work guys. 

**One thing that is a concern to me is triggering the events when ran not committed.**

This would mean anyone tailing the event stream could end up with events that aren't a true representation of the database state. If someone is tailing new users created to send an email, configure some third party or whatever, they're going to have a bad time. 

....Perhaps we handle that logically internally just using commits, OR at a worst case trigger another event on the rollback hook?

The `rowid` thing is also a valid concern. It would be useful to listen to full table operators without `WHERE` but it's probably not a deal breaker.

As for the API design... It looks good so far but we could potentially send the delta if the user requests it with a header or something...

MarinPostma:
@notrab I don't know if that's a big issue, since we don't send any content, it's just a false positive, so one would ready the row, and realise it hasn't changed, which I don't think is a big deal

notrab:
@MarinPostma yeah so I guess I wrote my comment based on the fact the data was sent. I think that's where this becomes really valuable to users.

If we don't, we should at least include the rowid, table name, database name, and what occured (create/update/delete)... which could still have the problem I was describing.

libsql-server/src/http/user/listen.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    let (reader, writer) = tokio::io::duplex(8 * 1024);
+
+    BLOCKING_RT.spawn_blocking(|| {
+        futures::executor::block_on(async move {
+            let mut writer = SyncIoBridge::new(writer);
+            // TODO: stop when request is cancelled
+            while recv_listen_msg(&mut writer, &mut rx, &action).await {}
+        })
+    });
```
MarinPostma, line 54: I think all of that would be better handled by: https://docs.rs/axum-extra/0.9.3/axum_extra/json_lines/struct.JsonLines.html

MarinPostma: all of this could be made very concise with https://docs.rs/tokio-stream/0.1.15/tokio_stream/wrappers/struct.BroadcastStream.html. something like (pseudo rust):

```rust
let stream = BroadcastStream::new(rx).filter(|event| event.action == action).map(|event| SubResponse::from(event));
JsonLines(stream)
```

libsql-server/src/broadcaster.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    pub fn subscribe(&self, table: &str) -> broadcast::Receiver<UpdateHookMsg> {
+        let mut senders = self.senders.lock();
+        if let Some(sender) = senders.get(table) {
+            return sender.subscribe();
+        }
+
+        let (sender, receiver) = broadcast::channel(16);
+        senders.insert(table.to_owned(), sender);
```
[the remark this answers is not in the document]
athoscouto:
```rust
    pub fn subscribe(&self, table: String) -> UpdateSubscription {
        let mut senders = self.senders.lock();
        let sender = senders.entry(table.clone()).or_insert_with(|| {
            let (sender, _) = broadcast::channel(16);
            sender
        });
        self.stream(table, sender.subscribe())
    }
```

like this? still worth it with the extra `clone()` and `subscribe()` calls?

glommer:
Got curious: would a javascript user consume this via `fetch` ?
How do they develop against this if they are using a `file` API ?

notrab:
@glommer @athoscouto if we're using server-sent-events, we could hook into the `EventSource` API to give developers a CLI to listen and forward events, as well as specific libraries to attach events.

I'd love to contribute the following when I return from vacation:

## 1. CLI

```bash
npx libsql-listen libsql://my-db-org.turso.io --token [TOKEN]
```

This would then forward events to a PORT. Perhaps `3000` by default since that's a popular one for JS developers. Users could override this by passing a custom `--forward-to` flag or something. 

This is useful because it can simulate the "webhook" experience by creating a proxy between your local development environment and the event stream.

## 2. Node Client

Something as basic as this could work for our MVP:

```ts
import { libSQLClient } from "@libsql/events"

const libsql = new libSQLClient({
  url: 'libsql://my-db-org.turso.io'
  authToken: '...'
})

libsql.on('SOMETHING', () => {
  // ...
})
```

I haven't looked at the type of events yet, but we could format those events nicely in the handler.

Creating this is one (max two) days work.

Lmkwyt

athoscouto:
I had to change the design a bit (or a lot).
The first concern was performance, we want to keep the overhead as small as possible on database operations.

When doing batch changes, the overhead of serializing and sending events with `rowid` was bigger than running the change itself.
This caused the broadcaster channel to drop a lot of events.
To prevent event dropping we could add backpressure (by making the channel sync), but this would hit our first concern.

If we aggregate events before sending them, then we don't see the same event-dropping problem.
If users want row granularity when listening to updates, they can and should use triggers to persist changes to a table.
Then they can query that table to see what changed.
This pattern is good because it allows users to incur very little overhead for general ("something changed") updates and opt-in on extra overhead if more granularity is needed.
Triggers also allow a very fine-grained control, allow users to create change entries conditionally if needed, and even select old and new versions of changed fields.

athoscouto:
From my tests, the biggest bottleneck of the current implementation is abusing `broadcast::Sender.send()`.
We can avoid this by aggregating messages before sending them to the broadcast channel. 
I'll leave this for a future iteration though, since it is an optimization we can make without breaking the current API.

libsql-server/src/http/user/listen.rs
```diff
@@ -0,0 +1,184 @@
...
+                    Ok(item) => match item {
+                        BroadcastMsg::Change { action, .. } => self.register(action),
+                        BroadcastMsg::Rollback => self.clear(),
+                        BroadcastMsg::Commit => self.flush(),
+                    },
+                    Err(_) => self.error(),
+                },
+            },
+        }
+    }
```
LucioFranco, line 150: Why do you have a ton of manual poll implementation? Can you use `async_stream` instead?

athoscouto: Because I need to aggregate several items from upstream into one (or maybe none) to send downstream.
Took a quick look at `async_stream` and couldn't figure out how to use it for this purpose. Do you have an example?

If this isn't a blocker, I'd rather not change this for now. On the next iteration, I want to nuke this. 
The broadcaster channel is the biggest perf bottleneck, so I want to do the aggregation before sending data through it.
This means we won't need to aggregate at the stream level, since we do it before.

libsql-server/src/http/user/listen.rs
```diff
@@ -0,0 +1,184 @@
...
+
+struct SubscriptionAggregator {
+    actions: Vec<Action>,
+    subscription: UpdateSubscription,
+    state: AggregatorState,
+    store: NamespaceStore,
+    namespace: NamespaceName,
+    table: String,
+    errored: bool,
+}
```
MarinPostma, line 88: I wonder: why don't you aggregate at the source? That would make things so much more efficient...

libsql-server/src/http/user/listen.rs
```diff
@@ -0,0 +1,184 @@
...
+    async fn new(
+        store: NamespaceStore,
+        namespace: NamespaceName,
+        table: String,
+        actions: Vec<Action>,
+    ) -> crate::Result<Self> {
+        let subscription = store
+            .subscribe(namespace.clone(), table.clone())
+            .await
+            .unwrap();
```
MarinPostma, line 100: this unwrap can fail

libsql-server/src/http/user/listen.rs
```diff
@@ -0,0 +1,184 @@
...
+            namespace,
+            table,
+            errored: false,
+        })
+    }
+}
+
+impl Drop for SubscriptionAggregator {
+    fn drop(&mut self) {
+        let namespace = mem::take(&mut self.namespace);
```
MarinPostma, line 115: namespace can be cloned

MarinPostma (commented):
I would suggest that we aggregate events at the source. The update hook increments the action, and the commit hook sends the aggregated data for dispatch. I think that would simplify things significantly, and reduce the overhead. With this design we are doing the aggregations multiple times (for each listener).

libsql-server/src/broadcaster.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub delete: u64,
+    #[serde(skip_serializing_if = "is_zero")]
+    pub insert: u64,
+    #[serde(skip_serializing_if = "is_zero")]
+    pub update: u64,
+}
+
+fn is_zero(num: &u64) -> bool {
+    *num == 0
+}
```
MarinPostma, line 26: 😄

libsql-server/src/broadcaster.rs
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    pub fn commit(&self) {
+        let senders = self.inner.senders.lock();
+        for (table, entry) in self.flush_changes() {
+            if let Some(sender) = senders.get(&table) {
+                _ = sender.send(entry);
+            }
+        }
```
MarinPostma, line 68: you could reuse that alloc by using `HashMap::drain` instead

libsql-server/src/http/user/listen.rs
```diff
@@ -0,0 +1,137 @@
...
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "snake_case")]
+enum AggregatorEvent {
+    Error(&'static str),
+    #[serde(untagged)]
+    Changes(BroadcastMsg),
+}
+
+struct Subscription {
+    store: Option<NamespaceStore>,
```
MarinPostma, line 81: why is that an option?

libsql-server/src/namespace/store.rs
```diff
@@ -469,6 +477,22 @@ impl NamespaceStore {
...
+        &self,
+        namespace: NamespaceName,
+        table: String,
+    ) -> BroadcastStream<BroadcastMsg> {
+        self.inner.broadcasters.subscribe(namespace, table)
+    }
+
+    pub(crate) async fn unsubscribe(&self, namespace: NamespaceName, table: &String) {
+        self.inner.broadcasters.unsubscribe(namespace, table);
+    }
```
MarinPostma, line 494: this doesn't need to be async

libsql-server/src/http/user/listen.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+impl Drop for Subscription {
+    fn drop(&mut self) {
+        if let Some(store) = self.store.take() {
+            let namespace = mem::take(&mut self.namespace);
+            let table = mem::take(&mut self.table);
+            let _ = tokio::spawn(async move {
+                store.unsubscribe(namespace, &table).await;
+            });
```
MarinPostma, line 93: this doesn't need to be async, so you don't need the `Option` for the store.

libsql-server/src/broadcaster.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub unknown: u64,
+    #[serde(skip_serializing_if = "is_zero")]
+    pub delete: u64,
+    #[serde(skip_serializing_if = "is_zero")]
+    pub insert: u64,
+    #[serde(skip_serializing_if = "is_zero")]
+    pub update: u64,
+}
+
+fn is_zero(num: &u64) -> bool {
```
LucioFranco, line 23:
```suggestion
#[inline(always)]
fn is_zero(num: &u64) -> bool {
```
may want to do this

libsql-server/src/http/user/listen.rs
```diff
@@ -0,0 +1,129 @@
...
+        };
+
+        let mut stream = store.subscribe(namespace.clone(), table.clone());
+
+        while let Some(item) = stream.next().await  {
+            match item {
+                Ok(msg) => if filter_actions(&msg, &actions) {
+                    yield AggregatorEvent::Changes(msg);
+                },
+                Err(_) => yield AggregatorEvent::Error(&LAGGED_MSG),
```
LucioFranco, line 110: could record a metric for this

LucioFranco (approved):

libsql-server/src/http/user/listen.rs
athoscouto: image

PR #1390, opened 2024-05-13
Add an admin endpoint to checkpoint

avinassh:
- [ ] Check safety: are we calling outside of bottomless? can this cause corruption?

---

e.g. usage 

```
curl -X POST http://localhost:8080/v1/namespaces/b16397ca-d793-11ee-b439-acde48001123/checkpoint -H "Content-Type: application/json" -d {} -v
```

### DIFF
--- a/libsql-server/src/http/admin/mod.rs
+++ b/libsql-server/src/http/admin/mod.rs
@@ -127,6 +127,10 @@ where
             "/v1/namespaces/:namespace/create",
             post(handle_create_namespace),
         )
+        .route(
+            "/v1/namespaces/:namespace/checkpoint",
+            post(handle_checkpoint),
+        )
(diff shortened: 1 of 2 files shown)


libsql-server/src/http/admin/mod.rs
```diff
@@ -429,3 +433,32 @@ async fn handle_delete_namespace<C>(
...
+
+#[derive(Deserialize, Default)]
+struct CheckpointReq {
+    #[serde(default)]
+    pub force: bool,
+}
+
+async fn handle_checkpoint<C>(
+    State(app_state): State<Arc<AppState<C>>>,
+    Path(namespace): Path<String>,
```
MarinPostma, line 445: there is an extractor for namespace name :)
```suggestion
    Path(namespace): Path<NamespaceName>,
```

PR #1399, opened 2024-05-17
fix metastore deadlock

MarinPostma:
This PR fixes many issues that caused the metastore and bottomless to deadlock.

- blocking on `wait_for_snapshot`: instead of blocking to wait for snapshot, we return busy if the previous snapshot isn't ready yet. That prevents blocking when bottomless becomes unavailable.
- timeout on `wait_for_committed`: if all the send snapshot tasks failed, wait for committed would wait forever for an event that would never happen; since jobs are not enqueued, the whole connection would stale: checkpoint would wait for a recent enough committed index, and send loop would wait for checkpoint to flush new frames => deadlock. Instead, we timed out so that the connection would be throttled due to unavailability, but we can still make progress.
- Meta store deadlock: there was a slim chance that `MetaStore::exists` would wait to acquire the lock on the configs, while we were also holding a long living lock on the `MetaInner` when inserting a new entry, and bottomless was unavailable. If the reactor happened to be on the same thread as the waiting thread, then the `wait_for_committed` timeout would never fire, and this would cause a deadlock. Finer-grained locking solves the issue in the short term, but the design is flawed, and this code needs to be rewritten.

### DIFF
--- a/bottomless/src/bottomless_wal.rs
+++ b/bottomless/src/bottomless_wal.rs
@@ -137,9 +137,9 @@ impl<T: Wal> WrapWal<T> for BottomlessWalWrapper {
                     replicator.skip_snapshot_for_current_generation();
                     return Err(Error::new(SQLITE_BUSY));
                 }
-                
+
                 let fut = tokio::time::timeout(std::time::Duration::from_secs(1), replicator.wait_until_committed(last_known_frame));
-                
+
                 match runtime.block_on(fut) {
                     Ok(Ok(_)) => (),
                     Ok(Err(e)) => {
--- a/bottomless/src/bottomless_wal.rs
+++ b/bottomless/src/bottomless_wal.rs
@@ -138,7 +138,10 @@ impl<T: Wal> WrapWal<T> for BottomlessWalWrapper {
                     return Err(Error::new(SQLITE_BUSY));
                 }
 
-                let fut = tokio::time::timeout(std::time::Duration::from_secs(1), replicator.wait_until_committed(last_known_frame));
+                let fut = tokio::time::timeout(
+                    std::time::Duration::from_secs(1),
+                    replicator.wait_until_committed(last_known_frame),
+                );
 
                 match runtime.block_on(fut) {
                     Ok(Ok(_)) => (),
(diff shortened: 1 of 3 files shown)


bottomless/src/replicator.rs
```diff
@@ -477,6 +477,30 @@ impl Replicator {
...
+                tracing::debug!("Not snapshotting, the main db file does not exist or is empty");
+                let _ = self.snapshot_notifier.send(Ok(Some(generation)));
+                return Ok(false);
+            }
+            tracing::debug!("waiting for generation snapshot {} to complete", generation);
+            let current = self.snapshot_waiter.borrow();
+            let snapshotted = match &*current {
+                Ok(Some(gen)) => *gen == generation,
+                Ok(None) => false,
+                Err(_) => false,
```
haaawk, line 492: NIt this could be `_ => false,` no?

bottomless/src/replicator.rs
```diff
@@ -477,6 +477,30 @@ impl Replicator {
         self.use_compression
     }
 
+    pub async fn is_snapshotted(&mut self) -> Result<bool> {
```
haaawk, line 480: This function does not seem to ever return Err. Can we just make it `-> bool`?

bottomless/src/bottomless_wal.rs
```diff
@@ -137,22 +137,41 @@ impl<T: Wal> WrapWal<T> for BottomlessWalWrapper {
...
-                    );
-                    return Err(Error::new(SQLITE_IOERR_WRITE));
+                let snapshotted = runtime.block_on(replicator.is_snapshotted());
+                match snapshotted {
+                    Ok(true) => (),
+                    Ok(false) => {
+                        tracing::warn!("previous generation not snapshotted, skipping checkpoint");
+                        return Err(Error::new(SQLITE_BUSY));
+                    }
+                    Err(e) => {
```
haaawk, line 167: is_snapshotted() never returns Err. Maybe we can make it return just `bool` and then get rid of this leg?

haaawk (approved):

LucioFranco (approved):

PR #1401, opened 2024-05-17
bottomless scheduler

MarinPostma:
implement scheduler

### DIFF
--- a/bottomless2/src/job.rs
+++ b/bottomless2/src/job.rs
@@ -1,14 +1,23 @@
+use std::sync::Arc;
+
 use crate::storage::Storage;
 use crate::Result;
-use crate::{NamespaceName, StoreSegmentRequest};
+use crate::StoreSegmentRequest;
+
+/// A request, with an id
+#[derive(Debug)]
+pub(crate) struct IndexedRequest<C> {
+    pub(crate) request: StoreSegmentRequest<C>,
+    pub(crate) id: u64,
+}
 
 /// A storage Job to be performed
+#[derive(Debug)]
 pub(crate) struct Job<S: Storage> {
-    storage: S,
-    namespace: NamespaceName,
+    pub(crate) storage: Arc<S>,
     /// Segment to store.
-    /// TODO: implement request batching (merge segment and send).
-    request: StoreSegmentRequest<S::Config>,
+    // TODO: implement request batching (merge segment and send).
+    pub(crate) request: IndexedRequest<S::Config>,
 }
 
 impl<S: Storage> Job<S> {
@@ -25,7 +34,7 @@ impl<S: Storage> Job<S> {
 
 pub(crate) struct JobResult<S: Storage> {
     /// The job that was performed
-    job: Job<S>,
+    pub(crate) job: Job<S>,
     /// The outcome of the job: the new durable index, or an error.
-    result: Result<u64>,
+    pub(crate) result: Result<u64>,
 }
--- a/bottomless2/src/lib.rs
+++ b/bottomless2/src/lib.rs
@@ -1,5 +1,6 @@
 #![allow(dead_code, unused_variables, async_fn_in_trait)]
 use std::path::PathBuf;
+use std::sync::Arc;
 use std::time::Duration;
 
 use chrono::{DateTime, Utc};
@@ -65,12 +66,12 @@ impl<S: Storage + 'static> BottomlessLoop<S> {
                             if shutting_down {
                                 tracing::info!("processed job, {} jobs remaining", self.in_flight_futs.len());
                             }
-                            self.scheduler.report(job_result);
+                            self.scheduler.report(job_result).await;
                         }
                         Err(e) => {
                             // job panicked. report and exit process. The program is crippled, from
                             // now on, so we just exit, and hope to restart on a fresh state.
-                            tracing::error!("fatal error: bottomless job panicked: {e}");
+                            tracing::error!("fatal error: bottomles job panicked: {e}");
                             std::process::exit(1);
                         }
                     }
@@ -118,7 +119,10 @@ pub struct Bottomless<C> {
 }
 
 #[derive(thiserror::Error, Debug)]
-pub enum Error {}
+pub enum Error {
+    #[error("an error occured while storing a segment: {0}")]
+    Store(String),
+}
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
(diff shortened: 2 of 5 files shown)


bottomless2/src/lib.rs
```diff
@@ -65,12 +66,12 @@ impl<S: Storage + 'static> BottomlessLoop<S> {
...
                                 tracing::info!("processed job, {} jobs remaining", self.in_flight_futs.len());
                             }
-                            self.scheduler.report(job_result);
+                            self.scheduler.report(job_result).await;
                         }
                         Err(e) => {
                             // job panicked. report and exit process. The program is crippled, from
                             // now on, so we just exit, and hope to restart on a fresh state.
-                            tracing::error!("fatal error: bottomless job panicked: {e}");
+                            tracing::error!("fatal error: bottomles job panicked: {e}");
```
MarinPostma, line 74:
```suggestion
                            tracing::error!("fatal error: bottomless job panicked: {e}");
```

bottomless2/src/job.rs
```diff
@@ -1,14 +1,23 @@
...
+
 use crate::storage::Storage;
 use crate::Result;
-use crate::{NamespaceName, StoreSegmentRequest};
+use crate::StoreSegmentRequest;
+
+/// A request, with an id
+#[derive(Debug)]
+pub(crate) struct IndexedRequest<C> {
+    pub(crate) request: StoreSegmentRequest<C>,
```
LucioFranco, line 10: Why are we pub crating everything here? imo this is an anti-pattern

bottomless2/src/lib.rs
```diff
@@ -164,8 +168,10 @@ impl<C> Bottomless<C> {
 }
 
 // TODO: comes from libsql-server, when everything comes together
-pub struct NamespaceName;
+#[derive(Hash, PartialEq, Eq, Clone, Debug)]
+pub struct NamespaceName(Arc<str>);
```
LucioFranco, line 172: should this just store a Bytes?

bottomless2/src/scheduler.rs
```diff
@@ -13,44 +30,342 @@ use crate::{
...
+        let id = self.next_request_id;
+        self.next_request_id += 1;
+        let name = request.namespace.clone();
+        let slot = IndexedRequest { request, id };
+        let requests = self.requests.entry(name.clone()).or_default();
+        requests.requests.push_back(slot);
+
+        tracing::debug!(job_id = id, "job registered");
+
+        // if there is a priority for this namespace already, it must be higher than ours, because
```
LucioFranco, line 74: Can you expand on this comment, I think this logic is a bit confusing and would be good to have a clarifying comment.

bottomless2/src/scheduler.rs
```diff
@@ -13,44 +30,342 @@ use crate::{
...
+        }
+
+        assert!(
+            self.queue.get_priority(&name).is_none(),
+            "there should be not enqueued jobs at this point"
+        );
+
+        if !requests.requests.is_empty() {
+            let first_id = requests.requests.front().unwrap().id;
+            self.queue.push(name, Reverse(first_id));
```
LucioFranco, line 144: feels a little odd to me that report also can enqueue the next job?

bottomless2/src/scheduler.rs
```diff
@@ -13,44 +30,342 @@ use crate::{
...
     pub fn has_work(&self) -> bool {
-        todo!()
+        !self.queue.is_empty()
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use std::path::PathBuf;
+
```
LucioFranco, line 165:
```suggestion
```

bottomless2/src/scheduler.rs
```diff
@@ -13,44 +30,342 @@ use crate::{
...
+    use chrono::Utc;
+
```
LucioFranco, line 167:
```suggestion
```

bottomless2/src/scheduler.rs
```diff
@@ -13,44 +30,342 @@ use crate::{
...
+    use crate::Error;
+
```
LucioFranco, line 169:
```suggestion
```

LucioFranco (approved):
nice work 👍 left some questions can all be addressed later.

bottomless2/src/job.rs
MarinPostma: I don't think it is, it's a primitive struct, has no behavior, no private state, no field validation and no stability guarantee, in other words, it's just a container. I would even argue that making those fields private is the anti-pattern here, as you'd make building the type more difficult, the code harder to refactor (adding a field means adding methods, and changing the constructor), opt-out of pattern matching, and overall make the code more verbose, for no benefit.

bottomless2/src/scheduler.rs
MarinPostma: it doesn't enqueue the next job, it just make that namespace available for scheduling again. The next job will be popped from he queue on the next call to `schedule`

bottomless2/src/scheduler.rs
MarinPostma: this is on purpose, I separate std imports, from extern, for intern. It's a common style convention (so much so that RA, and rustfmt does it automatically: https://github.com/rust-lang/rustfmt/issues/2979)

bottomless2/src/scheduler.rs
MarinPostma: it's quite intuitive really: if there's already some work in the queue for that namespace, then it must have been registered before, and if it was registered before, then it must be processed before, so we don't update the priority for that namespace by inserting a more recent job, because that would decrease the priority of the namespace

bottomless2/src/scheduler.rs
MarinPostma: the idea is that you can't schedule work for a namespace that already has in-flight work. So when you schedule a namespace, you pop it from the queue, take some work from it to create a job and return the job. As long as the job is not reported, no more work for that namespace can be scheduled. When the job is reported, we put it back in the queue so that more work can be performed on it, whenever it's that namespace turn to make progress.

bottomless2/src/scheduler.rs
```diff
@@ -13,44 +30,342 @@ use crate::{
...
-pub(crate) struct Scheduler<S> {
+pub(crate) struct Scheduler<S: Storage> {
     capacity: usize,
     /// notify new durability index for namespace
-    durable_notifier: mpsc::Sender<(libsql_wal::name::NamespaceName, u64)>,
-    storage: S,
+    durable_notifier: mpsc::Sender<(NamespaceName, u64)>,
+    requests: HashMap<NamespaceName, NamespaceRequests<S::Config>>,
+    queue: priority_queue::PriorityQueue<NamespaceName, Reverse<u64>>,
+    storage: Arc<S>,
```
MarinPostma, line 39: I'll move storage out of the scheduler, it doesn't make sense to have it here.

PR #1402, opened 2024-05-20
Vector search support

penberg:
This pull request adds initial support for vector search in libSQL.

### Highlights

* Vector column type for storing vectors in tables.
* Vector index creation that is automatically updated on table updates.
* Exact vector search with metadata filtering using plain SQL.
* Approximate vector search using the new vector_top_k() function that is backed by DiskANN-based vector index.

### Usage

Creating a table with a vector column:

```sql
CREATE TABLE movies (
  title TEXT, 
  year INT, 
  embedding FLOAT32(3)
);
```

Inserting vector data:

```sql
INSERT INTO movies (title, year, embedding) 
VALUES 
  (
    'Napoleon', 
    2023, 
    vector('[1,2,3]')
  ), 
  (
    'Black Hawk Down', 
    2001, 
    vector('[10,11,12]')
  ), 
  (
    'Gladiator', 
    2000, 
    vector('[7,8,9]')
  ), 
  (
    'Blade Runner', 
    1982, 
    vector('[4,5,6]')
  );
```

Creating an index on vector column:

```sql
CREATE INDEX movies_idx USING vector_cosine_ops ON movies (embedding);
```

Finding top-k similar rows (exact):

```sql
SELECT title, year FROM movies ORDER BY vector_distance_cos(embedding, '[3,1,2]') LIMIT 3;
```

Finding top-k similar rows (approximate):

```sql
SELECT 
  title, 
  year 
FROM 
  vector_top_k('movies_idx', '[4,5,6]', 3) 
JOIN
  movies 
ON 
  movies.rowid = id;
```

### Limitations

* Index key is always `rowid`, primary keys not supported.
* `CREATE INDEX` does not index rows that already exist in the base table.
* Vector index uses 32-bit per vector element, which causes redundant I/O and space amplification.


libsql-sqlite3/src/vectorInt.h
```diff
@@ -0,0 +1,62 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Objects */
+typedef struct Vector Vector;
+typedef struct DiskAnnIndex DiskAnnIndex;
+
+#define MAX_VECTOR_SZ 16000
```
alex-sherwin, line 14: Was the max hard-coded dimensions of 16,000 picked for a particular reason?  Other vector DBs support larger numbers of dimensions (qdrant is 65536, milvus defaults to 32768 and can be increased via configuration)

There are valid use cases for large dimensional vectors, so if there is no specific limitation that the 16,000 is for... could this be made larger?  I would set it to 65536 (uint16 max size) to be in better alignment with the large dimensional vectors that other vector DBs support

libsql-sqlite3/src/vectordiskann.c
```diff
@@ -0,0 +1,865 @@
...
+    "WHERE type = ? AND name = ?";
+  rc = sqlite3_prepare_v2(db, zInsertSql, -1, &pStmt, 0);
+  if( rc!=SQLITE_OK ){
+    printf("Failed to prepare statement\n");
+    return rc;
+  }
+  sqlite3_bind_text(pStmt, 1, "diskann", -1, SQLITE_STATIC);
+  sqlite3_bind_text(pStmt, 2, zIdxName, -1, SQLITE_STATIC);
+  if( sqlite3_step(pStmt)!=SQLITE_ROW ){
+    printf("Failed to step\n");
```
xiankai, line 834: My insert query fails at this part when I have an existing unique index on my table.

I tried to dig deeper but only got as far as realizing that this SQL is executed on my unique index as well, resulting in an error because the SQL returned an `SQLITE_DONE` due to no results.

Minimal reproduction steps if it helps:
```
CREATE TABLE movies (title TEXT UNIQUE, embedding FLOAT32(3));
CREATE INDEX movies_idx USING diskann_cosine_ops ON movies (embedding);
INSERT INTO movies (title, embedding) VALUES ('a', vector('[1,2,3]'));
```

Hoping for a fix so that I can use it with my table with unique indexes, but cheers for the vector search support!

pax-k:
Any status for this PR? We're hardly waiting for vector support 🙏🏻 Thanks for the work btw!

libsql-sqlite3/src/vectordiskann.c
penberg: @xiankai https://github.com/tursodatabase/libsql/pull/1402/commits/acc21621c96c56dcb31548b3b23fd595bfa4ea1d should fix that issue.

penberg:
@pax-k I am actively working with folks to iron out some bugs and then get this merged.

libsql-sqlite3/src/vectorInt.h
penberg: Fixed in https://github.com/tursodatabase/libsql/pull/1402/commits/469e127ec1145a040d06b19e5a6c8876024f2406, thanks @alex-sherwin!

libsql-sqlite3/src/build.c
```diff
@@ -5563,6 +5571,7 @@ KeyInfo *sqlite3KeyInfoOfIndex(Parse *pParse, Index *pIdx){
   }
   if( pKey ){
     assert( sqlite3KeyInfoIsWriteable(pKey) );
+    pKey->zIndexName = pIdx->zName;
```
haaawk, line 5574: Does any code in vector search depend on this or is this just an improvement to make zIndexName not null whenever possible?

haaawk: NVM I see that this is a new field.

libsql-sqlite3/src/vector.c
```diff
@@ -0,0 +1,581 @@
...
+  Column *pCol;
+  Table *pTab;
+  char *zSql;
+  int nDims;
+  int rc;
+  int i;
+
+  assert( pUsing!= 0);
+
+  for( i=0; i<pUsing->nId; i++ ){
```
haaawk, line 376: Is it ever allowed for vector index to have pUsing != 1?

libsql-sqlite3/src/vector.c
```diff
@@ -0,0 +1,581 @@
...
+    if( sqlite3_stricmp(pUsing->a[i].zName, "diskann_cosine_ops")==0 ){
+      nDistanceOps = VECTOR_DISTANCE_COS;
+      break;
+    }
+    sqlite3ErrorMsg(pParse, "Unknown indexing method: %s", pUsing->a[i].zName);
+    return -1;
+  }
+  zSql = sqlite3MPrintf(db, "CREATE TABLE IF NOT EXISTS %s_shadow (index_key INT, data BLOB)", pIdx->zName);
+  rc = sqlite3_exec(db, zSql, 0, 0, 0);
```
haaawk, line 385: I'm probably still not understanding something but it seems that `return -1` in line 393 and 399 will leave shadow table present despite index creation failing. Is this a problem? Probably nothing bad happens despite some junk being kept around. Creation of the index with the same name will just work because of `IF NOT EXISTIS` part in the SQL.

libsql-sqlite3/src/vector.c
```diff
@@ -0,0 +1,581 @@
...
+**/
+static int parseVectorDims(const char *zType){
+  int dims = 0;
+  if( sqlite3_strnicmp(zType, "FLOAT32(", 8)==0 || sqlite3_strnicmp(zType, "FLOAT64(", 8)==0){
+    const char *z = zType + 8;
+    while( *z && *z!=')' ){
+      if( !sqlite3Isdigit(*z) ){
+        return -1;
+      }
+      dims = dims*10 + (*z - '0');
```
haaawk, line 352: Is it possible to define FLOAT32(X) with X being bigger than MAX_INT? Would it make sense to check for overflow here and report user readable error explaining that they try to use too big number of dimensions in case of X > MAX_INT or even X > MAX_VECTOR_DIMS?

penberg:
This work has been merged as part of the following PRs:

https://github.com/tursodatabase/libsql/pull/1531

https://github.com/tursodatabase/libsql/pull/1551

https://github.com/tursodatabase/libsql/pull/1557

https://github.com/tursodatabase/libsql/pull/1560

Therefore, closing this.

PR #1406, opened 2024-05-22
fix namespace bin header

MarinPostma:
fix bug introduced by #1179


haaawk (approved):

MarinPostma:
After more consideration, the platform is in the wrong, closing.

MarinPostma:
After re-re inspecting, the header should, in fact, be base64 encoded, reopening

LucioFranco (approved):

PR #1423, opened 2024-05-27
React Native support

ospfranco:
Adds the necessary scripts to build the C bindings for iOS and Android which the in turn only need to be dropped into iOS/Android projects to be used via Obj-C++ or Android's JNI.

For my specific use case I use them in op-sqlite.

### DIFF
--- a/bindings/c/Cargo.toml
+++ b/bindings/c/Cargo.toml
@@ -24,5 +24,10 @@ libsql = { path = "../../libsql", features = ["encryption"] }
 libsql = { path = "../../libsql"}
 
 
-[profile.release]
-strip = "symbols"
+# The produced binaries are too large for mobiles
+# When compiling for iOS or Android, you should turn on symbol stripping, lto and cut debug symbols
+# [profile.release]
+# debug = false # Exclude debug symbols
+# strip = "symbols" # Exclude the rest of the symbols
+# # opt-level = "z" # Did not use this, but it equals C++'s optimize for size (O3?)
+# lto = true # Link time optimization, not sure what this does but it helps reduce the size
--- a/bindings/c/Cargo.toml
+++ b/bindings/c/Cargo.toml
@@ -13,6 +13,16 @@ cbindgen = "0.24.0"
 [dependencies]
 bytes = "1.5.0"
 lazy_static = "1.4.0"
-libsql = { path = "../../libsql", features = ["encryption"] }
 tokio = { version = "1.29.1", features = [ "rt-multi-thread" ] }
 hyper-rustls = { version = "0.25", features = ["webpki-roots"]}
+
+[target.'cfg(not(any(target_os = "ios", target_os = "android")))'.dependencies]
+libsql = { path = "../../libsql", features = ["encryption"] }
+
+# Disable encryption for ios and android targets
+[target.'cfg(any(target_os = "ios", target_os = "android"))'.dependencies]
+libsql = { path = "../../libsql"}
+
+
+[profile.release]
+strip = "symbols"
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,4 @@
 [toolchain]
 profile = "default"
 channel = "1.78.0"
+targets = ["x86_64-apple-ios", "aarch64-apple-ios", "aarch64-apple-ios-sim", "aarch64-linux-android", "armv7-linux-androideabi", "x86_64-linux-android", "i686-linux-android"]
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,7 @@ libsql-sqlite3/**.o.tmp
 /libsql-ffi/bundled/SQLite3MultipleCiphers/build/**
 # will be copied from bundled/src
 /libsql-ffi/bundled/SQLite3MultipleCiphers/src/sqlite3.c
+
+/bindings/c/generated
+/bindings/c/**.xcframework
+/bindings/**/.DS_Store
--- a/bindings/c/Makefile
+++ b/bindings/c/Makefile
@@ -1,6 +1,11 @@
 OS := $(shell uname)
 CFLAGS := -Iinclude
 LDFLAGS := -lm
+ARCHS_IOS = x86_64-apple-ios aarch64-apple-ios aarch64-apple-ios-sim
+ARCHS_ANDROID = aarch64-linux-android armv7-linux-androideabi x86_64-linux-android i686-linux-android
+LIB = libsql_experimental.a
+HEADER = libsql.h
+XCFRAMEWORK = libsql.xcframework
 
 # Set LIBSQL_PATH to the default path if not provided
 LIBSQL_EXPERIMENTAL_PATH ?= ../../target/release/libsql_experimental.a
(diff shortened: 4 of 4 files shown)


libsql-ffi/build.rs
```diff
@@ -124,127 +124,6 @@ pub fn build_bundled(out_dir: &str, out_path: &Path) {
         .flag("-D_POSIX_THREAD_SAFE_FUNCTIONS") // cross compile with MinGW
         .warnings(false);
 
-    if cfg!(feature = "wasmtime-bindings") {
```
ospfranco, line 127: This was a mistake. Let me know if you want this PR merged and I will re-add them.

libsql-ffi/build.rs
```diff
@@ -124,127 +124,6 @@ pub fn build_bundled(out_dir: &str, out_path: &Path) {
...
-        cfg.flag("-DSQLITE_HAS_CODEC").flag("-DSQLITE_TEMP_STORE=2");
-
-        let target = env::var("TARGET").unwrap();
-        let host = env::var("HOST").unwrap();
-
-        let is_windows = host.contains("windows") && target.contains("windows");
-        let is_apple = host.contains("apple") && target.contains("apple");
-
-        let lib_dir = env("OPENSSL_LIB_DIR").map(PathBuf::from);
-        let inc_dir = env("OPENSSL_INCLUDE_DIR").map(PathBuf::from);
```
ospfranco, line 141: They will however require changes to compile for android as I was having errors on Cmake when trying to compile against sqlcipher.

haaawk:
Some CI checks are failing @ospfranco

bindings/c/Cargo.toml
```diff
@@ -13,6 +13,9 @@ cbindgen = "0.24.0"
 [dependencies]
 bytes = "1.5.0"
 lazy_static = "1.4.0"
-libsql = { path = "../../libsql", features = ["encryption"] }
```
haaawk, line 16: This is not an acceptable change. Most of our projects that use C bindings depend on the encryption feature.

bindings/c/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.0.0"
 edition = "2021"
 
 [lib]
-crate-type = ["cdylib", "staticlib"]
```
haaawk, line 7: Removing cdylib for all targets is probably not acceptable.

bindings/c/Cargo.toml
```diff
@@ -13,6 +13,9 @@ cbindgen = "0.24.0"
 [dependencies]
 bytes = "1.5.0"
 lazy_static = "1.4.0"
-libsql = { path = "../../libsql", features = ["encryption"] }
+libsql = { path = "../../libsql"}
 tokio = { version = "1.29.1", features = [ "rt-multi-thread" ] }
 hyper-rustls = { version = "0.25", features = ["webpki-roots"]}
+
+[profile.release]
+strip = "symbols"
```
haaawk, line 21: This is something we need to talk about. Removing symbols makes debugging really hard. @penberg @MarinPostma @LucioFranco do you have any opinion about this?

libsql-ffi/build.rs
haaawk: We do want to merge the PR so please re-add them

rust-toolchain.toml
```diff
@@ -1,3 +1,4 @@
 [toolchain]
 profile = "default"
 channel = "1.78.0"
+targets = ["x86_64-apple-ios", "aarch64-apple-ios", "aarch64-apple-ios-sim", "aarch64-linux-android", "armv7-linux-androideabi", "x86_64-linux-android", "i686-linux-android"]
```
haaawk, line 4: Don't we need a desktop targets for linux and macos as well?

ospfranco:
Thanks for taking a look, however, I'm currently swamped with work and next week I'm going on vacation for several months. Feel free to take over this PR and modify it as you see fit:

- The changes on build.rs are reversible
- Encryption seems to be compiling sqlcipher (?) via cmake, which also includes OpenSSL flags, the problem the compilation step is hard to integrate with Android (and that's why I had to disable it). Someone else that knows the internals of libsql should give it a try.
- The rest seem trivial

After all is done and said, you only need to follow the guide I left you on notion to produce a static lib and run that against the opsqlite sample app. Once the emulator starts and everything is green you are ready to go.

haaawk:
> Thanks for taking a look, however, I'm currently swamped with work and next week I'm going on vacation for several months. Feel free to take over this PR and modify it as you see fit:

I'm not sure I will have capacity to take this PR over. Could you please finish it to the state that's mergable. Right now multiple CI checks are failing for it so it is impossible to accept such work and integrate it into the repository.

> 
> * The changes on build.rs are reversible

What does that mean? They are not needed? The changes that are required by Android and iOS need to be merged and if they are not compatible with other targets then they should be guarded with android or ios feature.


> * Encryption seems to be compiling sqlcipher (?) via cmake, which also includes OpenSSL flags, the problem the compilation step is hard to integrate with Android (and that's why I had to disable it). Someone else that knows the internals of libsql should give it a try. 

It's ok to disable the encryption for Android and maybe iOS for now but it should be done only for that targets not for all targets which breaks them.

> * The rest seem trivial
> 
> After all is done and said, you only need to follow the guide I left you on notion to produce a static lib and run that against the opsqlite sample app. Once the emulator starts and everything is green you are ready to go.

Could you paste a link to that guide please? I haven't seen it yet.

> 
> P.S. if you do turn on encryption, it will require a modification to the bindings to send the encryption key from JS, so, open a PR when the time comes and I will take a look.

rust-toolchain.toml
ospfranco: If you want, it makes it easier to set up cross compilation on new machines as you don't have to run the `rustup install target ...` commands but this is entirely optional

bindings/c/Cargo.toml
ospfranco: Tried this to reduce the size of the binary, Rust binaries are way to large for mobile targets. Currently with symbol stripping the binary sits at 60mbs, which is way too much.

Recently wrote more about this:

https://ospfranco.com/rust-reduce-binary-size/

I can remove it, but don't have the time to see how to apply stripping and lto optimization for conditional targets, did not find any documentation on how to do it.

ospfranco:
Leaving the guide to submitting updates to op-sqlite here for future reference:

https://discord.com/channels/@me/1243157733574840351/1255132819668209724

bindings/c/Cargo.toml
haaawk: Then maybe comment it out with a note that it should be uncommented for android/ios builds to reduce the size of the binary

haaawk (approved):

haaawk (approved):

bindings/c/Cargo.toml
MarinPostma: this is biting us in production already, I'm going to remove it. If you need stripped binaries, I think it's easy enough to call `strip` on your binary?

rust-toolchain.toml
MarinPostma: This force the installation of too many toolchains that are useless most of the time, and it just broke my installation by creating conflicts.

bindings/c/Cargo.toml
ospfranco: It's already removed, I commented it out

rust-toolchain.toml
ospfranco: Feel free to remove it

haaawk: How did it create conflicts? It seems pretty harmless except for using some disk space, no?

PR #1426, opened 2024-05-29
Add storage server proto definition

avinassh:
This patch adds the RPC methods required for the storage server. The methods are based on the Virtual WAL interface


libsql-storage/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+description = "libSQL WAL"
+repository = "https://github.com/tursodatabase/libsql"
+license = "MIT"
+
+[dependencies]
+prost = "0.12"
+tonic = { version = "0.10.0", features = ["tls"] }
+
+[dev-dependencies]
+prost-build = "0.12.0"
```
LucioFranco, line 14: nit: should just drop the trailing `.0` since that is the format we have elsewhere in this

libsql-storage/proto/storage.proto
```diff
@@ -0,0 +1,78 @@
+syntax = "proto3";
```
LucioFranco, line 1: nit: I prefer to go with `MessageRequest` and `MessageResponse` this seems to be the style google uses. I find resp and res to be really easy to confuse.

LucioFranco (approved):
left some small nits, overall this looks fine. I don't have enough details about each individual RPC to understand if we need those or if they are done right but should be easy to change those in the future so not gonna block on that.

avinassh:
@LucioFranco fixed the nits. Thank you 👍

PR #1431, opened 2024-05-31
Add storage server

avinassh:
This patch adds the storage gRPC server. It is based on the proto that was previously merged in https://github.com/tursodatabase/libsql/pull/1426

This adds a `FrameStore` trait which lets us implement different pluggable storage backends. Check the simple in-memory storage `InMemFrameStore` for a sample implementation

### DIFF
--- a/libsql-storage-server/src/service.rs
+++ b/libsql-storage-server/src/service.rs
@@ -1,27 +1,26 @@
 use std::sync::atomic::AtomicU32;
-use std::sync::Arc;
 
 use crate::memory_store::InMemFrameStore;
 use crate::store::FrameStore;
 use libsql_storage::rpc;
 use libsql_storage::rpc::storage_server::Storage;
-use tokio::sync::Mutex;
 use tonic::{Request, Response, Status};
 use tracing::{error, trace};
 
 pub struct Service {
-    store: Arc<Mutex<dyn FrameStore + Send + Sync>>,
+    store: Box<dyn FrameStore + Send + Sync>,
     db_size: AtomicU32,
 }
 
 impl Service {
     pub fn new() -> Self {
         Self {
-            store: Arc::new(Mutex::new(InMemFrameStore::new())),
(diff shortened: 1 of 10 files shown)


libsql-storage-server/src/service.rs
```diff
@@ -0,0 +1,153 @@
...
+use crate::memory_store::InMemFrameStore;
+use crate::store::FrameStore;
+use libsql_storage::rpc;
+use libsql_storage::rpc::storage_server::Storage;
+use tokio::sync::Mutex;
+use tonic::{Request, Response, Status};
+use tracing::{error, trace};
+
+pub struct Service {
+    store: Arc<Mutex<dyn FrameStore + Send + Sync>>,
```
LucioFranco, line 13: dont need to arc tonic  will do that for you. It also feels like this locking should happen at the store layer rather than here. Clients that go over the network in general don't need a mutex.

libsql-storage-server/src/store.rs
```diff
@@ -0,0 +1,19 @@
+use async_trait::async_trait;
+use bytes::Bytes;
+
+#[async_trait]
+pub trait FrameStore: Send + Sync {
+    async fn insert_frame(&mut self, namespace: &str, page_no: u64, frame: bytes::Bytes) -> u64;
+    async fn insert_frames(&mut self, namespace: &str, frames: Vec<FrameData>) -> u64;
```
LucioFranco, line 7: imo these should all take &self and handle locking internally.

Could we combine these two trait methods together?

LucioFranco (approved):

libsql-storage-server/src/store.rs
avinassh: > imo these should all take &self and handle locking internally.

right. I was planning to remove that in the future PRs, but I did it now. 

> Could we combine these two trait methods together?

I need them both for now. In the future versions, I will remove `insert_frame`.

PR #1435, opened 2024-06-03
Add experimental Redis and Foundation DB backends

avinassh:


libsql-storage-server/src/fdb_store.rs
```diff
@@ -0,0 +1,153 @@
...
+use foundationdb::Transaction;
+use tracing::error;
+
+pub struct FDBFrameStore {
+    _network: NetworkAutoStop,
+}
+
+impl FDBFrameStore {
+    pub fn new() -> Self {
+        let _network = unsafe { foundationdb::boot() };
```
avinassh, line 16: this is how the docs recommend too

libsql-storage-server/src/fdb_store.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        txn.set(&frame_data_key.as_bytes(), &frame.data);
+        txn.set(&frame_page_key.as_bytes(), &pack(&frame.page_no));
+        txn.set(&page_key.as_bytes(), &pack(&frame_no));
+    }
+}
+
+#[async_trait]
+impl FrameStore for FDBFrameStore {
+    async fn insert_frame(&self, namespace: &str, page_no: u32, frame: bytes::Bytes) -> u64 {
```
MarinPostma, line 55: why do we need both `insert_frame` and `insert_frames`?

MarinPostma (commented):
I think it would be nice to start working on the transaction model, because right now, you only support mutually exclusive readers/writer, and there a good amount of things that need to be set up for the wal to work properly under normal conditions

libsql-storage-server/src/fdb_store.rs
avinassh: Right, both are not needed. I will remove `insert_frame` in a follow up PR.

PR #1437, opened 2024-06-04
fix shared schema replicatioon

MarinPostma:
Allow shared schema database to be used with replication. There were two issues:

- the master schema database was not considered a primary, this is solved by exposing it's logger
- when we replicated a shared schema db without it's master first, a constraint was violated in the meta store. This is solved by forcing replication of the master schema first.


athoscouto (approved):
looks good, but remember remove the dbg macros

LucioFranco (approved):

PR #1442, opened 2024-06-05
libsql-wal; replication: Stream frames from current segment

MarinPostma:
Stream frames from the current segment. Frames from other segments will be handled in subsequent PRs.


libsql-wal/src/segment/current.rs
```diff
@@ -261,7 +265,35 @@ impl<F> CurrentSegment<F> {
...
+
+    // todo: maybe return boxed frames?
+    pub fn rev_frame_stream(&self) -> impl Stream<Item = Result<Frame>> + '_
+    where
+        F: FileExt,
+    {
+        async_stream::try_stream! {
+            let (start_frame_no, last_committed) = {
+                let header = self.header.lock();
+                dbg!((header.start_frame_no.get(), header.last_commited_frame_no.get()))
```
LucioFranco, line 277: dbg

LucioFranco: should we just convert these dbg into tracing::debug?

LucioFranco (approved):
lgtm lots of dbg that we could just convert to tracing debugs

libsql-wal/src/segment/current.rs
MarinPostma: will do as followup 👍

PR #1448, opened 2024-06-06
libsql wal optim frame reuse

MarinPostma:
This optim resuses frames for pages already written during a transaction. This can save a significant amount of disk for very large txns.


athoscouto (approved):
how does this affect replication?

MarinPostma:
@athoscouto it doesn't, this is internal to libsql-wal

MarinPostma:
besides, we don't replicate uncommitted transactions

PR #1460, opened 2024-06-10
Integrate durable WAL into libsql server

avinassh:
This patch adds the durable WAL from `libsql-storage` module to sqld. 

This has one (minor) breaking change. The CLI arg `use-libsql-wal` changed to `use-custom-wal` and it takes two options `libsql-wal` or `durable-wal`

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -130,10 +130,6 @@ jobs:
       env:
         LIBSQL_EXPERIMENTAL_PAGER: 1000
       run: cargo nextest run
-
-    - name: Run tests with durable WAL
-      run: cargo nextest run --features durable-wal
-
     - name: clean build dir
       run: rm -rf libsql-ffi/bundled/SQLite3MultipleCiphers/build
     - name: embedded replica encryption tests
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -130,6 +130,10 @@ jobs:
       env:
         LIBSQL_EXPERIMENTAL_PAGER: 1000
       run: cargo nextest run
+
+    - name: Run tests with durable WAL
+      run: cargo nextest run --features durable-wal
+
     - name: clean build dir
       run: rm -rf libsql-ffi/bundled/SQLite3MultipleCiphers/build
     - name: embedded replica encryption tests
--- a/libsql-server/Cargo.toml
+++ b/libsql-server/Cargo.toml
@@ -38,6 +38,7 @@ jsonwebtoken = "9"
 libsql = { path = "../libsql/", optional = true }
 libsql_replication = { path = "../libsql-replication" }
 libsql-wal = { path = "../libsql-wal/" }
+libsql-storage = { path = "../libsql-storage" }
 metrics = "0.21.1"
 metrics-util = "0.15"
 metrics-exporter-prometheus = "0.12.2"
--- a/libsql-server/Cargo.toml
+++ b/libsql-server/Cargo.toml
@@ -38,7 +38,7 @@ jsonwebtoken = "9"
 libsql = { path = "../libsql/", optional = true }
 libsql_replication = { path = "../libsql-replication" }
 libsql-wal = { path = "../libsql-wal/" }
-libsql-storage = { path = "../libsql-storage" }
+libsql-storage = { path = "../libsql-storage", optional = true }
 metrics = "0.21.1"
 metrics-util = "0.15"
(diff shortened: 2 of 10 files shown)


libsql-server/src/lib.rs
```diff
@@ -632,62 +641,74 @@ where
         Arc<dyn Fn() -> InnerWalManager + Sync + Send + 'static>,
```
avinassh, line 641:
```suggestion
        Arc<dyn Fn(ns: Option<NamespaceName>) -> InnerWalManager + Sync + Send + 'static>,
```

I have a question for the follow up PR. I need to change the definition of the `make_wal_manager` to pass Namespace to the `DurableWALManager`. I am wondering what would be the right place to do it?

libsql-server/Cargo.toml
```diff
@@ -38,6 +38,7 @@ jsonwebtoken = "9"
 libsql = { path = "../libsql/", optional = true }
 libsql_replication = { path = "../libsql-replication" }
 libsql-wal = { path = "../libsql-wal/" }
+libsql-storage = { path = "../libsql-storage" }
```
MarinPostma, line 41: can you make it a feature for now?

libsql-server/src/lib.rs
MarinPostma: that's not going to work, unfortunately, you can assume the namespace name at this level, but only when the db is opened. Consider what happens when you run `ATTACH otherdb`.

MarinPostma: IOW, you only know what db is opened when `WalManager::open` is called

libsql-server/src/lib.rs
```diff
@@ -632,62 +641,74 @@ where
...
-        let wal_path = self.path.join("wals");
+        if self.use_custom_wal.is_none() {
+            tracing::info!("using sqlite3 wal");
+            return Ok((
+                Arc::new(|| Either3::A(Sqlite3WalManager::default())),
+                Box::pin(ready(Ok(()))),
+            ));
+        };
+
+        let use_custom_wal = self.use_custom_wal.unwrap();
```
MarinPostma, line 652: turn that into a match, it's more idiomatic

libsql-server/src/lib.rs
MarinPostma: yeah, see how I did it below. I'm not a fan of doing it this way, but I haven't come up with a better alternative either.

libsql-server/Cargo.toml
avinassh: yes, I have made it now.

avinassh: I need `libsql-server-devel` docker images with `durable-wal` enabled, so I have updated the github workflow: `server-pr-images`. But unfortunately it can be tested after merging only.

libsql-server/src/lib.rs
avinassh: yes, I have refactored this method now

.github/workflows/rust.yml
```diff
@@ -130,6 +130,10 @@ jobs:
       env:
         LIBSQL_EXPERIMENTAL_PAGER: 1000
       run: cargo nextest run
+
+    - name: Run tests with durable WAL
```
MarinPostma, line 134: Is that necessary, since even with the feature on, tests only run with vanilla wal, if I understand correctly?

MarinPostma: if so, no need to lengthen the CI for nothing

libsql-server/src/connection/connection_manager.rs
```diff
@@ -6,7 +6,12 @@ use std::time::{Duration, Instant};
 use crossbeam::deque::Steal;
 use crossbeam::sync::{Parker, Unparker};
 use hashbrown::HashMap;
+#[cfg(feature = "durable-wal")]
+use libsql_storage::{DurableWal, DurableWalManager};
+#[cfg(not(feature = "durable-wal"))]
 use libsql_sys::wal::either::Either;
+#[cfg(feature = "durable-wal")]
+use libsql_sys::wal::either::Either3;
```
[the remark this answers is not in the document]
avinassh: done! Removed most `cfg` usages.

.github/workflows/rust.yml
avinassh: I wanted to use durable wal in the runner, but I forgot to remove it. I have removed it now

MarinPostma (approved):

PR #1461, opened 2024-06-10
fix fs storage impl

MarinPostma:


libsql-wal/src/bottomless/storage/fs.rs
```diff
@@ -48,9 +45,9 @@ impl<I: Io> Storage for FsStorage<I> {
 
         let path = self.prefix.join("segments").join(key);
 
-        let buf = Vec::with_capacity(segment_data.len().unwrap() as usize);
+        let buf = Vec::with_capacity(dbg!(segment_data.len().unwrap()) as usize);
```
LucioFranco, line 48: I will clean these up

PR #1463, opened 2024-06-10
Implement FromValue trait for byte arrays.

ignatz:
(small nit on the side and I feel bad for commenting on formatting, but could you check in a `.rustfmt.toml` that matches your style? I accidentally ran fmt on the repo by reflex... :hide: but I would also feel more confident in the consistency of my formatting if I could leave it to the robots)

### DIFF
--- a/libsql/src/errors.rs
+++ b/libsql/src/errors.rs
@@ -52,6 +52,8 @@ pub enum Error {
     InvalidTlsConfiguration(std::io::Error),
     #[error("Transactional batch error: {0}")]
     TransactionalBatchError(String),
+    #[error("Invalid blob size, expected {0}")]
+    InvalidBlobSize(usize),
 }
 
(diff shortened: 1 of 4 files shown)


MarinPostma:
hey, we use stock rustfmt, and the CI shouldn't pass without rustfmt, so go ahead :)

could you please add a small unit test somewhere, please? otherwise lgtm :)

Thanks!

ignatz:
> hey, we use stock rustfmt, and the CI shouldn't pass without rustfmt, so go ahead :)

Apologies and glad to hear I was wrong. I was suffering from a case of nested workspace syndrome :hide:

> could you please add a small unit test somewhere, please? otherwise lgtm :)

:+1: I didn't find a fitting place, so I added a new one. Happy to move it around. 

Thanks

MarinPostma:
thanks!

PR #1464, opened 2024-06-10
libsql-wal frame injector

MarinPostma:
The Injector allows frames to be injected into a replica's wal.

A flag is added to the current segment to determine whether frames are inserted in order or not. Primaries always write frames ordered in the current segment, but replicas may receive frames out-of-order.


libsql-wal/src/replication/injector.rs
```diff
@@ -0,0 +1,175 @@
...
+            max_tx_frame_no: 0,
+        })
+    }
+
+    pub async fn insert_frame(&mut self, frame: Box<Frame>) -> Result<()> {
+        let size_after = frame.size_after();
+        self.max_tx_frame_no = self.max_tx_frame_no.max(frame.header().frame_no());
+        self.buffer.push(frame);
+
+        if size_after.is_some() || self.capacity == self.buffer.len() {
```
LucioFranco, line 42: should this be `>=`? for the capacity check?

libsql-wal/src/replication/injector.rs
```diff
@@ -0,0 +1,175 @@
...
+    async fn flush(&mut self, size_after: Option<u32>) -> Result<()> {
+        let buffer = std::mem::take(&mut self.buffer);
+        let current = self.wal.current.load();
+        let commit_data = size_after.map(|size| (size, self.max_tx_frame_no));
+        if commit_data.is_some() {
+            self.max_tx_frame_no = 0;
+        }
+        let buffer = current
+            .insert_frames(buffer, commit_data, &mut self.tx)
+            .await?;
```
LucioFranco, line 58: do we run into any consistency issues if this await either gets cancelled or fails since we took the buffer it would get dropped here and lost.

libsql-wal/src/segment/current.rs
```diff
@@ -146,30 +221,29 @@ impl<F> CurrentSegment<F> {
...
-                if tx.not_empty() {
-                    let last_frame_no = tx.next_frame_no - 1;
-                    let mut header = { *self.header.lock() };
-                    header.last_commited_frame_no = last_frame_no.into();
-                    header.db_size = size_after.into();
-                    header.recompute_checksum();
+        if let Some(size_after) = size_after {
+            if tx.not_empty() {
+                let last_frame_no = tx.next_frame_no - 1;
+                let mut header = { *self.header.lock() };
```
LucioFranco, line 229: why these {}?

LucioFranco (approved):

libsql-wal/src/segment/current.rs
MarinPostma: Cuz I'm scared of those temporary variables now lol

libsql-wal/src/replication/injector.rs
MarinPostma: That would be safer yeah. But we're injecting frames one by one, so we will hit the ==

libsql-wal/src/replication/injector.rs
MarinPostma: The caller should rollback and start replicating from the latest committed in that case, not implemented yet

PR #1465, opened 2024-06-11
Use the namespace resolver to extract namespace

avinassh:
Storage server requires namespace UUID to operate. In the `WALManager` we do not have access to the namespace, so this patch extracts the namespace from the `db_path`

The large diff is due to I moved `NamespaceName` from `libsql-wal` to `libsql-sys` so that `libsql-storage` also can use it. Or else I could copy the resolver method https://github.com/tursodatabase/libsql/blob/7ebe45b52/libsql-server/src/lib.rs#L672-L684 inside `libsql-storage` and use that.


MarinPostma (commented):
quick comment: do not assume that NamespaceName is an Uuid
it can be anything.

MarinPostma (approved):

PR #1470, opened 2024-06-11
Add a link to my (experimental) D driver based on the C driver.

pdenapo:
I have created a D driver for libsql (It uses the C driver but replacing the C header for a D version). 

A nice feature is that some tests are included using the Silly test runner for D.


avinassh:
thank you @pdenapo! 🥳

PR #1471, opened 2024-06-11
fix current segment replicator

MarinPostma:
The previous implementation was reading 1 frame too many. The new implementation selects frames from the segment index to read the minimum amount of frames to bring the replica up to date.

The new segment index is probably too slow, I'll optimize the datastructure in a followup.

### DIFF
--- a/libsql-wal/src/segment/current.rs
+++ b/libsql-wal/src/segment/current.rs
@@ -1,4 +1,3 @@
-use std::collections::BTreeMap;
 use std::io::{BufWriter, IoSlice, Write};
 use std::num::NonZeroU64;
 use std::ops::DerefMut;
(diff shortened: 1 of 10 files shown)


libsql-wal/src/segment/current.rs
```diff
@@ -350,29 +350,36 @@ impl<F> CurrentSegment<F> {
...
-                let byte_offset = frame_offset(next_offset);
+        let (last_committed, size_after) = {
+            let header = self.header.lock();
+            (header.last_committed(), header.db_size.get())
+        };
 
+        // we get all frame offsets of pages that changed since start_frame_no
+        let mut iter = self.index.iter(start_frame_no, last_committed).peekable();
+
+        async_stream::try_stream! {
```
LucioFranco, line 369: consider using tokio's consume budget to make sure that we yield in a friendly manner?

LucioFranco (approved):

libsql-wal/src/segment/current.rs
MarinPostma: Can you explain why we would need that?

PR #1472, opened 2024-06-12
implemented custom headers passing in auth layer

shopifyski:
every auth strategy now advertises the headers it needs by implementing required_fields method
the higher layer is responsible for providing the fields in UserAuthContext
UserAuthContext is not passed in as a result type, it is always required
if the higher layer is not able to provide required_fields, it should assume the particular auth strategy will not work
if required fields are not provided, auth strategy should not attempt to fall back unless there is a clear algorithm to do so

### DIFF
--- a/libsql-server/src/auth/user_auth_strategies/http_basic.rs
+++ b/libsql-server/src/auth/user_auth_strategies/http_basic.rs
@@ -1,4 +1,7 @@
-use crate::auth::{AuthError, Authenticated};
+use crate::auth::{
+    constants::{AUTH_HEADER, GRPC_AUTH_HEADER},
+    AuthError, Authenticated,
+};
 
 use super::{UserAuthContext, UserAuthStrategy};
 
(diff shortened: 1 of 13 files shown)


libsql-server/src/auth/user_auth_strategies/http_basic.rs
```diff
@@ -7,27 +7,31 @@ pub struct HttpBasic {
...
 impl UserAuthStrategy for HttpBasic {
-    fn authenticate(
-        &self,
-        context: Result<UserAuthContext, AuthError>,
-    ) -> Result<Authenticated, AuthError> {
+    fn authenticate(&self, ctx: UserAuthContext) -> Result<Authenticated, AuthError> {
         tracing::trace!("executing http basic auth");
+        let auth_str = ctx
+            .custom_fields
+            .get("authorization")
```
LucioFranco, line 14: should we use the constant from the `http` crate here instead of hard coding this value? If the hashmap needs all lower case we could wrap the hashmap to always call `to_lower_case` to ensurse that we always hash the correct key.

We use this hard coded value in a lot of places it would be nice if we could dry that up. I feel somewhat uneasy about having a dynamic hashmap that has hard coded values in diff spots all over the code base, I think we should make this a bit more type safe since this code is pretty critical and called an almost every request.

LucioFranco (changes requested):

libsql-server/src/auth/user_auth_strategies/http_basic.rs
```diff
@@ -7,27 +10,31 @@ pub struct HttpBasic {
...
+        let creds_match = token.contains(expected_value);
         if creds_match {
             return Ok(Authenticated::FullAccess);
         }
-
         Err(AuthError::BasicRejected)
     }
+
+    fn required_fields(&self) -> Vec<&'static str> {
+        vec![AUTH_HEADER, GRPC_AUTH_HEADER]
```
LucioFranco, line 36: why does this one require both of these headers but the jwt doesn't require both?

libsql-server/src/auth/user_auth_strategies/mod.rs
```diff
@@ -3,58 +3,39 @@ pub mod http_basic;
...
 pub use jwt::Jwt;
 
-use super::{AuthError, Authenticated};
+use super::{constants::AUTH_HEADER, AuthError, Authenticated};
 
 #[derive(Debug)]
 pub struct UserAuthContext {
-    scheme: Option<String>,
-    token: Option<String>,
+    pub custom_fields: HashMap<&'static str, String>,
```
LucioFranco, line 14: can we remove this pub and encode the operations as functions on the struct?

libsql-server/src/auth/user_auth_strategies/mod.rs
```diff
@@ -64,11 +45,16 @@ impl UserAuthContext {
...
+        self.custom_fields.insert(key, value.into());
+    }
 }
 
 pub trait UserAuthStrategy: Sync + Send {
-    fn authenticate(
-        &self,
-        context: Result<UserAuthContext, AuthError>,
-    ) -> Result<Authenticated, AuthError>;
+    fn required_fields(&self) -> Vec<&'static str> {
```
LucioFranco, line 55: can we add some docs tho these fn so that we can encode their requirements

LucioFranco (changes requested):

LucioFranco (approved):

PR #1478, opened 2024-06-16
Make libsql-ffi's build more friendly to sandboxed environments.

levydsa:
Fixes <https://github.com/ipetkov/crane/issues/644>. I wonder why it copies the bindgen.rs file so many times to the same location.

Also, two symlinks were causing trouble when copying the source to the nix store, they are currently pointing to nothing.  So I just copied them from somewhere else in the project.

### DIFF
--- a/libsql-ffi/build.rs
+++ b/libsql-ffi/build.rs
@@ -38,7 +38,14 @@ fn main() {
     };
 
     let dir = env!("CARGO_MANIFEST_DIR");
-    std::fs::copy(format!("{dir}/{bindgen_rs_path}"), &out_path).unwrap();
+
+    Command::new("cp")
+        .arg("--no-preserve=mode,ownership")
+        .arg("-R")
+        .arg(format!("{dir}/{bindgen_rs_path}"))
+        .arg(&out_path)
+        .output()
+        .unwrap();
 
     println!("cargo:lib_dir={out_dir}");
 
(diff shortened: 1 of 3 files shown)


libsql-ffi/build.rs
```diff
@@ -38,7 +38,14 @@ fn main() {
...
     let dir = env!("CARGO_MANIFEST_DIR");
-    std::fs::copy(format!("{dir}/{bindgen_rs_path}"), &out_path).unwrap();
+
+    Command::new("cp")
+        .arg("--no-preserve=mode,ownership")
+        .arg("-R")
+        .arg(format!("{dir}/{bindgen_rs_path}"))
+        .arg(&out_path)
+        .output()
+        .unwrap();
```
penberg, line 48: I am fairly sure this won't work for Windows builds. @LucioFranco I see we have some Windows checks as part of the CI so I wonder why they didn't catch this?

LucioFranco: windows just checks a very small feature subset, as we don't really support windows for the full libsql c code iirc.

LucioFranco (approved):
After talking on discord this makes sense to me 👍

penberg:
@LucioFranco @levydsa We build on Windows all the time... For some reason, this does _not_ seem to break Windows build, but now I am curious to understand why that is. Clearly we can't just run a UNIX command like that from `build.rs` and be portable.

levydsa:
In this same build file there is a call to `make` and shell script `configure`. I think they are available via mingw, but I'm not sure. I'll try to find a more cross platform solution.

PR #1486, opened 2024-06-19
use mimalloc for sqlite

MarinPostma:
Sqlite and sqld use a different allocator, which can cause memory fragmentation. This PR provides a custom allocator for sqlite that uses rust's global allocator (mimalloc).

### DIFF
--- a/libsql-server/src/lib.rs
+++ b/libsql-server/src/lib.rs
@@ -1,5 +1,8 @@
 #![allow(clippy::type_complexity, clippy::too_many_arguments)]
 
+use std::alloc::Layout;
+use std::ffi::c_void;
+use std::mem::{align_of, size_of};
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
 use std::str::FromStr;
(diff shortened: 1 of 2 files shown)


LucioFranco (approved):

libsql-server/src/lib.rs
```diff
@@ -726,3 +735,68 @@ where
...
+/// the size of the allocation is stored as a usize before the returned pointer. A i32 would be
+/// sufficient, but we need the returned pointer to be aligned to 8
+fn setup_sqlite_alloc() {
+    use std::alloc::GlobalAlloc;
+
+    unsafe extern "C" fn malloc(size: i32) -> *mut c_void {
+        let size_total = size as usize + size_of::<usize>();
+        let layout = Layout::from_size_align(size_total, align_of::<usize>()).unwrap();
+        let ptr = GLOBAL.alloc(layout);
+        *(ptr as *mut usize) = size as usize;
```
sivukhin, line 749: Allocator can fail and return null - in this case `SEGFAULT` will happen here (although I hope that SQLite can handle out of memory conditions in some cases). Should `malloc` be implemented more carefully then?

PR #1497, opened 2024-06-24
flag sqlite mimalloc allocator

MarinPostma:
since fragmentation is not the cause of the leak, put the sqlite mimalloc allocator behind a flag for now

### DIFF
--- a/libsql-server/src/lib.rs
+++ b/libsql-server/src/lib.rs
@@ -391,7 +391,9 @@ where
         static INIT: std::sync::Once = std::sync::Once::new();
         let mut join_set = JoinSet::new();
 
-        setup_sqlite_alloc();
+        if std::env::var("LIBSQL_SQLITE_MIMALLOC").is_ok() {
+            setup_sqlite_alloc();
+        }
 
         INIT.call_once(|| {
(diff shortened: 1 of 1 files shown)


sivukhin:
@MarinPostma, JFYI, it seems like current implementation of allocator can cause SEGFAULT (see https://github.com/tursodatabase/libsql/pull/1486#discussion_r1647736447)

Should we fix it?

MarinPostma:
Yes! Do you have a reproducer? I kinda found weird stuff this weekend but didn't have the time to investigate, so I thought I'd just disable it for now

MarinPostma:
I just read the comment, I'll fix it now, but not as bad as I thought. It's almost impossible to get a system allocator return null

sivukhin:
No, I just read the code for now - don't have any example with real SEGFAULT.

I just thought that fix is easy (we just need to correctly handle nullptr output from the allocator) - and it's good to do that now given that it's unclear when `mimalloc` will be enabled again in `libsql` (and by this time all of us will forget about nuances of implementation).

But if you are busy you can merge this PR and I will create fix for `mimalloc` bit later.

PR #1502, opened 2024-06-25
Move in-memory cache to disk

avinassh:
Previously, `libsql-storage` stored all the data in-memory. It kept a in-memory frame cache and also cached all the transient transaction writes in-memory. This patch changes it to store the data on local disk. This data is transient in nature, disk loss should not cause any issues.

### DIFF
--- a/libsql-storage/src/lib.rs
+++ b/libsql-storage/src/lib.rs
@@ -1,14 +1,17 @@
-use std::collections::BTreeMap;
+mod local_cache;
+
 use std::ffi::OsStr;
 use std::os::unix::ffi::OsStrExt;
+use std::path::Path;
 use std::sync::{Arc, Mutex};
 
+use crate::local_cache::LocalCache;
+use crate::rpc::Frame;
 use libsql_sys::ffi::{SQLITE_ABORT, SQLITE_BUSY};
 use libsql_sys::name::{NamespaceName, NamespaceResolver};
 use libsql_sys::rusqlite;
 use libsql_sys::wal::{Result, Vfs, Wal, WalManager};
 use rpc::storage_client::StorageClient;
-use sieve_cache::SieveCache;
 use tonic::transport::Channel;
 use tracing::{error, trace};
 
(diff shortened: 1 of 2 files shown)


libsql-storage/src/lib.rs
```diff
@@ -74,13 +77,24 @@ impl WalManager for DurableWalManager {
     ) -> Result<Self::Wal> {
         let db_path = OsStr::from_bytes(&db_path.to_bytes());
         let namespace = self.resolver.resolve(db_path.as_ref());
+        let cache_path = {
+            let parent_dir = Path::new(db_path).parent().unwrap_or(Path::new("."));
+            &parent_dir.join(format!("{}_local_cache.db", namespace))
+        };
```
avinassh, line 83: currently, I am creating the cache file in the same location where db file exists.

libsql-storage/src/local_cache.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    pub fn get_all_pages(&self, txn_id: &str) -> Result<Vec<(u32, Vec<u8>)>> {
+        let mut stmt = self
+            .conn
+            .prepare("SELECT page_no, data FROM transactions WHERE txn_id = ?1")?;
+        let pages: Result<Vec<(u32, Vec<u8>)>> = stmt
+            .query_map(params![txn_id], |row| Ok((row.get(0)?, row.get(1)?)))?
+            .collect();
+        self.conn.execute(
```
sivukhin, line 89: Silly question: it's impossible to have `insert_page` happening between `SELECT` and `DELETE` statements in `get_all_pages` implementation, right?

sivukhin: BTW, is it okay to remove frames from local cache before we will actually successfully send them?

avinassh: > Silly question: it's impossible to have `insert_page` happening between `SELECT` and `DELETE` statements in `get_all_pages` implementation, right? 

not a silly question at all! And yes, it is not possible to have new pages of the transaction to be inserted while deletion is happening. 

> BTW, is it okay to remove frames from local cache before we will actually successfully send them?

Nope, not okay. This assumes the network call does not fail. There is also another issue that storage server has enough memory to receive all the frames. Ideally, we should be sending in batches and add retry mechanism. 

The code is of alpha quality and it is not in production. There are other issues like using `unwrap` everywhere and not handling errors properly. In the future iterations, I will fix these issues.

MarinPostma (approved):

PR #1506, opened 2024-06-25
build sqld debug

MarinPostma:
build an image with debug symbols, and make PR images ship with debug symbols too

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,10 +18,15 @@ RUN cat rust-toolchain.toml | grep "channel" | awk '{print $3}' | sed 's/\"//g' 
     && cargo install cargo-chef
 
 FROM chef AS planner
+ARG BUILD_DEBUG=false
+ENV CARGO_PROFILE_RELEASE_DEBUG=$BUILD_DEBUG
+RUN echo $CARGO_PROFILE_RELEASE_DEBUG
 COPY . .
 RUN cargo chef prepare --recipe-path recipe.json
 
 FROM chef AS builder
+ARG BUILD_DEBUG=false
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -18,10 +18,15 @@ RUN cat rust-toolchain.toml | grep "channel" | awk '{print $3}' | sed 's/\"//g'
     && cargo install cargo-chef
 
 FROM chef AS planner
+ARG BUILD_DEBUG=false
+ENV CARGO_PROFILE_RELEASE_DEBUG=$build_debug
```
sivukhin, line 22: Why lowercase `$build_debug` and not `$BUILD_DEBUG`?

Dockerfile
```diff
@@ -18,10 +18,15 @@ RUN cat rust-toolchain.toml | grep "channel" | awk '{print $3}' | sed 's/\"//g'
     && cargo install cargo-chef
 
 FROM chef AS planner
+ARG BUILD_DEBUG=false
```
sivukhin, line 21: Why not to just proxy `CARGO_PROFILE_RELEASE_DEBUG` env var as-is? So we can set it as a build arg explicitly: `CARGO_PROFILE_RELEASE_DEBUG=true`?

Dockerfile
MarinPostma: overlook, thanks for catching that!

Dockerfile
MarinPostma: I don't want to tie it to much to how it's currently built. Ultimately I'l like to either build with split debug info, and discard them when not in debug for local builds, and build the image from the official build for official release, so that we can use the debug info in production as well

PR #1507, opened 2024-06-25
Use `max_frame_no` for transaction reads and writes

avinassh:
- Storage server now uses `max_frame_no` from the txn to read frames
  It will make sure to not read any frames beyond the `max_frame_no`
  of the transaction. This fixes the bug in the isolation level.
- Storage server now checks `max_frame_no` before inserting. If it
  has changed meanwhile, then it rejects the insertion request.

This lets us have multiple (non concurrent) writers or even multiple
logical primary instances.

Minor changes:
-  Improved the way we handle keys in Foundation DB
- Added basic tonic error handling

### DIFF
--- a/libsql-storage-server/src/fdb_store.rs
+++ b/libsql-storage-server/src/fdb_store.rs
@@ -89,12 +89,18 @@ impl FDBFrameStore {
 
 #[async_trait]
 impl FrameStore for FDBFrameStore {
-    async fn insert_frames(&self, namespace: &str, _max_frame_no: u64, frames: Vec<Frame>) -> u64 {
-        let max_frame_key = format!("{}/max_frame_no", namespace);
+    async fn insert_frames(
+        &self,
+        namespace: &str,
+        max_frame_no: u64,
+        frames: Vec<Frame>,
+    ) -> Result<u64, Error> {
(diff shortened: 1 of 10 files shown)


libsql-storage-server/src/fdb_store.rs
```diff
@@ -40,77 +75,82 @@ impl FDBFrameStore {
...
+        max_frame_no: u64,
+        frames: Vec<Frame>,
+    ) -> Result<u64, Error> {
         let db = foundationdb::Database::default().unwrap();
         let txn = db.create_trx().expect("unable to create transaction");
         let mut frame_no = self.get_max_frame_no(&txn, namespace).await;
-
+        if frame_no != max_frame_no {
+            return Err(WriteConflict);
+        }
```
MarinPostma, line 103: how does fdb actually handles conflict? we could very well have to writer submitting a transaction with the same max_frame_no at the same time, no?

avinassh: Yes, but only one of them will succeed. 

FDB provides SSI - Strict Serialisable Isolation. If a transaction reads/writes some value, then another transaction cannot change it in between.

MarinPostma (approved):

PR #1510, opened 2024-06-27
libsql: Add Statement::run() API

penberg:
The `execute()` method returns an error if the query returns rows, which makes it unsuitable for running any type of SQL queries. Similarly, the `query()` method only works on SQL statements that return rows. Therefore, the `run()` method is provided to execute any type of SQL statement.

This is needed, for example, for the libSQL JavaScript library that needs to preserve `better-sqlite3` semantics.

### DIFF
--- a/libsql/src/local/impls.rs
+++ b/libsql/src/local/impls.rs
@@ -105,6 +105,13 @@ impl Stmt for LibsqlStmt {
             .map(|r| Rows { inner: Box::new(r) })
     }
 
+    async fn run(&mut self, params: &Params) -> Result<()> {
+        let params = params.clone();
+        let stmt = self.0.clone();
+
+        stmt.run(&params)
+    }
+
     fn reset(&mut self) {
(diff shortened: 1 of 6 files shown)


libsql/src/local/impls.rs
```diff
@@ -105,6 +105,13 @@ impl Stmt for LibsqlStmt {
             .map(|r| Rows { inner: Box::new(r) })
     }
 
+    async fn run(&mut self, params: &Params) -> Result<()> {
+        let params = params.clone();
```
haaawk, line 109: Why do we need to clone the params?

haaawk (approved):

libsql/src/local/impls.rs
penberg: @haaawk I don't think it is needed, but as both `execute()` and `query()` do it, I didn't want to break the symmetry... 🤔

PR #1511, opened 2024-06-27
libsql: Add Connection::total_changes() API

penberg:
It's a rusqlite API that we're missing.

### DIFF
--- a/libsql/src/hrana/connection.rs
+++ b/libsql/src/hrana/connection.rs
@@ -54,6 +54,10 @@ where
         self.current_stream().affected_row_count()
     }
 
+    pub fn total_changes(&self) -> u64 {
+        self.current_stream().total_changes()
+    }
+
     pub fn last_insert_rowid(&self) -> i64 {
         self.current_stream().last_insert_rowid()
     }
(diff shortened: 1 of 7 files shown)


libsql/src/hrana/connection.rs
```diff
@@ -54,6 +54,10 @@ where
         self.current_stream().affected_row_count()
     }
 
+    pub fn total_changes(&self) -> u64 {
+        self.current_stream().total_changes()
```
haaawk, line 58: What happens when the stream changes on connection?
Hrana allows multiple streams on a single connection - how does that fit with this?

haaawk: ok I see that `total changes` comes from server and is stream independent, correct?

haaawk: Ok it actually does not come from the server. Shouldn't we accummulate the counter in connection not in stream?

haaawk:
LGTM in general but I think for HRANA we should store the counter in connection not in stream.

libsql/src/hrana/connection.rs
penberg: You are right, but it's a bit involved to fix it. I opened https://github.com/tursodatabase/libsql/issues/1512 for tracking this, but merging this as intermediate solution.

PR #1513, opened 2024-06-27
fix panic on delete namespace

MarinPostma:
fix panic on delete namespace

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,3 @@
 [toolchain]
 profile = "default"
 channel = "1.78.0"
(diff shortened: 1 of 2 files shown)


rust-toolchain.toml
```diff
@@ -1,4 +1,3 @@
 [toolchain]
 profile = "default"
 channel = "1.78.0"
-targets = ["x86_64-apple-ios", "aarch64-apple-ios", "aarch64-apple-ios-sim", "aarch64-linux-android", "armv7-linux-androideabi", "x86_64-linux-android", "i686-linux-android"]
```
MarinPostma, line 4: This is breaking local installations

PR #1523, opened 2024-07-01
Add Dockerfile and Github action for Storage Server

avinassh:
Dockerfile is inspired from https://github.com/tursodatabase/libsql/blob/main/Dockerfile and I have tested the workflow in my fork.

PR #1524, opened 2024-07-01
Update workflow to release a `latest` dev image

avinassh:
forgot this in #1523 

this patch updates the workflow to generate the images with `latest` tag

### DIFF
--- a/.github/workflows/storage-server-images.yml
+++ b/.github/workflows/storage-server-images.yml
@@ -51,7 +51,9 @@ jobs:
         with:
           context: .
           push: true
-          tags: ${{ steps.meta.outputs.tags }}-${{ steps.get-short-sha.outputs.sha }}
+          tags: |
+            ${{ steps.meta.outputs.tags }}-${{ steps.get-short-sha.outputs.sha }}
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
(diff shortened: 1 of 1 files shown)


.github/workflows/storage-server-images.yml
```diff
@@ -51,7 +51,9 @@ jobs:
         with:
```
sivukhin, line 51: NIT (line 12, `IMAGE_NAME`): maybe we can make this workflow similar to `publish-server` and just use `tursodatabase/libsql-storage-server-devel` without `github.repository` interpolation?

sivukhin (approved):

.github/workflows/storage-server-images.yml
avinassh: we could, but then I won't be able to run trigger this workflow in my fork. May be once it is production ready and set,  we can rename the `IMAGE_NAME`. WDYT?

sivukhin: Ah, I see. This is NIT - so we can skip this for sure

But why are you developing this in fork actually?)

avinassh: I use my fork as a testbed. Easier to test changes. For e.g. Github workflows don't run properly unless they are in `main`. So to test, I merge in my fork's main, test, modify etc. Another reason is just habit, to make changes in my fork and send PR from them.

PR #1525, opened 2024-07-01
fix overflow in row counters

MarinPostma:
The rows read/written counters are stores in a u32, that can overflow. This PR "fixes" that by saturating the counter. This is a temporary fix, since increasing the size of the counter is more work than we are willing to put in right now.

### DIFF
--- a/libsql-sqlite3/src/vdbe.c
+++ b/libsql-sqlite3/src/vdbe.c
@@ -815,8 +815,8 @@ static void inc_row_read(Vdbe *p, int count) {
 }
 
 static void inc_row_written(Vdbe *p, int count) {
-    u32 *read = &p->aLibsqlCounter[LIBSQL_STMTSTATUS_ROWS_WRITTEN - LIBSQL_STMTSTATUS_BASE];
-    *read = saturating_add(*read, count);
+    u32 *write = &p->aLibsqlCounter[LIBSQL_STMTSTATUS_ROWS_WRITTEN - LIBSQL_STMTSTATUS_BASE];
+    *write = saturating_add(*write, count);
 }
 
 /*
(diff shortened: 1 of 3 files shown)


athoscouto (approved):

avinassh (approved):

libsql-sqlite3/src/vdbe.c
```diff
@@ -3719,7 +3734,7 @@ case OP_Count: {         /* out2 */
     nEntry = 0;  /* Not needed.  Only used to silence a warning. */
     i64 nPages = 0;
     rc = sqlite3BtreeCount(db, pCrsr, &nEntry, &nPages);
```
sivukhin, line 3736: @MarinPostma, not a subject of this PR, but I spotted a weird behaviour of the `sqlite3BtreeCount` - we don't count read pages in case of non success return code (for example, if query were interrupted). Should we fix it or this is desired behaviour?

libsql-sqlite3/src/vdbe.c
```diff
@@ -804,6 +804,21 @@ FuncDef *try_instantiate_wasm_function(sqlite3 *db, const char *pName, int nName
...
+    return (u32)MIN(0xFFFFFFFF, sum);
+}
+
+static void inc_row_read(Vdbe *p, int count) {
+    u32 *read = &p->aLibsqlCounter[LIBSQL_STMTSTATUS_ROWS_READ - LIBSQL_STMTSTATUS_BASE];
+    *read = saturating_add(*read, count);
+}
+
+static void inc_row_written(Vdbe *p, int count) {
+    u32 *read = &p->aLibsqlCounter[LIBSQL_STMTSTATUS_ROWS_WRITTEN - LIBSQL_STMTSTATUS_BASE];
```
sivukhin, line 818: NIT: `u32 *written = ...`

sivukhin (approved):

libsql-sqlite3/src/vdbe.c
MarinPostma: I don't really know, afaik the counter is not necessarily very accurate, sometimes it counts pages, sometimes it's rows. I think we should consider it to be an estimate of how many rows were written/read. From that perspective I don't think we care too much about accounting when there are errors. If somebody complains we'll look I to it I suppose, but we already saturating the counter on overflow so I don't think it matters too much to us 😄

PR #1528, opened 2024-07-02
fix: prevent incorrect range queries when searching for a page version

avinassh:
It is important to verify that the data we got from Foundation DB matches with what we want, since we are doing a range query.

for example, say we searched for ('db_name42', 10, 20). If this page does not exist, then it could match with ('db_name41', 10, 20) or with ('db_name42', 9, 20) since it does lexicographic search.


MarinPostma (approved):
is there a way to prevent fetching a page we are not interested in?

PR #1531, opened 2024-07-04
feat(vector): add basic functions (vector/vector_extract etc)

sivukhin:
## Context

This PR adds native support for vector functions in the LibSQL.

This PR were extracted from the branch https://github.com/tursodatabase/libsql/pull/1402/ in order to make review more granular and simpler

- Vectors are represented as binary blobs with vector components encoded in little-endian format according to IEEE-754 standard
    * `LibSQL` now supports f32/f64 vectors of any dimension not more than 65536 (see `MAX_VECTOR_SZ`)
    * Zero-length vectors are supported - and they are always represented as zero size BLOB (**not** NULL)
- `LibSQL` can extended vector encoding on disk with 1 trailing byte which will encode vector type information (for now: `1 = F32, 2 = F64`).
```
[F3 76 86 C4] [BC 9D F1 3F] [01] - this is f32 vector because type equals to 0x01 = 1 (value = [-1075.72,1.88763])
[F3 76 86 C4] [BC 9D F1 3F]      - this is f32 vector because default type is 1 (value = [-1075.72,1.88763])
[F3 76 86 C4   BC 9D F1 3F] [02] - this is f64 vector because type equals to 0x02 = 2 (value = [1.10101])
```

## Changes

- 5 new builtin functions are added:
    * `vector`/`vector32`/`vector64` - convert `TEXT` or `BLOB` to binary vector `BLOB`. If text is provided - then `vector`/`vector32` will produce `F32` vector, but `vector64` - `F64` vector
    * `vector_extract` - convert binary vector `BLOB` to the human readable `TEXT` string (like `[1,2,3]`)
    * `vector_distance_cos` - calculates cosine distance (**not** similarity) between vectors of same dimension and same type
- `SQLITE_OMIT_VECTOR` preprocessor directive and `autoconf` parameter `--disable-vector` which will remove any signs of vector functions from the final build
- `libsql_vector.test` - TCL based test suite

### DIFF
--- a/libsql-sqlite3/ext/crr/Makefile
+++ b/libsql-sqlite3/ext/crr/Makefile
@@ -1,3 +1,5 @@
+LDLIBS += -lm
+
 ifndef CI_GCC
 CC:=gcc
 else
--- a/libsql-sqlite3/ext/crr/Makefile
+++ b/libsql-sqlite3/ext/crr/Makefile
@@ -1,3 +1,4 @@
+# we need math.h for sqrt function in LIBSQL vector feature
 LDLIBS += -lm
 
 ifndef CI_GCC
--- a/libsql-sqlite3/src/vector.c
+++ b/libsql-sqlite3/src/vector.c
@@ -542,14 +542,6 @@ out_free:
 }
 
 /*
- * Marker function which is used in index creation syntax: CREATE INDEX idx ON t(libsql_vector_idx(emb));
-*/
-static void libsqlVectorIdx(sqlite3_context *context, int argc, sqlite3_value **argv){ 
-  // it's important for this function to be no-op as sqlite will apply this function to the column before feeding it to the index
-  sqlite3_result_value(context, argv[0]);
-}
-
-/*
 ** Register vector functions.
 */
 void sqlite3RegisterVectorFunctions(void){
@@ -559,8 +551,6 @@ void sqlite3RegisterVectorFunctions(void){
     FUNCTION(vector64,            1, 0, 0, vector64Func),
     FUNCTION(vector_extract,      1, 0, 0, vectorExtractFunc),
     FUNCTION(vector_distance_cos, 2, 0, 0, vectorDistanceCosFunc),
-
-    FUNCTION(libsql_vector_idx,  -1, 0, 0, libsqlVectorIdx),
   };
   sqlite3InsertBuiltinFuncs(aVectorFuncs, ArraySize(aVectorFuncs));
 }
--- a/libsql-sqlite3/src/vector.c
+++ b/libsql-sqlite3/src/vector.c
@@ -1,0 +1,572 @@
+/*
+** 2024-07-04
+**
+** Copyright 2024 the libSQL authors
+**
+** Permission is hereby granted, free of charge, to any person obtaining a copy of
+** this software and associated documentation files (the "Software"), to deal in
+** the Software without restriction, including without limitation the rights to
+** use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+** the Software, and to permit persons to whom the Software is furnished to do so,
+** subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in all
+** copies or substantial portions of the Software.
+**
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+** FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+** COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+** IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+** CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+**
+******************************************************************************
+**
+** libSQL basic vector functions
+*/
+#ifndef SQLITE_OMIT_VECTOR
+
+#include "sqliteInt.h"
+#include "vectorInt.h"
+
+#define MAX_FLOAT_CHAR_SZ  1024
+
+/**************************************************************************
+** Utility routines for dealing with Vector objects
+**************************************************************************/
+
+size_t vectorDataSize(VectorType type, VectorDims dims){
+  switch( type ){
+    case VECTOR_TYPE_FLOAT32:
+      return dims * sizeof(float);
+    case VECTOR_TYPE_FLOAT64:
+      return dims * sizeof(double);
+    default:
+      assert(0);
+  }
+  return 0;
+}
+
+void vectorInit(Vector *pVector, VectorType type, VectorDims dims, void *data){
+  pVector->type = type;
+  pVector->dims = dims;
+  pVector->data = data;
+  pVector->flags = 0;
+}
+
+/*
+ * Allocate a Vector object and its data buffer
+*/
+Vector *vectorAlloc(VectorType type, VectorDims dims){
+  void *pVector = sqlite3_malloc(sizeof(Vector) + vectorDataSize(type, dims));
+  if( pVector==NULL ){
+    return NULL;
+  }
+  vectorInit(pVector, type, dims, ((char*) pVector) + sizeof(Vector));
+  return pVector;
+}
+
+/*
+ * Allocate a Vector object and its data buffer from the SQLite context. 
+*/
+static Vector* vectorContextAlloc(sqlite3_context *context, int type, int dims){
+  void *pVector = sqlite3_malloc64(sizeof(Vector) + vectorDataSize(type, dims));
+  if( pVector==NULL ){
+    sqlite3_result_error_nomem(context);
+    return NULL;
+  }
+  vectorInit(pVector, type, dims, ((char*) pVector) + sizeof(Vector));
+  return pVector;
+}
+
+/*
+ * Free a Vector object and its data buffer allocated, unless the vector is static.
+*/
+void vectorFree(Vector *pVector){
+  if( pVector==0 ){
+    return;
+  }
+  if( pVector->flags & VECTOR_FLAGS_STATIC ){
+    return;
+  }
+  sqlite3_free(pVector);
+}
+
+float vectorDistanceCos(const Vector *pVector1, const Vector *pVector2){
+  assert(v1->type == v2->type);
+  switch (pVector1->type) {
+    case VECTOR_TYPE_FLOAT32:
+      return vectorF32DistanceCos(pVector1, pVector2);
+    case VECTOR_TYPE_FLOAT64:
+      return vectorF64DistanceCos(pVector1, pVector2);
+    default:
+      assert(0);
+  }
+  return 0;
+}
+
+/*
+ * Parses vector from text representation (e.g. '[1,2,3]'); vector type must be set
+*/
+static int vectorParseSqliteText(
+  sqlite3_value *arg,
+  Vector *pVector,
+  char **pzErrMsg
+){
+  const unsigned char *pzText;
+  double elem;
+  float *elemsFloat;
+  double *elemsDouble;
+  int iElem = 0;
+  // one more extra character in order to safely print data from elBuf with printf-like method; will be set to zero later
+  char valueBuf[MAX_FLOAT_CHAR_SZ + 1];
+  int iBuf = 0;
+
+  assert( v->type == VECTOR_TYPE_FLOAT32 || v->type == VECTOR_TYPE_FLOAT64 );
+
+  if( pVector->type == VECTOR_TYPE_FLOAT32 ){
+    elemsFloat = pVector->data;
+  } else if( pVector->type == VECTOR_TYPE_FLOAT64 ){
+    elemsDouble = pVector->data;
+  }
+
+  if( sqlite3_value_type(arg) != SQLITE_TEXT ){
+    *pzErrMsg = sqlite3_mprintf("invalid vector: not a text type");
+    goto error;
+  }
+
+  pzText = sqlite3_value_text(arg);
+  if ( pzText == NULL ) return 0;
+
+  while( pzText && sqlite3Isspace(*pzText) )
+    pzText++;
+
+  if( *pzText != '[' ){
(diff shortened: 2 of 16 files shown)


libsql-sqlite3/ext/crr/Makefile
```diff
@@ -1,3 +1,5 @@
+LDLIBS += -lm
```
sivukhin, line 1: I'm not sure why we need this...
@penberg - do you remember why you changed this Makefile?

haaawk:
CI is failing:
```
found uncommited changes in the bundle; please run 'cargo xtask build-bundled' and commit the result
```

libsql-sqlite3/src/vector.c
```diff
@@ -0,0 +1,558 @@
...
+  }
+  vectorInit(pVector, type, dims, ((char*) pVector) + sizeof(Vector));
+  return pVector;
+}
+
+/*
+ * Free a Vector object and its data buffer allocated, unless the vector is static.
+*/
+void vectorFree(Vector *pVector){
+  if( pVector==0 ){
```
haaawk, line 86: Why does one function use NULL and another 0? Is there a reason for this inconsistency?

sivukhin: No reason - I guess we better to explicitly use NULL everywhere. Will fix that

libsql-sqlite3/src/vectorfloat32.c
```diff
@@ -0,0 +1,238 @@
...
+
+  assert( pVector->type == VECTOR_TYPE_FLOAT32 );
+  assert( pVector->dims <= MAX_VECTOR_SZ );
+  assert( nBlobSize >= pVector->dims * sizeof(float) );
+
+  for(i = 0; i < pVector->dims; i++){
+    pPtr += serializeF32(pPtr, elems[i]);
+    len += sizeof(float);
+  }
+  return len;
```
haaawk, line 95: Nit: You don't need len variable at all. Just return pVector->dims * sizeof(float)

haaawk: It will be more performant I believe.

libsql-sqlite3/src/vectorfloat32.c
```diff
@@ -0,0 +1,238 @@
...
+  const unsigned char *pBlob,
+  size_t nBlobSize
+){
+  float *elems = pVector->data;
+  unsigned i;
+  pVector->type = VECTOR_TYPE_FLOAT32;
+  pVector->dims = nBlobSize / sizeof(float);
+
+  assert( pVector->dims <= MAX_VECTOR_SZ );
+  assert( nBlobSize >= vectorDataSize(pVector->type, pVector->dims) );
```
haaawk, line 109: Isn't this redundant when 3 lines above we set pVector->dims to nBlobSize / sizeof(float)?

libsql-sqlite3/src/vectorfloat32.c
```diff
@@ -0,0 +1,238 @@
...
+  float *elems = pVector->data;
+  size_t nBufSize;
+  size_t iBuf = 0;
+  char *pText;
+  char valueBuf[SINGLE_FLOAT_CHAR_LIMIT];
+
+  assert( pVector->type == VECTOR_TYPE_FLOAT32 );
+  assert( pVector->dims <= MAX_VECTOR_SZ );
+
+  nBufSize = 2 + pVector->dims * (SINGLE_FLOAT_CHAR_LIMIT + 1 /* plus comma */);
```
haaawk, line 160: Do we put comma after last element as well?

haaawk: It seems not.

haaawk: but I guess this extra byte is used for \0 - is that right?

libsql-sqlite3/src/vectorfloat32.c
```diff
@@ -0,0 +1,238 @@
...
+  assert( pVector->type == VECTOR_TYPE_FLOAT32 );
+  assert( 0 <= pVector->dims && pVector->dims <= MAX_VECTOR_SZ );
+
+  if( sqlite3_value_type(arg) != SQLITE_BLOB ){
+    *pzErr = sqlite3_mprintf("invalid f32 vector: not a blob type");
+    goto error;
+  }
+
+  pBlob = sqlite3_value_blob(arg);
+  if( sqlite3_value_bytes(arg) < 4 * pVector->dims ){
```
haaawk, line 224: Nit: sizeof(float) * pVector->dims?

libsql-sqlite3/src/vectorfloat32.c
sivukhin: No, we don't put. I think here we allocate 1 more extra unnecessary byte (it is really unnecessary since `sqlite3_result_text` ask for blob length explicitly). But I guess it's fine.

libsql-sqlite3/src/vectorfloat64.c
```diff
@@ -0,0 +1,240 @@
...
+
+  assert( pVector->type == VECTOR_TYPE_FLOAT64 );
+  assert( pVector->dims <= MAX_VECTOR_SZ );
+  assert( nBlobSize >= pVector->dims * sizeof(double) );
+
+  for (i = 0; i < pVector->dims; i++) {
+    pPtr += serializeF64(pPtr, elems[i]);
+    len += sizeof(double);
+  }
+  return len;
```
haaawk, line 100: Drop len and just return pVector->dims * sizeof(double)?

libsql-sqlite3/src/vectorfloat64.c
```diff
@@ -0,0 +1,240 @@
...
+  const unsigned char *pBlob,
+  size_t nBlobSize
+){
+  double *elems = pVector->data;
+  unsigned i;
+  pVector->type = VECTOR_TYPE_FLOAT64;
+  pVector->dims = nBlobSize / sizeof(double);
+
+  assert( pVector->dims <= MAX_VECTOR_SZ );
+  assert( nBlobSize >= vectorDataSize(pVector->type, pVector->dims) );
```
haaawk, line 114: Seems redundant because pVector->dims is set to nBlobSize / sizeof(double) 3 lines above.

libsql-sqlite3/src/vectorfloat64.c
```diff
@@ -0,0 +1,240 @@
...
+  assert( pVector->type == VECTOR_TYPE_FLOAT64 );
+  assert( pVector->dims <= MAX_VECTOR_SZ );
+
+  // allocate one extra trailing byte with vector blob type metadata
+  nBlobSize = vectorDataSize(pVector->type, pVector->dims) + 1;
+  pBlob = sqlite3_malloc64(nBlobSize);
+
+  if( pBlob ){
+    vectorF64SerializeToBlob(pVector, pBlob, nBlobSize - 1);
+    pBlob[nBlobSize - 1] = VECTOR_TYPE_FLOAT64;
```
haaawk, line 140: Why do we add this in vectorF64Serialize but we don't do this in vectorF32Serialize?

libsql-sqlite3/src/vectorfloat64.c
```diff
@@ -0,0 +1,240 @@
...
+  assert( pVector->type == VECTOR_TYPE_FLOAT64 );
+  assert( 0 <= pVector->dims && pVector->dims <= MAX_VECTOR_SZ );
+
+  if( sqlite3_value_type(arg) != SQLITE_BLOB ){
+    *pzErr = sqlite3_mprintf("invalid f64 vector: not a blob type");
+    goto error;
+  }
+
+  pBlob = sqlite3_value_blob(arg);
+  if( sqlite3_value_bytes(arg) < 8 * pVector->dims ){
```
haaawk, line 226: Nit: sizeof(double) * pVector->dims?

libsql-sqlite3/src/vectorfloat64.c
haaawk: We don't seem to use it in vectorF64DeserializeFromBlob

sivukhin: We can assert last byte in `vectorF64DeserializeFromBlob`, yes, but the resolution of type happening in `vector.c` and `vectorfloatxx.c` source code just dealing with what it got.

In `vectorF32Serialize` I decided to not to fill trailing byte as even sized blobs are f32 by default and this way we will produce more "natural to human" blobs for F32 vectors.

libsql-sqlite3/src/vector.c
```diff
@@ -0,0 +1,558 @@
...
+      elemsDouble[iElem++] = elem;
+    }
+    if( this == ']' ){
+      break;
+    }
+  }
+  while( pzText && sqlite3Isspace(*pzText) )
+    pzText++;
+
+  if( pzText != NULL && *pzText != ']' ){
```
haaawk, line 193: Shouldn't *pzText be always equals to ']' here? Why is it ok for pzText to be NULL?

sivukhin: Yes, we checked that `pzText` is not null early in this function - all this checks can be omitted. Will fix that.

libsql-sqlite3/src/vector.c
```diff
@@ -0,0 +1,558 @@
...
+
+  if( sqlite3_value_type(arg) != SQLITE_TEXT ){
+    *pzErrMsg = sqlite3_mprintf("invalid vector: not a text type");
+    goto error;
+  }
+
+  pzText = sqlite3_value_text(arg);
+  if ( pzText == NULL ) return 0;
+
+  while( pzText && sqlite3Isspace(*pzText) )
```
haaawk, line 141: I have some brain problems and can't see why pzText can every be false here.

libsql-sqlite3/src/vector.c
```diff
@@ -0,0 +1,558 @@
...
+    if( pVector->type == VECTOR_TYPE_FLOAT32 ){
+      elemsFloat[iElem++] = elem;
+    } else if( pVector->type == VECTOR_TYPE_FLOAT64 ){
+      elemsDouble[iElem++] = elem;
+    }
+    if( this == ']' ){
+      break;
+    }
+  }
+  while( pzText && sqlite3Isspace(*pzText) )
```
haaawk, line 190: Some here - why advancing a pointer can lead to this pointer being NULL?

libsql-sqlite3/src/vector.c
sivukhin: This is not your problem - this is code issue! Will fix this!

libsql-sqlite3/src/vectorfloat64.c
haaawk: Ok I see it's used in  https://github.com/tursodatabase/libsql/pull/1531/files#diff-98fd1b40c774432f5288bd6ae7000632bbffb8687f15734977cb8cc794ed1a41R237

libsql-sqlite3/src/vector.c
```diff
@@ -0,0 +1,558 @@
...
+  nBlobSize = sqlite3_value_bytes(arg);
+  if( nBlobSize % 2 != 0 ){ 
+    // we have trailing byte with explicit type definition
+    *pType = pBlob[nBlobSize - 1];
+  } else { 
+    // else, fallback to FLOAT32
+    *pType = VECTOR_TYPE_FLOAT32;
+  }
+  if( *pType == VECTOR_TYPE_FLOAT32 ){
+    *pDims = nBlobSize / 4;
```
haaawk, line 244: nit: / sizeof(float)

libsql-sqlite3/src/vector.c
```diff
@@ -0,0 +1,558 @@
...
+  } else if( *pType == VECTOR_TYPE_FLOAT64 ){
+    *pDims = nBlobSize / 8;
```
haaawk, line 246: nit: / sizeof(double)

libsql-sqlite3/src/vector.c
```diff
@@ -0,0 +1,558 @@
...
+    return -1;
+  }
+  *pDims = 0;
+  for(iText = 0; iText < textBytes; iText++){
+    if( text[iText] == ',' ){
+      *pDims += 1;
+    }
+    if( sqlite3Isdigit(text[iText]) ){
+      textHasDigit = 1;
+    }
```
haaawk, line 284: I'm not sure but shouldn't this be:
```
    if( text[iText] == ',' ){
      *pDims += 1;
      continue;
    }
    if( sqlite3Isdigit(text[iText]) ){
      textHasDigit = 1;
    }
```

libsql-sqlite3/src/vector.c
```diff
@@ -0,0 +1,558 @@
...
+  }
+  if( textHasDigit ){
+    *pDims += 1;
+  }
```
haaawk, line 288: Why are we adding one more dimension if there's a digit in the text?

libsql-sqlite3/src/vector.c
```diff
@@ -0,0 +1,558 @@
...
+    goto out_free;
+  }
+  if( dims1 != dims2 ){
+    sqlite3_result_error(context, "vectors must have the same length", -1);
+    goto out_free;
+  }
+  if( type1 != type2 ){
+    sqlite3_result_error(context, "vectors must have the same type", -1);
+    goto out_free;
+  }
```
haaawk, line 533: Would it make sense to check this before parsing vectors?

haaawk (commented):
Looks mostly good to me but I left a few nits and some questions.

libsql-sqlite3/src/vector.c
sivukhin: To distinguish between zero and one dimensional vectors (both have no commas)

libsql-sqlite3/src/vector.c
sivukhin: Yes, sure, we can make this check earlier

libsql-sqlite3/src/vector.c
haaawk: Ok I can see that now. Thanks!

libsql-sqlite3/src/vectorfloat32.c
haaawk: It is fine. Probably a comment would help to avoid confusion of the reader.
Also it seems that we can't remove that one extra char because it would make the lenght odd and this check would fail https://github.com/tursodatabase/libsql/pull/1531/files#diff-98fd1b40c774432f5288bd6ae7000632bbffb8687f15734977cb8cc794ed1a41R236

libsql-sqlite3/ext/crr/Makefile
penberg: Once you compile crr with vector, it needs sqrt for cosine distance function

haaawk (approved):

PR #1533, opened 2024-07-05
Make PRAGMA `synchronous` configurable for a Namespace

avinassh:
SQLite allows setting synchronous values to any four levels. This
patch provides an option to set the level for a namespace which will
be used for every connection. It defaults to `NORMAL`.

allowed values: https://www.sqlite.org/pragma.html#pragma_synchronous

e.g. to enable:

    curl -X POST http://sqld-admin-endpoint/v1/namespaces/<namespace>/config -H "Content-Type: application/json" -d '{"durability_mode": "strong"}' -v

### DIFF
--- a/libsql-replication/proto/metadata.proto
+++ b/libsql-replication/proto/metadata.proto
@@ -2,6 +2,13 @@ syntax = "proto3";
 
 package metadata;
 
+enum DurabilityMode {
+    RELAXED = 0;
+    STRONG = 1;
+    EXTRA = 2;
+    OFF = 3;
+}
+
 // Database config used to send db configs over the wire and stored
 // in the meta store.
 message DatabaseConfig {
@@ -12,11 +19,12 @@ message DatabaseConfig {
     // maximum db size (in pages)
     uint64 max_db_pages = 4;
     optional string heartbeat_url = 5;
-    optional string bottomless_db_id = 6; 
+    optional string bottomless_db_id = 6;
     optional string jwt_key = 7;
     optional uint64 txn_timeout_s = 8;
     bool allow_attach = 9;
     optional uint64 max_row_size = 10;
     optional bool shared_schema = 11;
     optional string shared_schema_name = 12;
+    DurabilityMode durability_mode = 13;
 }
--- a/libsql-replication/proto/metadata.proto
+++ b/libsql-replication/proto/metadata.proto
@@ -26,5 +26,5 @@ message DatabaseConfig {
     optional uint64 max_row_size = 10;
     optional bool shared_schema = 11;
     optional string shared_schema_name = 12;
-    DurabilityMode durability_mode = 13;
+    optional DurabilityMode durability_mode = 13;
 }
--- a/libsql-server/src/connection/config.rs
+++ b/libsql-server/src/connection/config.rs
@@ -113,19 +113,14 @@ impl From<&DatabaseConfig> for metadata::DatabaseConfig {
 }
 
 /// Durability mode specifies the `PRAGMA SYNCHRONOUS` setting for the connection
-#[derive(PartialEq, Clone, Copy, Debug, Deserialize, Serialize)]
+#[derive(PartialEq, Clone, Copy, Debug, Deserialize, Serialize, Default)]
 #[serde(rename_all = "lowercase")]
 pub enum DurabilityMode {
     Extra,
     Strong,
+    #[default]
     Relaxed,
     Off,
-}
-
-impl Default for DurabilityMode {
-    fn default() -> Self {
-        DurabilityMode::Relaxed
-    }
 }
 
 impl ToSql for DurabilityMode {
--- a/libsql-server/src/connection/config.rs
+++ b/libsql-server/src/connection/config.rs
@@ -1,11 +1,15 @@
 use crate::{namespace::NamespaceName, LIBSQL_PAGE_SIZE};
 use bytesize::mb;
+use rusqlite::types::ToSqlOutput;
+use rusqlite::ToSql;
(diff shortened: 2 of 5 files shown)


libsql-server/src/main.rs
```diff
@@ -255,6 +256,9 @@ struct Cli {
         default_value = "http://0.0.0.0:5002"
     )]
     storage_server_address: String,
+
+    #[clap(value_enum, long)]
+    pragma_synchronous: Option<PragmaSynchronous>,
```
avinassh, line 261: Couldn't think of a better parameter name, but I am open to suggestions.

libsql-replication/proto/metadata.proto
```diff
@@ -12,11 +12,12 @@ message DatabaseConfig {
...
     optional string heartbeat_url = 5;
-    optional string bottomless_db_id = 6; 
+    optional string bottomless_db_id = 6;
     optional string jwt_key = 7;
     optional uint64 txn_timeout_s = 8;
     bool allow_attach = 9;
     optional uint64 max_row_size = 10;
     optional bool shared_schema = 11;
     optional string shared_schema_name = 12;
+    optional string durability_mode = 13;
```
MarinPostma, line 22: I think we should use an enum

libsql-server/src/connection/config.rs
```diff
@@ -29,6 +33,8 @@ pub struct DatabaseConfig {
     pub is_shared_schema: bool,
     #[serde(default)]
     pub shared_schema_name: Option<NamespaceName>,
+    #[serde(default)]
+    pub durability_mode: Option<DurabilityMode>,
```
MarinPostma, line 37: Since we have an off mode, I think the option is not necessary. Let'a implement default for mode and use that directly

libsql-server/src/connection/config.rs
```diff
@@ -96,6 +107,54 @@ impl From<&DatabaseConfig> for metadata::DatabaseConfig {
...
+
+impl Display for DurabilityMode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let m = match self {
+            DurabilityMode::Extra => "extra".to_string(),
+            DurabilityMode::Strong => "strong".to_string(),
+            DurabilityMode::Relaxed => "relaxed".to_string(),
+            DurabilityMode::Off => "off".to_string(),
+        };
+        write!(f, "{}", m)
```
MarinPostma, line 158:
```suggestion
        let m = match self {
            DurabilityMode::Extra => "extra",
            DurabilityMode::Strong => "strong",
            DurabilityMode::Relaxed => "relaxed",
            DurabilityMode::Off => "off",
        };
        write!(f, "{m}")
```

libsql-server/src/connection/libsql.rs
```diff
@@ -455,6 +455,10 @@ impl<W: Wal> Connection<W> {
 
         let config = config_store.get();
         conn.pragma_update(None, "max_page_count", config.max_db_pages)?;
+        if let Some(mode) = config.durability_mode {
+            tracing::info!("setting PRAGMA synchronous to {:?}", mode);
```
MarinPostma, line 459:
```suggestion
            tracing::info!("setting PRAGMA synchronous to {mode:?}");
```
Clippy is gonna complain

MarinPostma (changes requested):

libsql-replication/proto/metadata.proto
```diff
@@ -12,11 +19,12 @@ message DatabaseConfig {
...
     optional string heartbeat_url = 5;
-    optional string bottomless_db_id = 6; 
+    optional string bottomless_db_id = 6;
     optional string jwt_key = 7;
     optional uint64 txn_timeout_s = 8;
     bool allow_attach = 9;
     optional uint64 max_row_size = 10;
     optional bool shared_schema = 11;
     optional string shared_schema_name = 12;
+    DurabilityMode durability_mode = 13;
```
haaawk, line 29: Should this be optional?

haaawk (approved):

libsql-replication/proto/metadata.proto
avinassh: I am wondering if that is needed. Because in the `DatabaseConfig` it is not optional, set to default `Relaxed`. So if we make this optional, for existing DBs, it will be `None` only at the first read. When the `DatabaseConfig` is saved, it will be set to the default value.

For newer DBs, this will always have a default value from `DatabaseConfig` and will never be `None`

WDYT?

MarinPostma: we should adhere to that: https://protobuf.dev/getting-started/pythontutorial/#extending-a-protobuf

avinassh: that is fair, I will update the PR

libsql-server/src/connection/config.rs
```diff
@@ -96,6 +107,85 @@ impl From<&DatabaseConfig> for metadata::DatabaseConfig {
...
+    Strong,
+    Relaxed,
+    Off,
+}
+
+impl Default for DurabilityMode {
+    fn default() -> Self {
+        DurabilityMode::Relaxed
+    }
+}
```
MarinPostma, line 129: nit:
```suggestion
#[derive(PartialEq, Clone, Copy, Debug, Deserialize, Serialize, Default)]
#[serde(rename_all = "lowercase")]
pub enum DurabilityMode {
    Extra,
    Strong,
    #[default]
    Relaxed,
    Off,
}
```

MarinPostma (approved):

PR #1535, opened 2024-07-05
connection pooling

MarinPostma:
Create a pool of ready connections to reduce connection create time.

### DIFF
--- a/libsql-server/src/connection/libsql.rs
+++ b/libsql-server/src/connection/libsql.rs
@@ -10,7 +10,7 @@ use libsql_sys::EncryptionConfig;
 use metrics::histogram;
 use parking_lot::Mutex;
 use rusqlite::ffi::SQLITE_BUSY;
-use rusqlite::{ErrorCode, OpenFlags};
+use rusqlite::OpenFlags;
 use tokio::sync::watch;
 use tokio::time::{Duration, Instant};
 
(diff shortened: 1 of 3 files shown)


libsql-server/src/connection/libsql.rs
```diff
@@ -76,60 +60,62 @@ where
...
+                wal_wrapper,
+                encryption_config,
+                block_writes,
+                resolve_attach_path,
+                connection_manager: ConnectionManager::new(txn_timeout),
+                make_wal_manager,
+            };
 
-    /// Tries to create a database, retrying if the database is busy.
-    async fn try_create_db(&self) -> Result<LibSqlConnection<W>> {
```
sivukhin, line 105: Why do we no longer need these retries?

libsql-server/src/connection/libsql.rs
```diff
@@ -76,60 +60,62 @@ where
...
-            connection_manager: ConnectionManager::new(txn_timeout),
-            make_wal_manager,
-        };
-
-        let db = this.try_create_db().await?;
-        this._db = Some(db);
-
-        Ok(this)
-    }
+        let (snd, recv) = async_channel::bounded(16);
```
sivukhin, line 63: Should we make this parameter configurable?

sivukhin, line 63: Noob rust question: `snd` ownership will be transferred to the tokio coroutine or how does this work here?

libsql-server/src/connection/libsql.rs
```diff
@@ -76,60 +60,62 @@ where
...
-                    )),
-                ) => {
-                    if retries < 100 {
-                        tracing::warn!("Database file is busy, retrying...");
-                        retries += 1;
-                        tokio::time::sleep(Duration::from_millis(100)).await
-                    } else {
-                        Err(err)?;
-                    }
+            loop {
```
sivukhin, line 83: Does sqld have some graceful shutdown logic? Should we stop this coroutine at some point and reclaim all resources?

libsql-server/src/connection/libsql.rs
```diff
@@ -76,60 +60,62 @@ where
...
+                let conn = inner.make_connection().await;
```
sivukhin, line 84: Just asking for the sake of clarity: it's not a connection pool in sense that we don't return anything to that pool (should we?) - we just want to bootstrap few connections in a background in order to remove initialization delay from request latencies.

BTW - after first 16 connections we will create them one by one. Should we try to maintain "free pool" size at certain level in order to guarantee same behaviour for startup and after it for sqld?

MarinPostma:
@sivukhin i should have put this into draft, just experimenting right now, I opened the or for athos to try it out

libsql-server/src/connection/libsql.rs
MarinPostma: It's some legacy stuff, trying to see if that works now, and I'd rather pull connection out of the sender until they stop erroring out

libsql-server/src/connection/libsql.rs
MarinPostma: Yes, and this contouring will exit when we drop all the receivers, this will happen when the namespace is deallocated

libsql-server/src/connection/libsql.rs
MarinPostma: Yeah not really a pool in the sense that connections are reused.

Definitely, the pool size will be configurable, and I'm also thinking about making it grow for namespaces that see load, something like "if the queue ever goes empty, bump it's size"

MarinPostma: Also, as soon as a connection is popped, a new one is added. The loop tries to maintain a full queue

libsql-server/src/connection/libsql.rs
MarinPostma: That's correct, it's moved in the async block

MarinPostma: I answered in another thread but yeah definitely, though I think the setting should be an upper bound, and we should bump the channel size depending on the actual load

MarinPostma: Iow, start small and bump as necessary

PR #1538, opened 2024-07-08
introduce checkpointer

MarinPostma:
introduce the checkpointer.

The checkpointer performs checkpoint for databses in the background. When a namespace is ready to be checkpointed, the checkpointer if notified, and will schedule the database for checkpointing.

### DIFF
--- a/libsql-wal/src/checkpointer.rs
+++ b/libsql-wal/src/checkpointer.rs
@@ -1,0 +1,382 @@
+use std::future::Future;
+use std::sync::Arc;
+
+use hashbrown::HashSet;
+use libsql_sys::name::NamespaceName;
+use tokio::sync::mpsc;
+use tokio::task::JoinSet;
+
+pub(crate) type NotifyCheckpointer = mpsc::Sender<NamespaceName>;
+
+pub trait PerformCheckpoint{
+    fn checkpoint(&self, namespace: &NamespaceName) -> impl Future<Output = crate::error::Result<()>> + Sync + Send;
+}
+
+const CHECKPOINTER_ERROR_THRES: usize = 16;
+
+/// The checkpointer checkpoint wal segments in the main db file, and deletes checkpointed
+/// segments.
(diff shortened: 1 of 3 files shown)


libsql-wal/src/checkpointer.rs
```diff
@@ -0,0 +1,416 @@
...
+    #[tracing::instrument(skip(self))]
+    pub async fn run(mut self) {
+        loop {
+            if self.should_exit() {
+                tracing::info!("checkpointer exited cleanly.");
+                return
+            }
+
+            if self.errors > CHECKPOINTER_ERROR_THRES {
+                todo!("handle too many consecutive errors");
```
MarinPostma, line 103: I'll deal with that later

libsql-wal/src/checkpointer.rs
```diff
@@ -0,0 +1,425 @@
...
+use tokio::task::JoinSet;
+
+use crate::io::Io;
+use crate::registry::WalRegistry;
+
+pub(crate) type NotifyCheckpointer = mpsc::Sender<NamespaceName>;
+
+type LibsqlCheckpointer<IO, S> = Checkpointer<WalRegistry<IO, S>>;
+
+impl<IO, S> LibsqlCheckpointer<IO, S>
```
LucioFranco, line 16: I find this really awkward to read with an impl for a type alias, I would maybe suggest newtyping at least?

libsql-wal/src/checkpointer.rs
```diff
@@ -0,0 +1,425 @@
...
+
+impl<IO, S> PerformCheckpoint for WalRegistry<IO, S>
+where
+    IO: Io,
+    S: Sync + Send + 'static,
+{
+    fn checkpoint(
+        &self,
+        namespace: &NamespaceName,
+    ) -> impl Future<Output = crate::error::Result<()>> + Send {
```
LucioFranco, line 45: this can be an async fn, only the trait def needs the impl future + send

PR #1543, opened 2024-07-09
introduce first iteration of the zig build

levydsa:


levydsa:
This tests seem to fail only because of the lack of the zig binary on the PATH. I don't have experience with this, can someone try guide me on how to fix that?

penberg:
Hey @levydsa! There's bunch of changes to `libsql-sqlite3/src` which I would prefer not to see for a build system change because it makes merging from upstream harder. What's the reason for them? For example, there's one that changes struct initialization to use C99 style, which looks dubious to me.

levydsa:
The fields of that struct were changed at some point and this definition was incorrect. You can see a warning in the old build of a function pointer type mismatch. Zig makes this a error by default. Using struct initialization prevents this from happening again. I can revert if it I was actually correct. Other changes are to make it easier to build without the amalgamation, but no substantial change other than some additional declarations and includes. Can you point out some other places that could cause problems?

penberg:
Hey @levydsa, ideally a build system change like this changes _nothing_ in `libsql3-sqlite/src` directory. So let's turn off all the default warnings and so forth and only change things that are absolutely required to have a Zig build.

levydsa:
Ok, then. I'll do my best to preserve `src/`.

levydsa:
@penberg Done! I can't remove more than that without having to manually concat files to expose declarations.

penberg:
We don't have plans to switch to the Zig build so closing this.

PR #1545, opened 2024-07-10
Improve local cache and prefer local reads

avinassh:
Till now, `DurableWAL` used Storage Server for finding page versions. If the local cache's `max_frame_no` is same as storage server's and if a txn's `max_frame_no` is less or same as cache's `max_frame_no`, then all reads and writes can proceed using local data itself. Do note that writes from a stale primary will be rejected anyways, since we handle conflicts at Storage Server.

Currently, there is no replication built and it assumes a single primary writer. So, after write, it updates the local cache. In the future PRs, I will add replication. 

What's changed: 
- Update the cache after writes
- Use local cache for most operations

What's next:
- A primary can lag anytime, so we need replication or way to quickly bring up the cache to speed
- If the primary is lagging, we switch to remote mode

### DIFF
--- a/libsql-storage/src/lib.rs
+++ b/libsql-storage/src/lib.rs
@@ -23,12 +23,7 @@ pub mod rpc {
 // What does (not) work:
 // - there are no read txn locks nor upgrades
 // - no lock stealing
-// - write set is kept in mem
-// - txns don't use max_frame_no yet
 // - no savepoints, yet
-// - no multi tenancy, uses `default` namespace
-// - txn can read new frames after it started (since there are no read locks)
-// - requires huge memory as it assumes all the txn data fits in the memory
 
 #[derive(Clone, Default)]
 pub struct DurableWalConfig {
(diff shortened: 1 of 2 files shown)


libsql-storage/src/lib.rs
```diff
@@ -201,8 +196,16 @@ impl Wal for DurableWal {
...
+        let remote_frame_no = tokio::task::block_in_place(|| rt.block_on(self.frames_count()));
+        let local_frame_no = self.local_cache.get_max_frame_num().unwrap();
+        if local_frame_no < remote_frame_no {
+            warn!(
+                "cache (max_frame_num = {:?}) is lagging behind the remote(max_frame_num = {:?})!",
+                local_frame_no, remote_frame_no
+            )
+            // TODO: replicate data from source
+        }
+        self.max_frame_no = local_frame_no;
```
avinassh, line 208: In the follow up PR, I will switch to `remote` mode if I find primary is lagging.

PR #1546, opened 2024-07-11
add benchmark scripts

sivukhin:
## Context

Several benchmark tools which I use for vector feature. This PR includes following scripts:
- `benchtest` - generic tool which measure performance of any sequence of SQL statements
- `anntest` - tool for measuring `n-recall@k` of ANN vector search
- `blobtest` - simple tool which aims to prove that `sqlite3_blob_reopen` API can result in way better performance

See [README.md](https://github.com/tursodatabase/libsql/blob/vector-benchmark-tools/libsql-sqlite3/benchmark/README.md) for more details


penberg:
Nice, @sivukhin!

haaawk:
In anntest you call `srand(time(NULL));` at the beginning of main but you don't do that in the other two tests. Is this on purpose?

haaawk (approved):

sivukhin:
@haaawk - I don't need `srand` in the `anntest` actually. This is "legacy" - initially I generated ANN workload right inside `anntest` C code but later decided to move it to the python `workload.py` script as it's easier to generate random floats with different distributions from there.

Removed it from the branch

PR #1554, opened 2024-07-14
libsql wal s3 backend

MarinPostma:
implement the S3 backend for AsyncStorage that stores segments on S3. Also add a small CLI to test libsql wal


libsql-wal/src/storage/async_storage.rs
```diff
@@ -128,31 +154,62 @@ where
...
+        rcv.blocking_recv().unwrap()
     }
 }
 
-pub struct AsyncStorageInitConfig<S> {
-    storage: S,
-    max_in_flight_jobs: usize,
+pub struct AsyncStorageInitConfig<B> {
+    pub backend: Arc<B>,
+    pub max_in_flight_jobs: usize,
```
LucioFranco, line 197: can we avoid these pubs?

libsql-wal/src/storage/backend/s3.rs
```diff
@@ -1,126 +1,471 @@
...
+        let body = SdkBody::retryable(move || {
+            let s = Self::new_inner(self.inner.clone());
+            SdkBody::from_body_1_x(s)
+        });
+
+        ByteStream::new(body)
+    }
+}
+
+impl<F> http_body::Body for FileStreamBody<F>
```
LucioFranco, line 416: I don't totally follow why we need a custom body implementation? Might be worth adding a comment why here

LucioFranco (approved):
I have a few questions but approving to move this forward.

libsql-wal/src/storage/async_storage.rs
MarinPostma: it's a config struct

libsql-wal/src/storage/backend/s3.rs
MarinPostma: I couldn't find how to wrap a stream, but I reckon we may have used https://docs.rs/stream-body/latest/stream_body/

PR #1557, opened 2024-07-15
Vector aux functions

sivukhin:
## Context

Second branch in the series for DiskANN implementation. This PR introduce few utility functions and classes which aims to simplify interop between SQLite & DiskANN implementation

- `VectorIdxParams` - binary container of vector index params. Instead of storing parameters in the rigid schema we decided to use binary format over which we have full control (so we can add parameters dynamically without complicated schema changes and control default values easily)
    * The format by itself is very simply: every value in container spans 9 bytes when first byte interpreted as a tag and 8 next bytes interpreted as an u64 integer or f64 float
- `VectorInRow` - container which stores input row for `INSERT`/`DELETE` operations
- `VectorOutRows` - container which stores output rows for `SEARCH` operation

Table of supported index parameters for now:
```
static struct VectorParamName VECTOR_PARAM_NAMES[] = { 
  { "type",     VECTOR_INDEX_TYPE_PARAM_ID,    0 /* string */, "diskann", VECTOR_INDEX_TYPE_DISKANN },
  { "metric",   VECTOR_METRIC_TYPE_PARAM_ID,   0 /* string */, "cosine", VECTOR_METRIC_TYPE_COS },
  { "alpha",    VECTOR_PRUNING_ALPHA_PARAM_ID, 2 /* float */,   0, 0 },
  { "search_l", VECTOR_SEARCH_L_PARAM_ID,      1 /* integer */, 0, 0 },
  { "insert_l", VECTOR_INSERT_L_PARAM_ID,      2 /* integer */, 0, 0 },
};
```

For example, this is correct index creation statement (in follow up branches):
```sql
CREATE INDEX t_idx ON t ( 
    libsql_vector_idx(emb, 'type=diskann', 'metric=cosine', 'alpha=1.2', 'search_l=70', 'insert_l=120') 
);
```
But also we will provide reasonable defaults and user still can simply write:
```sql
CREATE INDEX t_idx ON t ( libsql_vector_idx(emb ) );
```
## Testing

- Simple test cases for `VectorIdxParams` added in `test_libsql_diskann.c` file
- Other classes are pretty hard to test in isolation - so I decided to leave them without unit tests (but integration tests are added in the following branches)

### DIFF
--- a/libsql-sqlite3/src/vectorIndex.c
+++ b/libsql-sqlite3/src/vectorIndex.c
@@ -1,0 +1,462 @@
+/*
+** 2024-03-18
+**
+** Copyright 2024 the libSQL authors
+**
+** Permission is hereby granted, free of charge, to any person obtaining a copy of
+** this software and associated documentation files (the "Software"), to deal in
+** the Software without restriction, including without limitation the rights to
+** use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+** the Software, and to permit persons to whom the Software is furnished to do so,
+** subject to the following conditions:
+**
+** The above copyright notice and this permission notice shall be included in all
+** copies or substantial portions of the Software.
+**
+** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+** FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+** COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+** IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+** CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+**
+******************************************************************************
+**
+** libSQL vector search.
+*/
+#ifndef SQLITE_OMIT_VECTOR
+#include "sqlite3.h"
+#include "vdbeInt.h"
+#include "sqliteInt.h"
+#include "vectorIndexInt.h"
+
+/**************************************************************************
+** VectorIdxParams utilities
+****************************************************************************/
+
+void vectorIdxParamsInit(VectorIdxParams *pParams, u8 *pBinBuf, int nBinSize) {
+  assert( nBinSize <= VECTOR_INDEX_PARAMS_BUF_SIZE );
+
+  pParams->nBinSize = nBinSize;
+  if( pBinBuf != NULL ){
+    memcpy(pParams->pBinBuf, pBinBuf, nBinSize);
+  }
+}
+
+u64 vectorIdxParamsGetU64(const VectorIdxParams *pParams, char tag) {
+  int i, offset;
+  u64 value = 0;
+  for (i = 0; i + 9 <= pParams->nBinSize; i += 9){
+    if( pParams->pBinBuf[i] != tag ){
+      continue;
+    }
+    // choose latest value from the VectorIdxParams bin
+    value = 0;
+    for(offset = 0; offset < 8; offset++){
+      value |= ((u64)(pParams->pBinBuf[i + 1 + offset]) << (u64)(8 * offset));
+    }
+  }
+  return value;
+}
+
+int vectorIdxParamsPutU64(VectorIdxParams *pParams, char tag, u64 value) {
+  int i;
+  if( pParams->nBinSize + 9 > VECTOR_INDEX_PARAMS_BUF_SIZE ){
+    return -1;
+  }
+  pParams->pBinBuf[pParams->nBinSize++] = tag;
+  for(i = 0; i < 8; i++){
+    pParams->pBinBuf[pParams->nBinSize++] = value & 0xff;
+    value >>= 8;
+  }
+  return 0;
+}
+
+double vectorIdxParamsGetF64(const VectorIdxParams *pParams, char tag) {
+  u64 value = vectorIdxParamsGetU64(pParams, tag);
+  return *((double*)&value);
+}
+
+int vectorIdxParamsPutF64(VectorIdxParams *pParams, char tag, double value) {
+  return vectorIdxParamsPutU64(pParams, tag, *((u64*)&value));
+}
+
+/**************************************************************************
+** VectorIdxKey utilities
+****************************************************************************/
+
+int vectorIdxKeyGet(Table *pTable, VectorIdxKey *pKey, const char **pzErrMsg) {
+  int i;
+  Index *pPk;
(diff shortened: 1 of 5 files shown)


libsql-sqlite3/src/vectorIndex.c
```diff
@@ -0,0 +1,462 @@
...
+    if( pParams->pBinBuf[i] != tag ){
+      continue;
+    }
+    // choose latest value from the VectorIdxParams bin
+    value = 0;
+    for(offset = 0; offset < 8; offset++){
+      value |= ((u64)(pParams->pBinBuf[i + 1 + offset]) << (u64)(8 * offset));
+    }
+  }
+  return value;
```
haaawk, line 59: Would it make sense to go backwards and stop at first element that has the right tag?

libsql-sqlite3/src/vectorIndex.c
```diff
@@ -0,0 +1,462 @@
...
+    }else {
+      size = snprintf(pBuf, nBufSize, ",%s%d", prefix, i);
+    }
+    if( size < 0 ){
+      return -1;
+    }
+    pBuf += size;
+    nBufSize -= size;
+  }
+  if( nBufSize <= 0 ){
```
haaawk, line 158: isn't it possible to finish the loop above with nBufSize == 0 but still correctly?
Wouldn't a check `if( nBufSize < 0 || ( i < nKeyColumns && nBufSize == 0 ) )` be better here?

libsql-sqlite3/src/vectorIndex.c
```diff
@@ -0,0 +1,462 @@
...
+** VectorInRow utilities
+****************************************************************************/
+
+sqlite3_value* vectorInRowKey(const VectorInRow *pVectorInRow, int iKey) {
+  assert( 0 <= iKey && iKey < pVectorInRow->nKeys );
+  return pVectorInRow->pKeyValues + iKey;
+}
+
+i64 vectorInRowLegacyId(const VectorInRow *pVectorInRow) {
+  if( pVectorInRow->nKeys == 1 && sqlite3_value_type(pVectorInRow->pKeyValues + 0) == SQLITE_INTEGER ){
```
haaawk, line 174: `+ 0`?

libsql-sqlite3/src/vectorIndex.c
```diff
@@ -0,0 +1,462 @@
...
+    *(pBuf++) = ',';
+  }
+  *(pBuf - 1) = '\0';
+  return 0;
+}
+
+int vectorInRowAlloc(sqlite3 *db, const UnpackedRecord *pRecord, VectorInRow *pVectorInRow, char **pzErrMsg) {
+  int rc = SQLITE_OK;
+  int type, dims;
+  struct sqlite3_value *pVectorValue = pRecord->aMem + 0;
```
haaawk, line 208: `+ 0`?

libsql-sqlite3/src/vectorIndex.c
```diff
@@ -0,0 +1,462 @@
...
+    // users can put strings (e.g. '[1,2,3]') in the table and we should process them correctly
+    if( vectorParse(pVectorValue, pVectorInRow->pVector, pzErrMsg) != 0 ){
+      rc = SQLITE_ERROR;
+      goto out;
+    }
+  }
+  rc = SQLITE_OK;
+out:
+  if( rc != SQLITE_OK ){
+    vectorFree(pVectorInRow->pVector);
```
haaawk, line 246: Is it ok to free this vector before it's allocated? Line 225 can jump here with rc != SQLITE_OK

haaawk: So can a failure in vectorAlloc

libsql-sqlite3/src/vectorIndex.c
```diff
@@ -0,0 +1,462 @@
...
+  pRows->nRows = nRows;
+  pRows->nCols = nCols;
+  pRows->aRowids = NULL;
+  pRows->ppValues = NULL;
+
+  if( (u64)nRows * (u64)nCols > VECTOR_OUT_ROWS_MAX_CELLS ){
+    return SQLITE_NOMEM_BKPT;
+  }
+
+  if( nCols == 1 && firstColumnAff == SQLITE_AFF_INTEGER ){
```
haaawk, line 270: Is this always true that when there's 1 integer column then it will be a rowid? What about the tables with no rowid that have single integer primary key?

sivukhin: I must rename `aRowids` array to something more generic (`aIntColumn` maybe).

This code actually just wants to optimize use case when we need to return single integer column in the output.

libsql-sqlite3/src/vectorIndex.c
sivukhin: Yes, this is bug. I guess I forgot ` && pVectorInRow->pVector != NULL` guard here :-(

libsql-sqlite3/src/vectorIndex.c
sivukhin: Yes, this is bit convoluted. But, `pVectorInRow->pKeyValues` is an array actually and here I wanted to highlight that we are accessing first element of this array.

I can write something like `&pVectorInRow->pKeyValues[0]` instead - but not sure if this will make more sense...

haaawk (approved):

libsql-sqlite3/src/vectorIndex.c
haaawk: I understand now. Thanks

libsql-sqlite3/src/vectorIndex.c
haaawk: [0] is probably a bit more idiomatic - I don't know

libsql-sqlite3/src/vectorIndex.c
sivukhin: I think we must have one free byte in the buffer because sprintf "returns the number of characters that would have been written if n had been sufficiently large, **not counting the terminating null character**."

libsql-sqlite3/src/vectorIndex.c
sivukhin: Thats actually open question. Initially I thought that current logic will allow us easily implement "overrides" - so it has bit more flexibility in a sense.

But now I'm actually not sure if it's actually useful feature of the format...

libsql-sqlite3/src/vectorIndex.c
haaawk: Makes sense. Thanks!

libsql-sqlite3/src/vectorIndex.c
haaawk: So going backwards and stoping at the first element with the right tag would be functionally the same as the code in the PR it would just maybe be simpler/more performant as we would compose value only once.

sivukhin: I decided to leave current implementation as is.

Actually, since I want this parsing method to be safe, it became bit tricker as you need to calculate starting offset from the end (I don't want to rely on fact that payloads always will be multiple of 9; maybe in future we will add more compact parameters representation which will be controlled by tag bits).

haaawk: Makes senes.

PR #1560, opened 2024-07-16
Vector diskann impl

sivukhin:
## Context

Third branch in the series for DiskANN implementation. This PR introduce DiskANN algorithm itself :tada: 

The algorithm core based on the paper [LM-DiskANN: Low Memory Footprint in
Disk-Native Dynamic Graph-Based ANN Indexing](https://cse.unl.edu/~yu/homepage/publications/paper/2023.LM-DiskANN-Low%20Memory%20Footprint%20in%20Disk-Native%20Dynamic%20Graph-Based%20ANN%20Indexing.pdf)

Current implementation can be suboptimal in some places due to usage of very trivial data structures (lists / arrays). That's how it was intended - we will improve performance in the subsequent PRs but make this one simpler.

Nevertheless, this PR tries to address most heavy operation in the algorithm - reading blobs from disk - and aims to make as little reads as possible (utilizing `sqlite3_blob_reopen` if possible). 

From the performance perspective rough hierarchy of operation cost looks like this:
1. read/write diskann node block
2. distance calculation between two vectors (for example, openai vectors are 1536 dimensions long - so this is clearly very slow operation)
3. operations with non-compressed vector payload
4. all other operations

## Changes

- Implement DiskANN algorithm
- Adjust `BlobSpot` code for cases when previous call to open/reopen failed (and blob became useless as all subsequent operations will return `SQLITE_ABORT` error; read `sqlite3_blob_*` docstrings for more details)
- Add basic test in the `test_libsql_diskann.c` (it's a bit hacky for now - but it's fine since we soon will get merge full integration with SQLite and will no longer need these tests anymore, actually)


libsql-sqlite3/src/vectordiskann.c
```diff
@@ -67,6 +73,34 @@
...
+  u64 nRowid;           /* node id */
+  int visited;          /* is this node visited? */
+  DiskAnnNode *pNext;   /* next node in the visited list */
+  BlobSpot *pBlobSpot;  /* reference to the blob with node data (can be NULL when data actually is not needed; for example - node waiting in the queue) */
+};
+
+/*
+ * DiskAnnSearchCtx stores information required for search operation to succeed
+ *
+ * search context usually "borrows" candidates (storing them in aCandidates or visitedList)
```
haaawk, line 90: by "borrows" you mean takes ownership?

sivukhin: Yes

libsql-sqlite3/src/vectordiskann.c
```diff
@@ -345,4 +403,1048 @@ void nodeBinDebug(const DiskAnnIndex *pIndex, const BlobSpot *pBlobSpot) {
...
+  char *zSql = sqlite3MPrintf(db, "DROP TABLE %s_shadow", zIdxName);
+  int rc = sqlite3_exec(db, zSql, 0, 0, 0);
+  sqlite3DbFree(db, zSql);
+  return rc;
+}
+
+/*
+ * Select random row from the shadow table and set its rowid to pRowid
+ * returns SQLITE_DONE if no row found (this will be used to determine case when table is empty)
+ * TODO: we need to make this selection procedure faster - now it works in linear time
```
haaawk, line 484: Would
```
SELECT rowid 
FROM %s
LIMIT 1 
OFFSET ABS(RANDOM()) % MAX((SELECT COUNT(*) FROM %s), 1)
```
be better?

haaawk: For 10000 rows it's already 10x faster
```
sqlite> select count(*) from t;
QUERY PLAN
`--SCAN t
10004
Run Time: real 0.000 user 0.000083 sys 0.000071
sqlite> select rowid from t limit 1 offset (abs(random()) % (select count(*) from t));
QUERY PLAN
|--SCALAR SUBQUERY 1
|  `--SCAN t
`--SCAN t
8514
Run Time: real 0.001 user 0.000260 sys 0.000067
sqlite> select rowid from t order by random() limit 1;
QUERY PLAN
|--SCAN t
`--USE TEMP B-TREE FOR ORDER BY
5910
Run Time: real 0.002 user 0.002724 sys 0.000130
```

libsql-sqlite3/src/vectordiskann.c
```diff
@@ -345,4 +403,1048 @@ void nodeBinDebug(const DiskAnnIndex *pIndex, const BlobSpot *pBlobSpot) {
...
+  }
+  rc = sqlite3_prepare_v2(pIndex->db, zSql, -1, &pStmt, 0);
+  if( rc != SQLITE_OK ){
+    goto out;
+  }
+  rc = sqlite3_step(pStmt);
+  if( rc != SQLITE_ROW ){
+    goto out;
+  }
+  *pRowid = sqlite3_column_int64(pStmt, 0);
```
haaawk, line 504: Don't we have to check:
```
  if( sqlite3_column_type(pStmt, 0) != SQLITE_INTEGER ){
    rc = SQLITE_ERROR;
    goto out;
  }
```
like in the function below?

libsql-sqlite3/src/vectordiskann.c
```diff
@@ -345,4 +403,1048 @@ void nodeBinDebug(const DiskAnnIndex *pIndex, const BlobSpot *pBlobSpot) {
...
+  return rc;
+}
+
+/*
+ * Delete row from the shadow table
+*/
+static int diskAnnDeleteShadowRow(const DiskAnnIndex *pIndex, i64 nRowid){
+  int rc;
+  sqlite3_stmt *pStmt = NULL;
+  char *zSql = sqlite3MPrintf(pIndex->db, "DELETE FROM %s_shadow WHERE rowid = ?", pIndex->zName);
```
haaawk, line 684: Would it make sense to use pIndex->zShadow and %s here instead of pIndex->zName and %s_shadow?

libsql-sqlite3/src/vectordiskann.c
```diff
@@ -345,4 +403,1048 @@ void nodeBinDebug(const DiskAnnIndex *pIndex, const BlobSpot *pBlobSpot) {
...
+    float nodeToEdge, hintToEdge;
+    if( i == iInserted ){
+      i++;
+      continue;
+    }
+    nodeBinEdge(pIndex, pNodeBlob, i, NULL, &edgeVector);
+    nodeToEdge = diskAnnVectorDistance(pIndex, &nodeVector, &edgeVector);
+    hintToEdge = diskAnnVectorDistance(pIndex, &hintVector, &edgeVector);
+    if( nodeToEdge > pIndex->pruningAlpha * hintToEdge ){
+      nodeBinDeleteEdge(pIndex, pNodeBlob, i);
```
haaawk, line 979: This is quadratic, no? It feels that we could delete edges and compress them only once to not move the elements too many time, no?

haaawk (approved):

libsql-sqlite3/src/vectordiskann.c
sivukhin: Great suggestion, thanks!

libsql-sqlite3/src/vectordiskann.c
sivukhin: I added asserts in similar places everywhere (since we have full control over the query - we actually can just assert this invariant, I think)

libsql-sqlite3/src/vectordiskann.c
sivukhin: Yes, fixed this by changing `nodeBin{Delete,Insert}Edge` implementations: now insert just replace edge at the position, and delete just swap current edge with last edge in the array

PR #1566, opened 2024-07-17
Add libsql_open_sync_with_config

### LINKED ISSUE #1561: Feature Request: Expose `sync_interval` parameter in libsql's C bindings
Currently, there is no exposed `sync_interval` parameter in libsql's C bindings, which limits the flexibility of custom implementations built from source when using libsql's embedded replicas.

It would be beneficial to expose a `sync_interval` parameter in the C bindings, similar to how it's implemented in the TypeScript/JavaScript client API for libSQL. This would allow C implementations to configure automatic periodic sync.

For reference, the TypeScript/JavaScript client API for libSQL allows setting a sync interval like so:

```typescript
const client = createClient({
  url: 'file:path/to/db-file.db',
  authToken: '...',
  syncUrl: '...',
  syncInterval: 60,
});
```

haaawk:
Expose interval sync in C bindings.

Fixes #1561

### DIFF
--- a/bindings/c/include/libsql.h
+++ b/bindings/c/include/libsql.h
@@ -27,6 +27,16 @@ typedef struct libsql_stmt libsql_stmt;
 
 typedef const libsql_database *libsql_database_t;
 
+typedef struct {
+  const char *db_path;
+  const char *primary_url;
+  const char *auth_token;
+  char read_your_writes;
+  const char *encryption_key;
+  int sync_interval;
+  char with_webpki;
+} libsql_config;
+
 typedef const libsql_connection *libsql_connection_t;
 
 typedef const libsql_stmt *libsql_stmt_t;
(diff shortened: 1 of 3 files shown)


penberg:
I am fine with this, but curious to know if we can make the API simpler with a `libsql_config` struct. Is that something we can easily use from Go, for example?

haaawk:
> I am fine with this, but curious to know if we can make the API simpler with a `libsql_config` struct. Is that something we can easily use from Go, for example?

I would have to check. That would also require update to op-sqlite too.

haaawk:
Let me know if this looks better @penberg

haaawk:
I was able to switch go-libsql to use libsql_config and all the tests have passed

haaawk:
I was able to switch go-libsql to use libsql_config and all the tests passed -> https://github.com/tursodatabase/go-libsql/pull/32

penberg (approved):

PR #1569, opened 2024-07-17
libsql wal restore

MarinPostma:


libsql-wal/src/bins/shell/main.rs
```diff
@@ -163,14 +178,27 @@ where
...
-                    conn.query_row(&q, (), |row| {
-                        println!("{row:?}");
-                        Ok(())
-                    })
-                    .optional()
-                }) {
-                    println!("error: {e}");
+                match block_in_place(|| conn.prepare(&q)) {
+                    Ok(mut stmt) => {
+                        match block_in_place(|| stmt.query_map((), |row| {
```
LucioFranco, line 183: how does nesting these work?

libsql-wal/src/bins/shell/main.rs
```diff
@@ -204,33 +232,30 @@ async fn handle_builtin<S>(
...
+        .endpoint_url(cli.s3_args.s3_url.as_ref().unwrap())
         .build();
     let backend = Arc::new(
         S3Backend::from_sdk_config(
             config.clone(),
-            bucket_name.to_string(),
-            cluster_id.to_string(),
+            cli.s3_args.s3_bucket.as_ref().unwrap().to_string(),
+            cli.s3_args.cluster_id.as_ref().unwrap().to_string(),
         )
```
LucioFranco, line 259: I think this should use the creds resolver that is built in so that it pulls aws creds from your `~/.aws/` dir or env vars instead of doing custom credential resolution.

LucioFranco (approved):

libsql-wal/src/bins/shell/main.rs
MarinPostma: do you know how to do that? can you submit a PR?

libsql-wal/src/bins/shell/main.rs
MarinPostma: they are not nested :)

PR #1571, opened 2024-07-18
integrate diskann to the sqlite code

sivukhin:
## Context

Fourth and final branch in the initial series for DiskANN implementation which glue all things together and integrate DiskANN into the libSQL completely 🎉🎉🎉

This PR hooks into the SQLite to make DiskANN indices work:
1. `idxType` enumeration extended with new `SQLITE_IDXTYPE_VECTOR = 4` (note the if mask `0b11 = 3` will be applied to `idxType` then vector index type "became" `SQLITE_IDXTYPE_APPDEF`)
2. `OP_OpenVectorIdx` VM op-code were added - although we maybe can change `OP_OpenWrite` op-code behaviour - new op-code approach is more safe and give us more freedom and safety. For example, `P5` register serve it's own need for new op code and if `OPFLAG_FORDELETE` is set then we will perform truncate of index before allocating new cursor (this is needed for `REINDEX` command)
3. `OP_IdxInsert` and `OP_IdxDelete` adjusted to handle special vector index type
4. Disable `xferOptimization` for vector indices (SQLite optimization which copy internal table/indices structures without explicit reindexing for queries like `INSERT INTO t SELECT * FROM q`)
5.  Add `libsql_vector_idx` no-op function which used as a marker in the vector index creation syntax `CREATE INDEX t_idx ON t (libsql_vector_idx(e))`
6. Implement `vector_top_k(idx, vector, k)` virtual table

## Testing

- `libsql_vector_index.test` TCL based test suite
- `benchtest` & `anntest`


libsql-sqlite3/src/vectorfloat32.c
```diff
@@ -198,6 +198,23 @@ float vectorF32DistanceCos(const Vector *v1, const Vector *v2){
...
+  float *e1 = v1->data;
+  float *e2 = v2->data;
+  int i;
+
+  assert( v1->dims == v2->dims );
+  assert( v1->type == VECTOR_TYPE_FLOAT32 );
+  assert( v2->type == VECTOR_TYPE_FLOAT32 );
+
+  for(i = 0; i < v1->dims; i++){
+    float d = e1[i]-e2[i];
```
haaawk, line 212: Would it lead to a better precision if we use double here and make sum a double as well and then cast the value to float only before returning?

libsql-sqlite3/src/vectorIndex.c
```diff
@@ -459,4 +462,498 @@ int parseVectorIdxParams(Parse *pParse, VectorIdxParams *pParams, int type, int
...
+      return -1;
+    }
+    return 0;
+  }
+  *pErrMsg = "unexpected vector column type";
+  return -1;
+}
+
+int initVectorIndexMetaTable(sqlite3* db) {
+  static const char *zSql = "CREATE TABLE IF NOT EXISTS " VECTOR_INDEX_GLOBAL_META_TABLE " ( name TEXT, metadata BLOB );";
```
haaawk, line 554: Should we make name the primary key and add `WITHOUT ROWID`?

haaawk (approved):
LGTM. I might do another pass tomorrow as there is quite a lot of code here. @penberg it would be good for you to have a look at this one as well. You probably have more knowledge around this stuff.

libsql-sqlite3/src/sqliteInt.h
```diff
@@ -2702,7 +2703,6 @@ struct UnpackedRecord {
   u8 eqSeen;          /* True if an equality comparison has been seen */
 };
 
-
```
penberg, line 2705: Diff noise.

libsql-sqlite3/src/build.c
```diff
@@ -830,7 +833,7 @@ static void SQLITE_NOINLINE deleteTable(sqlite3 *db, Table *pTable){
   for(pIndex = pTable->pIndex; pIndex; pIndex=pNext){
     pNext = pIndex->pNext;
     assert( pIndex->pSchema==pTable->pSchema
-         || (IsVirtual(pTable) && pIndex->idxType!=SQLITE_IDXTYPE_APPDEF) );
+         || (IsVirtual(pTable) && (pIndex->idxType&3)!=SQLITE_IDXTYPE_APPDEF) ); // '&3' is the LibSQL fix to treat VECTOR index as APPDEF
```
penberg, line 836: I am not sure I understand what this is all about.

In any case, please hide the "&3" thing in a helper function like `getIndexType(pIndex)` to keep it more manageable

libsql-sqlite3/src/vectorfloat32.c
sivukhin: Yes, maybe. But we need to change this everywhere in the file, I guess.

I will leave for now `float` version and maybe we can change it later

libsql-sqlite3/src/select.c
```diff
@@ -1512,6 +1512,7 @@ KeyInfo *sqlite3KeyInfoAlloc(sqlite3 *db, int N, int X){
     p->enc = ENC(db);
     p->db = db;
     p->nRef = 1;
+    p->zIndexName = NULL; // LibSQL patch: necessary fix for vector search to make it work
```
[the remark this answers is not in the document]
sivukhin: I removed these comments but added some motivation in `zIndexName` field comment

libsql-sqlite3/src/vectorIndex.c
sivukhin: Yes, thanks! New schema highlighted a bug: SQLite can reparse schema (for several reasons, for example validation that after `ALTER TABLE` schema still valid) - and we don't want to insert anything in this case

libsql-sqlite3/src/build.c
sivukhin: Added `IsAppDefIndex` helper

libsql-sqlite3/src/vector.c
```diff
@@ -117,6 +117,19 @@ float vectorDistanceCos(const Vector *pVector1, const Vector *pVector2){
...
+    case VECTOR_TYPE_FLOAT32:
+      return vectorF32DistanceL2(pVector1, pVector2);
+    case VECTOR_TYPE_FLOAT64:
+      return vectorF64DistanceL2(pVector1, pVector2);
+    default:
+      assert(0);
+  }
+  return 0;
+}
+
```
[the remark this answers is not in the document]
sivukhin: Yes, true. But it's quite innocent - so I decided to leave it here... :pleading_face:

libsql-sqlite3/src/insert.c
```diff
@@ -2692,9 +2692,32 @@ int sqlite3OpenTableAndIndices(
...
+       * we can't emit OP_OpenVectorIdx command for this operation;
+       *
+       * As vector index creates empty B-tree index - it's safe to issue
+       * OP_OpenRead command for it
+       *
+       * TODO: with current implementation, integrity_check will output error
+       * for vector index as rows will be missed in it
+       * It's better to remove this error in future - but for now it's unclear
+       * how to do that with minimal code changes
+       */
```
MarinPostma, line 2706: does that mean that we can't perform integrity checks if vectors are used?

sivukhin: You can, but they will report integrity violation warnings for vector indices (e.g. `{row 1 missing from index t_integrity_idx} {wrong # of entries in index t_integrity_idx}`)

Do we use them anywhere in the platform for some needs?

MarinPostma: We have used them in the past to restore corrupted database

PR #1582, opened 2024-07-22
Vector search respect db idx

sivukhin:
## Context

Current code always assume that we will work with `"main"` database which is merely true in real life. 

Not only explicit `ATTACH` will not work with current code, but also `VACUUM INTO 't.db'` will fail as internally it attaches temp DB with "vacuumed" state.

This PR carefully propagate `iDb` / `zDbSName` to vector API and work with provided schema instead of unconditionarl `"main"`.

One caveat is that table-valued functions can't be invoked for any schema other then `main` - but this is a restriction of SQLite (https://www.sqlite.org/vtab.html#eponymous_virtual_tables). I guess for now this is fine.

## Testing

- TCL tests are passing


penberg (approved):

sivukhin:
Rust tests are broken due to the expired JWT token. 
Lucio should fix this eventually.

PR #1584, opened 2024-07-22
segment index header

MarinPostma:
- **add LIBSQL_MAGIC number const**
- **update InvalidIndex error**
- **add SegmentIndexHeader**
- **store segment index header to s3**
- **verify segment header when downloading from s3**

### DIFF
--- a/libsql-wal/src/storage/backend/s3.rs
+++ b/libsql-wal/src/storage/backend/s3.rs
@@ -27,6 +27,7 @@ use crate::io::{FileExt, Io, StdIO};
 use crate::segment::compacted::CompactedSegmentDataHeader;
 use crate::segment::Frame;
 use crate::storage::{Error, RestoreOptions, Result};
+use crate::LIBSQL_MAGIC;
 
 pub struct S3Backend<IO> {
     client: Client,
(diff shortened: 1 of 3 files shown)


libsql-wal/src/storage/backend/s3.rs
```diff
@@ -153,14 +166,19 @@ impl<IO: Io> S3Backend<IO> {
...
-            .collect()
-            .await
-            .map_err(|e| Error::unhandled(e, ""))?
-            .to_vec();
-        let index = fst::Map::new(bytes).map_err(|_| Error::InvalidIndex)?;
+        let mut stream = self.s3_get(config, s3_index_key).await?.into_async_read();
+        let mut header: SegmentIndexHeader = SegmentIndexHeader::new_zeroed();
+        stream.read_exact(header.as_bytes_mut()).await?;
+        if header.magic.get() != LIBSQL_MAGIC && header.version.get() != 1 {
+            return Err(Error::InvalidIndex("index header magic or version invalid"));
```
LucioFranco, line 173: maybe down the line I feel like we should make these two separate errors so we know which one triggered this.

LucioFranco (approved):

libsql-wal/src/storage/backend/s3.rs
MarinPostma: i agree when we have more version we can do that, but right it only means that the segment is corrupted

PR #1590, opened 2024-07-22
use gosu instead of native docker USER in the entrypoint

sivukhin:
## Context

We are using `USER` command to run `sqld` process from `sqld` user but this is quite non-flexible. For example, if we will try to mount local user directory to the container it will fail with an error:

```sh
$> docker run --name some-sqld -it \
    -v $(pwd):/var/lib/sqld \
    -e SQLD_NODE=primary \
    ghcr.io/tursodatabase/libsql-server:latest /bin/sqld
...
Error: Permission denied (os error 13)
```

This PR changes the way we spawn `sqld` process: now we run `docker-wrapper.sh` with root permission which internally change permission of the data dir recursively and then spawn `docker-entrypoint.sh` from the `sqld` user with the [`gosu`](https://github.com/tianon/gosu) utility

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM rust:slim-bullseye AS chef
 RUN apt update \
     && apt install -y libclang-dev clang \
         build-essential tcl protobuf-compiler file \
-        libssl-dev pkg-config git tcl cmake \
+        libssl-dev pkg-config git cmake \
     && apt clean \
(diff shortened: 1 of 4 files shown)


penberg (approved):
Looks good to me, but let's get some more eyes on this.

haaawk (approved):

athoscouto (approved):

PR #1592, opened 2024-07-23
libsql: provide more return info for sync

LucioFranco:
This PR changes the return type of sync to a struct that provides the current commit fno and how many frames that sync call required. This should give more insight to users to understand how much data has been synced.

### DIFF
--- a/libsql-server/tests/embedded_replica/mod.rs
+++ b/libsql-server/tests/embedded_replica/mod.rs
@@ -759,7 +759,7 @@ fn replicate_with_snapshots() {
         assert_eq!(stat, 427);
 
         let rep = db.sync().await.unwrap();
-        assert_eq!(rep.frames_synced(), 427);
+        assert_eq!(rep.frames_synced(), 0);
 
         let conn = db.connect().unwrap();
 
@@ -1336,7 +1336,7 @@ fn replicated_return() {
 
         let rep = db.sync().await.unwrap();
         assert_eq!(rep.frame_no(), Some(10));
-        assert_eq!(rep.frames_synced(), 11);
+        assert_eq!(rep.frames_synced(), 9);
 
         // Regenerate log
         notify.notify_waiters();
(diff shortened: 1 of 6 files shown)


libsql-server/tests/embedded_replica/mod.rs
```diff
@@ -1226,3 +1226,134 @@ fn txn_bug_issue_1283() {
...
+        assert_eq!(rep.start_frame_no(), Some(1));
+
+        notify.notify_waiters();
+        notify.notified().await;
+
+        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+
+        let rep = db.sync().await.unwrap();
+        assert_eq!(rep.frame_no(), Some(4));
+        assert_eq!(rep.start_frame_no(), Some(4));
```
LucioFranco, line 1353: I would expect this to be 0 or 1 but it doesn't seem to re replicate, any ideas @MarinPostma

libsql/src/replication/mod.rs
```diff
@@ -229,7 +255,20 @@ impl EmbeddedReplicator {
             }
         }
 
-        Ok(replicator.client_mut().committed_frame_no())
+        let last_frames_synced = self.last_frames_synced.fetch_add(
+            replicator.frames_synced(),
+            std::sync::atomic::Ordering::Relaxed,
```
haaawk, line 260: I usually recommend using sequential ordering unless you know this is a bottleneck in any way.

haaawk (approved):
It is not helping in review that you're reformatting code not changed in the PR.

libsql/src/replication/mod.rs
LucioFranco: While I don't disagree, I think the fact that this isn't being ordered with anything else + being a fetch_add means that these orderings don't mean much in the end and thus relaxed is fine.

LucioFranco:
> It is not helping in review that you're reformatting code not changed in the PR.

I am not sure why my editor is formatting something but running `cargo fmt` isn't. I need to look into that since those changes should be enforced by CI.

PR #1595, opened 2024-07-23
set error to pParse and preserve internal invariant about possible error inside the condition

sivukhin:
## Context

`quick.test` failing with debug build due to broken invariants:

```
recoverfault-1-oom-persistent.1428... Ok
testfixture: sqlite3.c:127443: sqlite3KeyInfoOfIndex: Assertion `pParse->rc==SQLITE_ERROR_MISSING_COLLSEQ' failed.
[1]    180982 IOT instruction (core dumped)  ./testfixture test/quick.test
```

This is due to violation of sqlite3 invariant which was introduced in the PR #1571 

This PR align our patch with sqlite invariants and expectations

### DIFF
--- a/libsql-sqlite3/src/build.c
+++ b/libsql-sqlite3/src/build.c
@@ -5684,6 +5684,9 @@ out_nomem:
   if( zDbSName != NULL ){
     sqlite3DbFree(pParse->db, zDbSName);
   }
+  if( pKey != NULL ){
+    sqlite3KeyInfoUnref(pKey);
+  }
   return sqlite3OomFault(pParse->db);
 }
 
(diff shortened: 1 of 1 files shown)


libsql-sqlite3/src/build.c
```diff
@@ -5628,19 +5628,30 @@ KeyInfo *sqlite3KeyInfoOfIndex(Parse *pParse, Index *pIdx){
   int nCol = pIdx->nColumn;
   int nKey = pIdx->nKeyCol;
   KeyInfo *pKey;
+  char* zIndexName = NULL, *zDbSName = NULL;
```
haaawk, line 5631: Do we need those variables at all? Couldn't we just use pKey->zIndexName and pKey->zDbSName directly?

haaawk (approved):

PR #1596, opened 2024-07-24
Vector search adaptive block

sivukhin:
## Context

Now we hard coded 65536 as a default block size which is waste of space especially when we are working with relatively low-dimensional embeddings (like 64d). 

This PR introduces `max_edges=[int]` index parameter and also estimate reasonable default in case if parameter is omitted.

The logic behind current estimate is the following:
1. We don't want to have uncontrollable write amplification - so we should have hard upper bound for disk amplification due to neighbours redundancy in every node block. Current hard limit is `50x` redundancy compared to the node storage overhead (note, that with compression of neighbours this hard limit will automatically adjust and allow us to store more neighbours per node block)
2. The amount of neighbours should grows with the dimension of the embeddings space. I don't have references about why something like $D^{\frac{1}{C}}$ (where $D$ - space dimensionality) must be used, but I have hand-wavy arguments for myself like this: with $D^\frac{1}{C}$ neighbours for every vertex we will have $O(D)$ neighbours in $C$-radius of our graph and with $O(D)$ neighbours (if they are random enough) we have an ability to go into direction which will strictly improve our distance (generally, in $D$-dimensional metric space you need at least $D+1$ neighbours ([simplex](https://en.wikipedia.org/wiki/Simplex)) in order to have an improvement in any direction).
In my experiments $2 \sqrt{D} \ldots 3 \sqrt{D}$ shows pretty good `10-recall@10` values (90%+) - so I choose $3 \sqrt{D}$ as a second part of the `max_edges` estimation.

Overall, formula looks like this:

$max\\_edges = \min( 3 \sqrt{D}, 50 \frac{node\\_overhead(type, dims)}{edge\\_overhead(type, dims)} )$

## Changes

- Added `max_edges` parameter in the index creating syntax: 
```sql
CREATE INDEX t_idx ON t ( libsql_vector_idx(e, 'max_edges=32' ) )
```
- Estimate default value for `max_edges` with the formula from above
- Treat `block_size` binary parameter literally if it greater than 128, else - shift the value by `DISKANN_BLOCK_SIZE_SHIFT` to get real block size in bytes

### DIFF
--- a/libsql-sqlite3/src/vectorIndex.c
+++ b/libsql-sqlite3/src/vectorIndex.c
@@ -383,18 +383,19 @@ static struct VectorColumnType VECTOR_COLUMN_TYPES[] = {
 struct VectorParamName {
   const char *zName;
   int tag;
-  int type; // 0 - enum, 1 - integer, 2 - float
+  int type; // 0 - string enum, 1 - integer, 2 - float
   const char *zValueStr;
   u64 value;
 };
 
 static struct VectorParamName VECTOR_PARAM_NAMES[] = {
-  { "type",     VECTOR_INDEX_TYPE_PARAM_ID, 0, "diskann", VECTOR_INDEX_TYPE_DISKANN },
-  { "metric",   VECTOR_METRIC_TYPE_PARAM_ID, 0, "cosine", VECTOR_METRIC_TYPE_COS },
-  { "metric",   VECTOR_METRIC_TYPE_PARAM_ID, 0, "l2",     VECTOR_METRIC_TYPE_L2 },
-  { "alpha",    VECTOR_PRUNING_ALPHA_PARAM_ID, 2, 0, 0 },
-  { "search_l", VECTOR_SEARCH_L_PARAM_ID, 1, 0, 0 },
-  { "insert_l", VECTOR_INSERT_L_PARAM_ID, 2, 0, 0 },
+  { "type",      VECTOR_INDEX_TYPE_PARAM_ID,    0, "diskann", VECTOR_INDEX_TYPE_DISKANN },
+  { "metric",    VECTOR_METRIC_TYPE_PARAM_ID,   0, "cosine", VECTOR_METRIC_TYPE_COS },
+  { "metric",    VECTOR_METRIC_TYPE_PARAM_ID,   0, "l2",     VECTOR_METRIC_TYPE_L2 },
+  { "alpha",     VECTOR_PRUNING_ALPHA_PARAM_ID, 2, 0, 0 },
+  { "search_l",  VECTOR_SEARCH_L_PARAM_ID,      1, 0, 0 },
+  { "insert_l",  VECTOR_INSERT_L_PARAM_ID,      1, 0, 0 },
+  { "max_edges", VECTOR_MAX_EDGES_PARAM_ID,     1, 0, 0 },
 };
 
 static int parseVectorIdxParam(const char *zParam, VectorIdxParams *pParams, const char **pErrMsg) {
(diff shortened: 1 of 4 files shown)


libsql-sqlite3/src/vectorIndex.c
```diff
@@ -363,18 +363,19 @@ static struct VectorColumnType VECTOR_COLUMN_TYPES[] = {
...
-  { "alpha",    VECTOR_PRUNING_ALPHA_PARAM_ID, 2, 0, 0 },
-  { "search_l", VECTOR_SEARCH_L_PARAM_ID, 1, 0, 0 },
-  { "insert_l", VECTOR_INSERT_L_PARAM_ID, 2, 0, 0 },
+  { "type",      VECTOR_INDEX_TYPE_PARAM_ID,    0, "diskann", VECTOR_INDEX_TYPE_DISKANN },
+  { "metric",    VECTOR_METRIC_TYPE_PARAM_ID,   0, "cosine", VECTOR_METRIC_TYPE_COS },
+  { "metric",    VECTOR_METRIC_TYPE_PARAM_ID,   0, "l2",     VECTOR_METRIC_TYPE_L2 },
+  { "alpha",     VECTOR_PRUNING_ALPHA_PARAM_ID, 2, 0, 0 },
+  { "search_l",  VECTOR_SEARCH_L_PARAM_ID,      1, 0, 0 },
+  { "insert_l",  VECTOR_INSERT_L_PARAM_ID,      1, 0, 0 },
+  { "max_edges", VECTOR_MAX_EDGES_PARAM_ID,     1, 0, 0 },
```
penberg, line 378: `max_edges` seems a bit odd name for the parameter. Did you consider `max_neighbors`?

libsql-sqlite3/src/vectordiskann.c
```diff
@@ -64,6 +65,11 @@
 // stack simplify memory managment code and also doesn't impose very strict limits here since 128 bytes for column names should be enough for almost all use cases
 #define DISKANN_SQL_RENDER_LIMIT 128
 
+// limit to the maximum size of DiskANN block (128 MB)
+// even with 1MB we can store tens of thousands of nodes in several GBs - which is already too much
+// but we are "generous" here and allow user to store up to 128MB blobs
+#define DISKANN_MAX_BLOCK_SZ 134217728
```
penberg, line 71: I wonder if anything actually works with block size this big? I think 1 MB is probably a reasonable limit.

sivukhin: `128MB` is just pretty rough and loose upper bound. 
I don't want to limit blocks to 1MB because 1MB + eps (2MB for example) also can be usable. 
But `128MB` is just insane size for single node which no-one should actually use.

libsql-sqlite3/src/vectorIndex.c
sivukhin: Yes, agreed. The only reason why I choose `edge` here is because I decided to use this terminology in the `diskann` code (because node and edge is more concise "terms" compared to node and neighbor, from my subjective point of view).

But for users we don't need to preserve this internal terminology.

sivukhin: Hm....actually, should we choose something else?
I realized that there are two different spellings of the word: `neighbors` and `neighbours` which may be confusing...

sivukhin: We can shorten setting name down to `max_neigh`...

sivukhin: Or we can support both variants actually

PR #1369, opened 2024-05-02
doc: add authentication section + extra configuration

jeremybarbet:

PR #1602, opened 2024-07-25
doc: add authentication section + extra configuration

haaawk:


haaawk:
> Also found [this part](https://github.com/tursodatabase/libsql/pull/1369/commits/b106c025bd148410e698f90f43904e284a5c3608#diff-005fcade827e716a352a3d1d3c65f153687096adee7029e3d2924f037dd7efd4R137-R139) useful imo

I didn't want to include a link to old version of the code. At the same time there's no way to reliably point to selected lines in the file on main.

flexchar:
Hey @haaawk, thank you for this! 

I've been struggling several hours trying to generate JWT with both public key and token. Would you be as kind to provide a sample script how one could make it?


I tried many variations including 

```bash
ssh-keygen -t ed25519 -f jwt_ed25519_key -N ""
cat jwt_ed25519_key.pub | cut -d' ' -f2 | base64 -d | base64 -w 0 | tr '+/' '-_' | tr -d '=' > jwt_ed25519_key_base64.txt
# then manually encoding to url safe base64 and passing as ENV
```

I know security is hard, but is it really this difficult...




```ts
import jwt from 'jsonwebtoken';

const privateKey = await Bun.file('jwt_ed25519_key').text();

const payload = {
    sub: 'user@example.com', // subject
    name: 'John Doe', // your name
    iat: Math.floor(Date.now() / 1000), // issued at time
};

const token = jwt.sign(payload, privateKey, { algorithm: 'ES256' });

console.log('JWT:', token);
```

But it seems that there's always something that breaks.

flexchar:
It must be my lucky day since I came across a reply in Discord server by the [@darkterminal](https://github.com/darkterminal) 

> You can use this scripts: https://github.com/tursodatabase/libsql/tree/main/libsql-server/scripts
> 
> 1. Build your docker image (Make sure you includes the gen_jwt.py script)
> 2. Start a container from your image
> 3. Once inside the container, navigate to the directory containing the gen_jwt.py script and execute it with Python.
> 
> Or you can create Rest API to generate the `token` and `ro_token` that can hit inside or outside your container by your Rest API Endpoint.

Ref: https://discord.com/channels/933071162680958986/1259965600034193519/1259971864499978351


The python script got me going, finally! 

The only catch was to install correct python package, not `jwt` but `pip install PyJWT`.

haaawk:
@flexchar if you're using Turso you can obtain a token using `turso db tokens create <db name>`
Otherwise the python script you mentioned is good.

flexchar:
@haaawk how do you specify the host of your self-hosted `libsql` when you run `turso db tokens create <db name>`?

haaawk:
> @haaawk how do you specify the host of your self-hosted `libsql` when you run `turso db tokens create <db name>`?

That won't work for self-hosted

PR #1603, opened 2024-07-25
add more tests for vector feature

sivukhin:
## Context

Add simple "fuzz"-test which just execute multiple `INSERT/UPDATE/DELETE/SEARCH` operations against table with vector index + extend TCL test suite with couple more potentially tricky cases (rollback of the transaction and table with several vector columns)

fuzz-test is `ignored` by default but can be ran on demand with following command (on my laptop it runs within 15 minutes):
```sh
$> cargo test vector_fuzz_test -- --nocapture --include-ignored
...
running 1 test
============== ATTEMPT 0 (seed 5584570835305270879u64, dim 163, operations 1) ================
...
============== ATTEMPT 9999 (seed 11747876430315692925u64, dim 899, operations 73) ================
test vector_test ... ok
...
```

### DIFF
--- a/libsql/tests/integration_tests.rs
+++ b/libsql/tests/integration_tests.rs
@@ -1,14 +1,14 @@
 #![allow(deprecated)]
 
-use rand::prelude::*;
-use rand::distributions::Uniform;
-use std::collections::HashSet;
 use futures::{StreamExt, TryStreamExt};
 use libsql::{
     named_params, params,
     params::{IntoParams, IntoValue},
     Connection, Database, Value,
 };
+use rand::distributions::Uniform;
+use rand::prelude::*;
+use std::collections::HashSet;
 
 async fn setup() -> Connection {
     let db = Database::open(":memory:").unwrap();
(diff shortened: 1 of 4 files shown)


libsql/tests/integration_tests.rs
```diff
@@ -650,3 +653,102 @@ async fn deserialize_row() {
...
+        let _ = conn
+            .execute(
+                &format!(
+                    "CREATE TABLE users (id INTEGER PRIMARY KEY, v FLOAT32({}) )",
+                    dim
+                ),
+                (),
+            )
+            .await;
+        // println!("CREATE TABLE users (id INTEGER PRIMARY KEY, v FLOAT32({}) );", dim);
```
penberg, line 686: Why not make these printouts unconditional?

penberg (approved):
Nice!

libsql/tests/integration_tests.rs
sivukhin: They can be pretty large (I generate embeddings of dimensions up to 1536).

The idea is that you can run fuzz test and wait for exit. If failure happen you can uncomment these lines, copy seed to the source code and reproduce failure.

PR #1608, opened 2024-07-25
replicate from stored segments

MarinPostma:
This PR makes the necessary changes to perform replication from already stored segments.

Replication from a stored segment is performed when the requested frame number is not in the current log or the sealed segment list. We may then need to fetch those frames from durable storage.

After replicating all the frames we could from the current and sealed logs, we fetched stored segment indexes. If the index contains a frame we haven't seen already, we fetch the segment and replicate missing frames. We do that until we have replicated all the pages required until the requested frame_no.

As an optimization, we need to implement caching of segments and indexes to avoid fetching segments over and over again.


libsql-wal/src/replication/replicator.rs
```diff
@@ -76,26 +79,51 @@ impl<IO: Io> Replicator<IO> {
 
                     // replicate from tail
                     if should_replicate_from_tail {
-                        let (stream, replicated_until) = current.tail().stream_pages_from(self.next_frame_no, &mut seen).await;
-                        tokio::pin!(stream);
-                        let mut stream = stream.peekable();
+                        let replicated_until = {
+                            let (stream, replicated_until) = current
+                                .tail()
+                                .stream_pages_from( replicated_until, self.next_frame_no, &mut seen).await;
```
LucioFranco, line 85:
```suggestion
                                .stream_pages_from(replicated_until, self.next_frame_no, &mut seen).await;
```

libsql-wal/src/replication/replicator.rs
```diff
@@ -38,7 +40,7 @@ impl<IO: Io> Replicator<IO> {
     ///
     /// In a single replication step, the replicator guarantees that a minimal set of frames is
     /// sent to the replica.
-    pub fn frame_stream(&mut self) -> impl Stream<Item = Result<Frame>> + '_ {
+    pub fn frame_stream(&mut self) -> impl Stream<Item = Result<Box<Frame>>> + '_ {
```
LucioFranco, line 43: I would appreciate some comments in here explaining some of the logic

LucioFranco (approved):

PR #1611, opened 2024-07-26
Vector more functions

sivukhin:


sivukhin:
Close for now - but maybe go back to this later!

PR #1459, opened 2024-06-10
libsql: SQLite extension loading support

penberg:
This adds a load_extension() API for loading SQLite extensions, which is API compatible with Rusqlite.

Fixes #402

PR #1612, opened 2024-07-29
C bindings: Loading extensions

haaawk:
Add new `libsql_load_extension` function to C bindings to make it possible to load an extension on a connection.

This is needed for multiple things like allowing extensions in Go client or supporting CRSQLITE in OPSQLITE on mobile.

### DIFF
--- a/bindings/c/src/lib.rs
+++ b/bindings/c/src/lib.rs
@@ -6,6 +6,7 @@ extern crate lazy_static;
 mod types;
 
 use crate::types::libsql_config;
+use libsql::{errors, LoadExtensionGuard};
 use tokio::runtime::Runtime;
 use types::{
     blob, libsql_connection, libsql_connection_t, libsql_database, libsql_database_t, libsql_row,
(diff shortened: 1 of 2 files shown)


bindings/c/src/lib.rs
```diff
@@ -301,6 +302,55 @@ pub unsafe extern "C" fn libsql_connect(
...
+        }
+    };
+    let mut entry_point_option = None;
+    if !entry_point.is_null() {
+        let entry_point = unsafe { std::ffi::CStr::from_ptr(entry_point) };
+        entry_point_option = match entry_point.to_str() {
+            Ok(entry_point) => Some(entry_point),
+            Err(e) => {
+                set_err_msg(format!("Wrong entry point: {}", e), out_err_msg);
+                return 4;
```
sivukhin, line 331: Do these 2/4/5/6 error code has special meaning? Or it's just some numbers to distinguish between different error causes?

bindings/c/src/lib.rs
```diff
@@ -301,6 +302,55 @@ pub unsafe extern "C" fn libsql_connect(
...
+            }
+        };
+    }
+    if conn.is_null() {
+        set_err_msg("Null connection".to_string(), out_err_msg);
+        return 5;
+    }
+    let conn = conn.get_ref();
+    match RT.block_on(async move {
+        let _guard = LoadExtensionGuard::new(conn)?;
```
sivukhin, line 341: Do we need to use guard here or we better to expose enable_load_extension(onoff: bool) method instead?

BTW, `LoadExtensionGuard` ignore potential error which can happen on `drop` of the guard during disable of `enable_extension` API

bindings/c/src/lib.rs
haaawk: Just different values to make it easier to distinguish which code path caused the error.

bindings/c/src/lib.rs
haaawk: I guess this is more of a comment to the PR that introduced it (https://github.com/tursodatabase/libsql/pull/1459) as the problem you're describing is in Rust API. I'm using the Rust API in a recommended way here. You have a fair point but I don't want to fix the Rust API in this PR. Please open an issue and let's fix the Rust API in separate PR.

sivukhin (approved):

PR #1613, opened 2024-07-30
current segment recovery

MarinPostma:
The goal of this PR is to make the current segment recoverable in case of a crash. Taking inspiration from SQLite's wal, we introduce a grinning checksum to the current log, where each frame is hashed with the hash of the previous frame. If we find an unsealed segment on startup, we enter recovery mode for that segment, meaning that we iterate over its frames, recomputing the checksums. We stop whenever we hit a checksum that breaks the list invariant and patch the segment. This means that in NORMAL synchronous mode, we can lose committed data on the crash, but this is expected. We'll introduce a more robust sync mode later that flushes on every commit and has stronger guarantees.

This scheme introduces a bit more complexity:
- double flush on seal: on seal, we first flush all data to disk before updating the header and flushing again. This makes the SEALED flag on the segment reliable: if sealed is present, we can guarantee that all data was flushed before that, and therefore we can trust the segment's integrity. Without that, issues could arise on a partial flush where the header was flushed but not the rest of the segment.
- When we overwrite frames in a transaction, we need to recompute checksums.

I have added a test that simulates a crash during flush, where not all pages are flushed to disk, to demonstrate recovery.

### DIFF
--- a/libsql-wal/src/registry.rs
+++ b/libsql-wal/src/registry.rs
@@ -8,6 +8,7 @@ use libsql_sys::ffi::Sqlite3DbHeader;
 use parking_lot::{Condvar, Mutex};
 use tokio::sync::{mpsc, Notify, Semaphore};
 use tokio::task::JoinSet;
+use rand::Rng;
 use zerocopy::{AsBytes, FromZeroes};
 
 use crate::checkpointer::CheckpointMessage;
@@ -32,7 +33,7 @@ enum Slot<IO: Io> {
 
 /// Wal Registry maintains a set of shared Wal, and their respective set of files.
 pub struct WalRegistry<IO: Io, S> {
-    fs: IO,
+    io: IO,
     path: PathBuf,
     shutdown: AtomicBool,
     opened: DashMap<NamespaceName, Slot<IO>>,
@@ -59,7 +60,7 @@ impl<IO: Io, S> WalRegistry<IO, S> {
     ) -> Result<Self> {
         io.create_dir_all(&path)?;
         let registry = Self {
-            fs: io,
+            io,
             path,
             opened: Default::default(),
             shutdown: Default::default(),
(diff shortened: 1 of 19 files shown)


libsql-wal/src/registry.rs
```diff
@@ -126,9 +128,8 @@ where
                     update_durable(fno, notifier, durable_frame_no, namespace).await;
                 })
             });
-            self.storage
-                .store(&shared.namespace, sealed.clone(), None, cb);
-            new.tail().push(sealed);
+            new.tail().push(sealed.clone());
+            self.storage.store(&shared.namespace, sealed, None, cb);
```
LucioFranco, line 132: was there a bug here? Curious why these got switched

libsql-wal/src/segment/current.rs
```diff
@@ -58,6 +66,9 @@ impl<F> CurrentSegment<F> {
             index_size: 0.into(),
             header_cheksum: 0.into(),
             flags: 0.into(),
+            magic: LIBSQL_MAGIC.into(),
+            version: 1.into(),
```
LucioFranco, line 70: const?

libsql-wal/src/registry.rs
MarinPostma: Yes i explained in the commit message. I have a storage implementation that's so fast it tried to checkpoint before we added the sealed sediment to the list. Swapping the two so that the check pointer always finds a segment of it's notified

libsql-wal/src/segment/current.rs
MarinPostma: 🫡

libsql-wal/src/segment/sealed.rs
```diff
@@ -218,32 +222,79 @@ impl<F: FileExt> SealedSegment<F> {
...
+        let mut current_tx = Vec::new();
+        let mut last_committed = 0;
+        let mut size_after = 0;
+        let mut frame_count = 0;
+        for i in 0.. {
+            let offset = checked_frame_offset(i as u32);
+            match file.read_exact_at(frame.as_bytes_mut(), offset) {
+                Ok(_) => {
+                    let new_checksum = frame.frame.checksum(current_checksum);
+                    // this is the first checksum that don't match the checksum chain, drop the
```
LucioFranco, line 245:
```suggestion
                    // this is the first checksum that doesn't match the checksum chain, drop the
```

LucioFranco (approved):

PR #1616, opened 2024-07-31
type erase override config

MarinPostma:
Type erase confic type, making code simpler to read


LucioFranco (approved):
I very much dislike the Any stuff but at this point lets merge and deal with it later

PR #1623, opened 2024-08-01
Breaking change: Make Rows and Row API more consistent.

ignatz:
A few notes: I went the path of least resistance also assuming it would
    break fewer folks, i.e. make Row more like Rows and thus usize -> i32.
    
    Arguably, an unsigned type might be more appropriate both for length and
    indexes. I understand that the i32 stems from the sqlite bindings, which
    returns/accepts c_ints. Yet, Statement and Row made the jump to an
    unsigned, probably drawing the same conclusion, whereas Rows preserved
    its proximity to the c-bindings. This is probably an artifact?
    
    Also going on a limb, mapping c_int -> i32 is already a non-portable
    choice, with precision for c_int being platform dependent.

### DIFF
--- a/libsql/src/de.rs
+++ b/libsql/src/de.rs
@@ -68,7 +68,7 @@ impl<'de> Deserializer<'de> for RowDeserializer<'de> {
 
         visitor.visit_map(RowMapAccess {
             row: self.row,
-            idx: 0..self.row.inner.column_count(),
+            idx: 0..(self.row.inner.column_count() as usize),
             value: None,
         })
     }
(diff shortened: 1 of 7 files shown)


haaawk (approved):

ManevilleF:
@ignatz I don't understand the reason of removing `usize` in favor of `i32`, especially since it seems that the inner implementation is a `usize`  for `column_count`, cast as a `i32`.
Is it to be closer to the C impl ? Shouldn't we want a more rusty interface ?

ignatz:
I don't take a strong stance on i32 vs usize, as you already point out there are good arguments that can be made for either. My changes main goal was to have Rows::column_count and Row::column_count return the same type. As I said in the description, I felt that usize -> i32 was the path of less resistance

PR #1629, opened 2024-08-02
c: add replicated data for sync

LucioFranco:
This adds a new replicated struct that is output during sync for the C bindings.

### DIFF
--- a/bindings/c/src/lib.rs
+++ b/bindings/c/src/lib.rs
@@ -34,14 +34,32 @@ unsafe fn set_err_msg(msg: String, output: *mut *const std::ffi::c_char) {
 #[no_mangle]
 pub unsafe extern "C" fn libsql_sync(
     db: libsql_database_t,
+    out_err_msg: *mut *const std::ffi::c_char,
+) -> std::ffi::c_int {
+    let db = db.get_ref();
+    match RT.block_on(db.sync()) {
+        Ok(_) => 0,
+        Err(e) => {
+            set_err_msg(format!("Error syncing database: {e}"), out_err_msg);
+            1
+        }
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn libsql_sync2(
+    db: libsql_database_t,
     out_replicated: *mut replicated,
     out_err_msg: *mut *const std::ffi::c_char,
 ) -> std::ffi::c_int {
     let db = db.get_ref();
     match RT.block_on(db.sync()) {
         Ok(replicated) => {
-            (*out_replicated).frame_no = replicated.frame_no().unwrap_or(0) as usize;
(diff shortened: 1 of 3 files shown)


haaawk:
This is a breaking change. It will break go-libsql and op-sqlite that use sync api. Please prepare PRs to update those two users of the API.

bindings/c/src/types.rs
```diff
@@ -115,6 +115,12 @@ impl From<&mut libsql_connection> for libsql_connection_t {
     }
 }
 
+#[repr(C)]
+pub struct replicated {
+    pub frame_no: usize,
+    pub frames_synced: usize,
```
haaawk, line 121: Please use types from std::ffi like `std::ffi::c_int`

haaawk:
> This is a breaking change. It will break go-libsql and op-sqlite that use sync api. Please prepare PRs to update those two users of the API.

Also https://github.com/tursodatabase/libsql_android/

haaawk:
I would much prefer to add a new function that returns frames sync as a total always growing value and don't do breaking changes at this point.

bindings/c/src/lib.rs
```diff
@@ -46,6 +46,26 @@ pub unsafe extern "C" fn libsql_sync(
...
+#[no_mangle]
+pub unsafe extern "C" fn libsql_sync2(
+    db: libsql_database_t,
+    out_replicated: *mut replicated,
+    out_err_msg: *mut *const std::ffi::c_char,
+) -> std::ffi::c_int {
+    let db = db.get_ref();
+    match RT.block_on(db.sync()) {
+        Ok(replicated) => {
+            (*out_replicated).frame_no = replicated.frame_no().unwrap_or(0) as i32;
```
haaawk, line 58: Shouldn't we check if out_replicated is not null?

LucioFranco: Added that check thanks

bindings/c/src/lib.rs
```diff
@@ -46,6 +46,29 @@ pub unsafe extern "C" fn libsql_sync(
...
+#[no_mangle]
+pub unsafe extern "C" fn libsql_sync2(
+    db: libsql_database_t,
+    out_replicated: *mut replicated,
+    out_err_msg: *mut *const std::ffi::c_char,
+) -> std::ffi::c_int {
+    let db = db.get_ref();
+    match RT.block_on(db.sync()) {
+        Ok(replicated) => {
+            if out_replicated != std::ptr::null_mut() {
```
haaawk, line 58: Is this different than `!out_replicated.is_null()` ?

LucioFranco: oh no its not I forget that you can call that on a ptr, let me try that

PR #1632, opened 2024-08-03
Namespace configurators

MarinPostma:
In this PR, we rework how the namespace store instantiates namespaces to make a place for libsql wal replicas. To that end, we introduce the `NamespaceConfgurators,` which are responsible for instantiating replicas, primaries, or schema databases. The `NamespaceConfigurators` is set up during program initialization with whatever state is required to instantiate a particular namespace type. Depending on the configuration, the `NamespaceStore` will call the appropriate namespace configurator to load the namespace.

This is useful for libsql-wal since we must keep a handle to the `libsql-wal` registry to get an injector or a replicator instance.

To reduce the scope of this PR, the implementation of the replication protocol using libsql wal will be integrated later.


libsql-server/src/namespace/configurator/primary.rs
```diff
@@ -4,125 +4,192 @@ use std::{path::Path, pin::Pin, sync::Arc};
...
+        let connection_maker = Arc::new(connection_maker);
+
+        if meta_store_handle.get().shared_schema_name.is_some() {
+            let block_writes = block_writes.clone();
+            let conn = connection_maker.create().await?;
+            tokio::task::spawn_blocking(move || {
+                conn.with_raw(|conn| -> crate::Result<()> {
+                    setup_migration_table(conn)?;
+                    if has_pending_migration_task(conn)? {
+                        block_writes.store(true, Ordering::SeqCst);
```
LucioFranco, line 86: if this fails do we just keep it blocked forever? Is that going to be a problem?

libsql-server/src/namespace/configurator/mod.rs
```diff
@@ -73,13 +99,30 @@ impl NamespaceConfigurators {
...
-    ) -> Pin<Box<dyn Future<Output = crate::Result<super::Namespace>> + Send + 'a>>;
+    ) -> Pin<Box<dyn Future<Output = crate::Result<Namespace>> + Send + 'a>>;
+
+    fn cleanup<'a>(
+        &'a self,
+        namespace: &'a NamespaceName,
+        db_config: &'a DatabaseConfig,
+        prune_all: bool,
+        bottomless_db_id_init: NamespaceBottomlessDbIdInit,
+    ) -> Pin<Box<dyn Future<Output = crate::Result<()>> + Send + 'a>>;
```
LucioFranco, line 129: why not use `async_trait`?

libsql-server/src/http/admin/stats.rs
```diff
@@ -140,10 +140,12 @@ pub(super) async fn handle_stats<C>(
     State(app_state): State<Arc<AppState<C>>>,
     Path(namespace): Path<String>,
 ) -> crate::Result<Json<StatsResponse>> {
+    dbg!();
```
LucioFranco: I am reviewing commit by commit, so not sure if these got removed but wanted to note it.

libsql-server/src/lib.rs
```diff
@@ -623,16 +581,19 @@ where
                 .configure(&mut join_set);
             }
             DatabaseKind::Replica => {
+                dbg!();
```
LucioFranco: more here

libsql-server/src/lib.rs
```diff
@@ -28,6 +28,7 @@ use auth::Auth;
 use config::{
     AdminApiConfig, DbConfig, HeartbeatConfig, RpcClientConfig, RpcServerConfig, UserApiConfig,
 };
+use futures::future::{pending, ready};
```
LucioFranco, line 31: btw these exist now in std https://doc.rust-lang.org/stable/std/future/index.html

LucioFranco (approved):

libsql-server/src/namespace/configurator/mod.rs
MarinPostma: no real reason, except that async_trait doesn't play nicely with LSP, and it expends to something quite trivial...

libsql-server/src/namespace/configurator/primary.rs
MarinPostma: this is not new, just moved over here. if that fails, the namespace won't be loaded, so we assume that there will be a retry. I that keeps failing forever then yes, but that's another problem (like out of disk or some stuff like that)

libsql-server/src/lib.rs
MarinPostma: ah right, for some reason RA imports those by default

PR #1640, opened 2024-08-06
Explicitly disable SEH on Windows

penberg:
We don't support SEH on Windows because of WAL virtualization so let's just disable it in the source code. This comes after spending 30 minutes of quality time with our build system and CMake in an attempt to disable it externally...

### DIFF
--- a/libsql-sqlite3/src/sqliteInt.h
+++ b/libsql-sqlite3/src/sqliteInt.h
@@ -324,15 +324,17 @@
 #  endif
 #endif
 
-/*
-** Enable SQLITE_USE_SEH by default on MSVC builds.  Only omit
-** SEH support if the -DSQLITE_OMIT_SEH option is given.
-*/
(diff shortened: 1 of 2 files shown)


libsql-sqlite3/src/sqliteInt.h
```diff
@@ -324,15 +324,17 @@
 #  endif
 #endif
 
-/*
-** Enable SQLITE_USE_SEH by default on MSVC builds.  Only omit
-** SEH support if the -DSQLITE_OMIT_SEH option is given.
-*/
-#if defined(_MSC_VER) && !defined(SQLITE_OMIT_SEH)
```
sivukhin, line 331: Why can't we set `-DSQLITE_OMIT_SEH` in the build.rs?

haaawk: It does not help. I tried that on libsql_js branch

haaawk (approved):

PR #1644, opened 2024-08-07
cancel query when request is dropped

MarinPostma:
This PR implements cancelling a query when the underlying request is dropped. This is done by adding a progress handler to the connection that checks every 100 instruction if the request has been dropped.

### DIFF
--- a/libsql-server/src/connection/libsql.rs
+++ b/libsql-server/src/connection/libsql.rs
@@ -15,7 +15,9 @@ use tokio::sync::watch;
 use tokio::time::{Duration, Instant};
 
 use crate::error::Error;
-use crate::metrics::{DESCRIBE_COUNT, PROGRAM_EXEC_COUNT, VACUUM_COUNT, WAL_CHECKPOINT_COUNT};
+use crate::metrics::{
+    DESCRIBE_COUNT, PROGRAM_EXEC_COUNT, QUERY_CANCELED, VACUUM_COUNT, WAL_CHECKPOINT_COUNT,
+};
 use crate::namespace::broadcasters::BroadcasterHandle;
 use crate::namespace::meta_store::MetaStoreHandle;
 use crate::namespace::ResolveNamespacePathFn;
@@ -512,6 +514,7 @@ impl<W: Wal> Connection<W> {
             Some(move || {
                 let canceled = canceled.load(Ordering::Relaxed);
                 if canceled {
+                    QUERY_CANCELED.increment(1);
                     tracing::trace!("request canceled");
                 }
                 canceled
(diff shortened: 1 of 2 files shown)


libsql-server/src/connection/libsql.rs
```diff
@@ -391,14 +391,43 @@ where
...
     ) -> Result<(B, Program)> {
+        struct Bomb {
+            canceled: Arc<AtomicBool>,
+            defused: bool,
+        }
+
+        impl Drop for Bomb {
+            fn drop(&mut self) {
+                if !self.defused {
+                    tracing::debug!("cancelling request");
```
LucioFranco, line 402: can we improve botht he debug logs so they are unique and easier to follow. I think seeing a bunch of these in the logs when turning on debug will be annoying so mayb we consider going to trace? maybe for atleast one of them?

MarinPostma: Hopefully not too many request should get cancelled, but no worries dowgrading to trace

haaawk (approved):

libsql-server/src/connection/libsql.rs
```diff
@@ -391,14 +393,43 @@ where
...
+        };
+
+        let mut bomb = Bomb {
+            canceled,
+            defused: false,
+        };
+
         PROGRAM_EXEC_COUNT.increment(1);
 
         check_program_auth(&ctx, &pgm, &self.inner.lock().config_store.get())?;
```
sivukhin, line 423: Will cancellation be triggered if auth check will fail?

sivukhin: Maybe better to put `bomb` after this check right before exec?

MarinPostma: The flag will be raised yes, but but won't execute the program so this won't do anything. What do you think could be an issue? Actually we probably never need to defuse the bomb, since we always re-initialize it.

MarinPostma: The only issue with this implementation is if 2 execute are performed concurrently on the same connection, but that isn't possible with hrana

sivukhin: IIUC, in this case log `tracing::trace!("cancelling request");` will be produced which may be a bit misleading

MarinPostma: fair

libsql-server/src/connection/libsql.rs
```diff
@@ -391,14 +391,43 @@ where
         ctx: RequestContext,
         builder: B,
     ) -> Result<(B, Program)> {
+        struct Bomb {
```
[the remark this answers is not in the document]
MarinPostma: this term is perfect 😄

PR #1652, opened 2024-08-08
libsql wal replicator

MarinPostma:
In this PR, we abstract the injector logic from libqsl-replicator into the Injector trait. The former `Injector` now becomes `SqliteInjector` and we introduce a new `LibsqlInjector` that uses libsql wal to perform injection. We also extend the replication protocol handshake so that clients can request a prefered wal version.


libsql-replication/src/injector/libsql_injector.rs
```diff
@@ -0,0 +1,45 @@
...
+
+pub struct LibsqlInjector {
+    injector: Injector<StdIO>,
+}
+
+impl super::Injector for LibsqlInjector {
+    async fn inject_frame(&mut self, frame: RpcFrame) -> Result<Option<FrameNo>> {
+        // this is a bit annoying be we want to read the frame, and it has to be aligned, so we
+        // must copy it...
+        // FIXME: optimize this.
```
LucioFranco, line 21: are we tracking these anywhere?

libsql-wal/src/shared_wal.rs
```diff
@@ -20,7 +20,7 @@ use libsql_sys::name::NamespaceName;
 
 #[derive(Default)]
 pub struct WalLock {
-    pub(crate) tx_id: Arc<Mutex<Option<u64>>>,
+    pub(crate) tx_id: Arc<async_lock::Mutex<Option<u64>>>,
```
LucioFranco, line 23: why async lock over the tokio one? Is it for the blocking stuff?

LucioFranco (approved):
left a few questions but otherwise lgtm

libsql-wal/src/shared_wal.rs
MarinPostma: eventually, I would like to make Warin runtime agnostic, so I try to avoid tokio primitives outside of the IO trait

libsql-replication/src/injector/libsql_injector.rs
MarinPostma: not really, I'll `grep FIXME` when I find time

PR #1656, opened 2024-08-09
 Ensure that blobs match the length of arrays they're being deserialized into and otherwise fail.

ignatz:


ignatz:
Friendly ping, this feels like a straight forward correctness improvement

ignatz:
friendly ping

PR #1663, opened 2024-08-12
libsql: add `tls` feature

LucioFranco:
This adds a new `tls` feature that is enabled by default, if this feature is disabled building a libsql connection will panic with a message asking you to configure a http connector. This allows users to bring their own http connector and more importantly their own TLS lib with their own versions without needing to compile rustls which we use by default.

This resolves solana-sdk >2 build issues with uses an older version of `curve25519-dalek` that pings `zeroize` to `<1.4`. New versions of `rustls` require `1.7` of `zeroize` thus causing issues when building `rustls` for libsql with the `tls` feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6251,7 +6251,6 @@ dependencies = [
  "percent-encoding",
  "pin-project",
  "prost",
- "rustls-native-certs 0.7.1",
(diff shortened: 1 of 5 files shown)


LucioFranco:
I believe this is a breaking change but it should affect very little users only those that disable default features.

PR #1667, opened 2024-08-14
libsql-wal final integration

MarinPostma:
This PR wires most of libsql-wal related stuff into sqld, and fixes uncovered bugs along the way.

*what works*:
- primary, replicas and schema using the new wal. Replication only works with storage enabled
- s3 storage
- embedded replicas (untested yet)

The new wal exists side by side with the legacy wal. To enable it, pass the `--use-custom-wal libsql-wal` arg.
Bottomless configuration remains unchanged.

*what doesn't work yet*:
- forking
- some quirks with the replication index (will fix as a followup)
- replica checkpointing
- replication without bottomless enabled
- automatic restore


libsql-server/src/connection/connection_core.rs
```diff
@@ -0,0 +1,809 @@
...
+            (config, stats, block_writes, resolve_attach_path)
+        };
+
+        builder.init(&this.lock().builder_config)?;
+        let mut vm = Vm::new(
+            builder,
+            &pgm,
+            move |stmt_kind| {
+                let should_block = match stmt_kind {
+                    StmtKind::Read | StmtKind::TxnBegin => config.block_reads,
```
LucioFranco, line 221: hmmm we don't check for block_reads here like we do for writes?

libsql-server/src/database/mod.rs
```diff
@@ -184,7 +184,7 @@ impl Database {
         match self {
             Database::Primary(p) => Some(p.wal_wrapper.wrapper().logger()),
             Database::Replica(_) => None,
-            Database::Schema(s) => Some(s.wal_wrapper.wrapper().logger()),
+            Database::Schema(s) => Some(s.wal_wrapper.as_ref().unwrap().wrapper().logger()),
```
LucioFranco, line 259: is this unwrap dangerous? Maybe we should make this an except?

libsql-server/src/connection/connection_core.rs
MarinPostma: No idea, I just copied the code over here no additional changes

libsql-server/src/database/mod.rs
MarinPostma: It's not dangerous because we never take it out, we could even make it type safe by adding a generic additional extra-data field set to the wal wrapper

We can add an expect no problem though

libsql-server/src/connection/libsql.rs
```diff
@@ -1,760 +1,107 @@
...
-                    block_writes,
-                    resolve_attach_path,
-                )?;
+                    inner.current_frame_no_receiver.clone(),
+                    inner.block_writes.clone(),
+                    inner.resolve_attach_path.clone(),
+                )
+            })
+            .await
+            .unwrap()?;
```
LucioFranco, line 76: is this unwrap dangerous?

libsql-server/src/lib.rs
```diff
@@ -198,17 +198,86 @@ struct Services<A, P, S, C> {
...
+
+    /// run the passed future until shutdown is called, then call the passed teardown future
+    #[track_caller]
+    pub fn spawn_until_shutdown_with_teardown<F, T>(&mut self, fut: F, teardown: T)
+    where
+        F: Future<Output = anyhow::Result<()>> + Send + 'static,
+        T: Future<Output = anyhow::Result<()>> + Send + 'static,
+    {
+        let shutdown = self.shutdown.clone();
+        self.join_set.spawn(async move {
```
LucioFranco, line 245: should we pass the current span via instrumenting the spawned future?

libsql-server/src/lib.rs
```diff
@@ -788,25 +803,74 @@ where
...
+                max_in_flight_jobs: 16,
+            };
+            let (storage, storage_loop) = AsyncStorage::new(config).await;
+
+            task_manager.spawn_with_shutdown_notify(|_| async move {
+                storage_loop.run().await;
+                Ok(())
+            });
+
+            Either::A(storage)
```
LucioFranco, line 857: I think we want to configure it this way so that it doesn't double load the config. I need to look deeper if this code actually does what bottomless did before, but in [this](https://github.com/tursodatabase/libsql/blob/main/bottomless/src/replicator.rs#L127) code I don't use `load_defaults`.

libsql-wal/src/segment/list.rs
```diff
@@ -337,6 +334,46 @@ impl<T> List<T> {
...
 }
 
+mod send_fst_ops {
+    use std::ops::{Deref, DerefMut};
+    use std::sync::Arc;
+
+    use fst::map::{OpBuilder, Union};
+
+    /// Safety: Union contains a Box<dyn trait> that doesn't require Send, to it's not send.
+    /// That's an issue for us, but all the indexes we have are safe to send, so we're good.
```
LucioFranco, line 344: why are they safe? Can the compiler assume something here and create a ub? Would be good to expand this the logic here seems a bit weak.

libsql-wal/src/storage/mod.rs
```diff
@@ -181,6 +181,9 @@ pub trait Storage: Send + Sync + 'static {
     ) -> Result<Map<Arc<[u8]>>>;
 
     async fn fetch_segment_data(
+    fn shutdown(&self) -> impl Future<Output = ()> + Send {
+        async { () }
+    }
```
LucioFranco, line 191: I am gonna assume this is just a weird diff

LucioFranco (approved):

MarinPostma:
thanks @LucioFranco will address all the comments in followups

PR #1668, opened 2024-08-15
replace parking_lot with tokio::sync::Mutex in some places

sivukhin:
## Context

Current `sqld` code abuses tokio runtime which can lead to complete deadlock of the server.

One issue which got triggered pretty frequently relying on the following facts about `sqld` internals:
1. Many **async** request handlers in `sqld` checks that metadata has namespace in it with sync lock `metadata.inner.configs` (see `MetaStore::exists` method)
2. Metastore remove operation take `metadata.inner.configs` lock first and then try to take `inner.conn` lock
3. Background checkpoint operation which run on blocking thread take `inner.conn` lock for the time of the checkpoint process

So, if (2) took `metadata.inner.configs` lock while (3) is in process, then all request **tasks** which hits (1) will be **blocked** on the **parking_lot** mutex until (3) released the lock and (2) finished. This can easily block all tokio worker threads and lead to complete server deadlock.

This PR mitigate this specific scenario by introducing 2 fixes:
1. `MetaStoreInner.configs` and `MetaStoreInner.conn` now uses `tokio::sync::Mutex` instead of `parking_lot::Mutex`. This will prevent scenario from above as **async** tasks now will be blocked on **async** lock and runtime will be able to switch them from worker threads and put some other workload on them
2. Metastore remove operation now take `inner.conn` lock first and then take `inner.configs` lock. This will prevent the cases where `inner.configs` lock is taken for too long by remove operation while awaiting next `inner.conn` lock. As we are using `configs` lock in both async & sync context and also this lock required for quick checks in metastore - it's better to not hold it for too long.

### DIFF
--- a/libsql-server/src/connection/libsql.rs
+++ b/libsql-server/src/connection/libsql.rs
@@ -415,7 +415,8 @@ where
 
         PROGRAM_EXEC_COUNT.increment(1);
 
-        check_program_auth(&ctx, &pgm, &self.inner.lock().config_store.get())?;
+        let config = self.inner.lock().config_store.get();
+        check_program_auth(&ctx, &pgm, &config).await?;
 
         // create the bomb right before spawning the blocking task.
         let mut bomb = Bomb {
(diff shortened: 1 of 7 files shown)


haaawk:
Why is this a draft @sivukhin ?

haaawk (approved):

penberg (approved):

PR #1669, opened 2024-08-15
expose basic tokio runtime metrics

sivukhin:
## Context

Expose basic tokio runtime metrics into the Prometheus:
```
tokio_runtime_blocking_queue_depth 0
tokio_runtime_injection_queue_depth 3
tokio_runtime_io_driver_fd_deregistered_count 5149
tokio_runtime_io_driver_fd_registered_count 5364
tokio_runtime_io_driver_ready_count 36342
tokio_runtime_num_blocking_threads 41
tokio_runtime_num_idle_blocking_threads 41
tokio_runtime_num_workers 8
tokio_runtime_remote_schedule_count 53401
```

### DIFF
--- a/libsql-server/src/http/admin/mod.rs
+++ b/libsql-server/src/http/admin/mod.rs
@@ -113,12 +113,7 @@ where
                     .absolute(metrics.io_driver_ready_count() as u64);
                 crate::metrics::TOKIO_RUNTIME_REMOTE_SCHEDULE_COUNT
                     .absolute(metrics.remote_schedule_count() as u64);
-                tokio::time::sleep(std::time::Duration::from_secs(1)).await;
-            }
-        });
 
-        tokio::task::spawn(async move {
-            loop {
                 crate::metrics::SERVER_COUNT.set(1.0);
                 tokio::time::sleep(std::time::Duration::from_secs(1)).await;
             }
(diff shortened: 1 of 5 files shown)


athoscouto (commented):
looks good, but I'll defer the approval to @MarinPostma et al

libsql-server/src/http/admin/mod.rs
```diff
@@ -93,9 +93,34 @@ where
...
+                    .absolute(metrics.io_driver_fd_registered_count() as u64);
+                crate::metrics::TOKIO_RUNTIME_IO_DRIVER_READY_COUNT
+                    .absolute(metrics.io_driver_ready_count() as u64);
+                crate::metrics::TOKIO_RUNTIME_REMOTE_SCHEDULE_COUNT
+                    .absolute(metrics.remote_schedule_count() as u64);
                 tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+            }
+        });
 
+        tokio::task::spawn(async move {
```
MarinPostma, line 120: Why do we keep setting this counter, and why do we need to do it in a separate task?

MarinPostma (approved):

libsql-server/src/http/admin/mod.rs
sivukhin: Ah, sure. I removed second task but preserve metric just to not break any dashboard panels which can rely on it

PR #1671, opened 2024-08-16
feat(libsql-server): send correct sse response

notrab:
The response now sends the proper expected event stream

```
:keep-alive

event:changes
data:{"insert":1}

:keep-alive

:keep-alive
```

### DIFF
--- a/libsql-server/src/http/user/listen.rs
+++ b/libsql-server/src/http/user/listen.rs
@@ -6,13 +6,19 @@ use crate::{
     namespace::{NamespaceName, NamespaceStore},
 };
 use axum::extract::State as AxumState;
-use axum::http::header::{CACHE_CONTROL, CONTENT_TYPE};
-use axum::http::{HeaderValue, Uri};
-use axum::response::{IntoResponse, Redirect, Response};
-use axum_extra::{extract::Query, json_lines::JsonLines};
+use axum::http::Uri;
+use axum::response::{
+    sse::{Event, Sse},
+    IntoResponse, Redirect,
+};
+use axum_extra::extract::Query;
 use futures::{Stream, StreamExt};
 use hyper::HeaderMap;
 use serde::{Deserialize, Serialize};
+use std::boxed::Box;
+use std::convert::Infallible;
+use std::pin::Pin;
+use std::time::Duration;
 use tokio_stream::wrappers::errors::BroadcastStreamRecvError;
 
 use super::db_factory::namespace_from_headers;
@@ -33,16 +39,13 @@ pub struct ListenQuery {
     action: Option<Vec<Action>>,
 }
 
-const EVENT_STREAM: HeaderValue = HeaderValue::from_static("text/event-stream");
-const NO_CACHE: HeaderValue = HeaderValue::from_static("no-cache");
-
 pub(super) async fn handle_listen(
     auth: Authenticated,
     AxumState(state): AxumState<AppState>,
     headers: HeaderMap,
     uri: Uri,
     query: Query<ListenQuery>,
-) -> crate::Result<Response> {
+) -> crate::Result<impl IntoResponse> {
     let namespace = namespace_from_headers(
         &headers,
         state.disable_default_namespace,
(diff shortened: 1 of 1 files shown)


ccadenac:
Tested locally and events are being sent out in the correct format as shown in the attached Postman screenshot.
sandbox

libsql-server/src/http/user/listen.rs
```diff
@@ -6,18 +6,30 @@ use crate::{
...
+use std::boxed::Box;
+use std::convert::Infallible;
+use std::pin::Pin;
+use std::time::Duration;
 use tokio_stream::wrappers::errors::BroadcastStreamRecvError;
 
 use super::db_factory::namespace_from_headers;
 use super::AppState;
 
+const LAGGED_MSG: &str = "some changes were lost";
```
LucioFranco, line 27: I wonder if there is a way we can encode this error better so a user doesn't have to string match on a lagged event vs other types of errors.

LucioFranco (approved):

libsql-server/src/http/user/listen.rs
```diff
@@ -103,42 +138,57 @@ async fn listen_stream(
...
+                Ok(msg) if filter_actions(&msg, &actions) => {
                     LISTEN_EVENTS_SENT.increment(1);
                     yield AggregatorEvent::Changes(msg);
                 },
                 Err(BroadcastStreamRecvError::Lagged(n)) => {
                     LISTEN_EVENTS_DROPPED.increment(n as u64);
-                    yield AggregatorEvent::Error(&LAGGED_MSG);
+                    yield AggregatorEvent::Error(LAGGED_MSG);
                 },
+                _ => {}
```
notrab, line 154:
```suggestion
                Err(e) => {
                    yield Err(Error::Internal(format!("Unexpected error in listen stream: {}", e)))?;
                },
```

@athoscouto would this work?

libsql-server/src/http/user/listen.rs
```diff
@@ -53,47 +84,51 @@ pub(super) async fn handle_listen(
...
     table: String,
-}
-
-impl Drop for Subscription {
-    fn drop(&mut self) {
-        self.store.unsubscribe(self.namespace.clone(), &self.table);
-    }
+    actions: Option<Vec<Action>>,
+) -> SseStream {
+    Box::pin(
```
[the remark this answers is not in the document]
notrab: It seemed quite common and used in other examples I found, and I think this used in the SSE examples is due to using trait objects. Not sure on the performance implications though.

libsql-server/src/http/user/listen.rs
notrab: I'm guessing we could do something like this:

```rs
#[derive(Debug, Serialize)]
#[serde(tag = "code", content = "message")]
enum ErrorEvent {
    Lagged(String),
    Other(String),
}
```

and then use that. But, this part of the code I didn't change, I just moved the position of it.

libsql-server/src/http/user/listen.rs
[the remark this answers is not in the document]
MarinPostma: pining is simply a marker for the compiler, so it has no performance implication. A future needs to be pinned in order to be polled (same goes for streams). Long story short, you don't really have a choice

libsql-server/src/http/user/listen.rs
MarinPostma: let's log them as well

MarinPostma (approved):

libsql-server/src/http/user/listen.rs
MarinPostma: Maybe just warn them

notrab: I pushed an update, let me know if that works.

libsql-server/src/http/user/listen.rs
```diff
@@ -149,9 +151,29 @@ async fn listen_stream(
...
+                            "Unexpected error in listen stream"
+                    );
+                },
+                _ => {
+                    debug!(
+                        namespace = %namespace,
+                        table = %table,
+                        "Filtered out message in listen stream"
+                    );
+                }
```
MarinPostma, line 167: this branch can never be reached because the other branches already cover all the patterns

notrab: I thought `Err(BroadcastStreamRecvError::Lagged(n))` was only invoked when that lagged msg matched? Everything else fallsthrough to `_`?

athoscouto: I suggest reverting these changes as a whole. Before the compiler would let us know in case any new errors (that could be added by changes somewhere else) went unhandled. In this way, we don't have this guarantee. 

Also, warning when events are dropped may make the drop problem worse, since one reason we may drop events is that the time it takes to serialize them is higher than the time we take to generate them.

My understanding is that they add no functional changes, but let me know if I've missed something.

athoscouto (approved):

PR #1678, opened 2024-08-19
vector search: hide vector index for sqlite planner

sivukhin:
## Context

PR https://github.com/tursodatabase/libsql/pull/1594 tried to solve issue with sqlite planner using vector index for some optimizations (for example in `SELECT COUNT(*) FROM t` queries).

Turned out hacks to avoid usage of vector indices in planner were not enough. Actually, `bUnordered` field set by `sqlite3` during index analysis process and manipulating with it can be dangerous and error-prone (as sqlite3 can reset it at any moment).

So, this PR fixes this more robustly:
1. `opMask` is set to 0 for vector index in `whereLoopAddBtreeIndex` which must additionally prevent sqlite to use index in any operations (`>`, `>=`, `=`, `IN`) (actually, it shouldn't happen because index has expression - but still - this is another guard)
2. `whereIsCoveringIndex` return false for vector index
3. Loop over indices in query planner skips vector indices completely
4. All places where optimizer prevented to use `bUnordered` indices were extended to have additional check for `idxIsVector` field

### DIFF
--- a/libsql-sqlite3/test/libsql_vector_index.test
+++ b/libsql-sqlite3/test/libsql_vector_index.test
@@ -225,6 +225,20 @@ do_execsql_test vector-index-dont-affect-sql {
   SELECT rowid FROM t_vector_other_sql WHERE emb = vector('[7,8]');
 } {4 1 1 2 3 4}
 
+do_execsql_test vector-partial-index-dont-affect-sql {
+  CREATE TABLE t_vector_other_sql_part ( emb FLOAT32(2), t TEXT );
+  INSERT INTO t_vector_other_sql_part VALUES (vector('[1,2]'), 'paper'), (vector('[3,4]'), 'journal');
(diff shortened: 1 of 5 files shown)


libsql-sqlite3/test/libsql_vector_index.test
```diff
@@ -307,6 +321,8 @@ do_execsql_test vector-f64-index {
   SELECT * FROM vector_top_k('t_f64_idx', vector64('[1,2]'), 2);
 } {1 2}
 
+reset_db
```
sivukhin, line 324: We currently have some issue and test without db reset starting to generate generic sqlite error in some unexpected places.

This is really weird but I don't understand why it happens and how we can fix it.
Last time I investigated this issue I noticed that `SELECT` over `libsql_vector_meta_shadow` table with condition over `name` field sometimes returns nothing but if you will force full scan of the table - then everything will be fine.

I suspect that issue arise when split of `libsql_vector_meta_shadow` B-Tree happens. But still not sure why we have this issue...

CC @penberg

penberg (approved):

PR #1683, opened 2024-08-20
disable foreign keys in schema migration

MarinPostma:
This PR works around disabling foreign_keys being a noop in schema migrations. During the validation of the migration script we allows a prologue and an epilogue of pragmas. If such a prologue is present, we assume that foreign_keys must be disabled for the migration. This is a bit hacky since we don't really validate what pragma is actually being used, but disabling foreign_keys is such a common pattern that assuming it should be ok

### DIFF
--- a/libsql-server/src/database/schema.rs
+++ b/libsql-server/src/database/schema.rs
@@ -51,12 +51,15 @@ impl<C: crate::connection::Connection> crate::connection::Connection for SchemaC
         } else {
             check_program_auth(&ctx, &migration, &self.config.get()).await?;
             let connection = self.connection.clone();
-            validate_migration(&mut migration)?;
+            let disable_foreign_key = validate_migration(&mut migration)?;
             let migration = Arc::new(migration);
             let builder = tokio::task::spawn_blocking({
                 let migration = migration.clone();
                 move || {
                     let res = connection.with_raw(|conn| -> crate::Result<_> {
+                        if disable_foreign_key {
+                            conn.execute("PRAGMA foreign_keys=off", ())?;
+                        }
                         let mut txn = conn
                             .transaction_with_behavior(rusqlite::TransactionBehavior::Immediate)
                             .map_err(|_| {
@@ -73,6 +76,9 @@ impl<C: crate::connection::Connection> crate::connection::Connection for SchemaC
                             &QueryBuilderConfig::default(),
                         );
                         txn.rollback().unwrap();
+                        if disable_foreign_key {
+                            conn.execute("PRAGMA foreign_keys=on", ())?;
+                        }
                         Ok(ret?)
                     });
 
--- a/libsql-server/src/schema/db.rs
+++ b/libsql-server/src/schema/db.rs
@@ -9,6 +9,7 @@ use crate::namespace::NamespaceName;
 use crate::schema::status::{MigrationJobProgress, MigrationJobSummary};
 
 use super::status::MigrationProgress;
+use super::validate_migration;
 use super::{
     status::{MigrationJob, MigrationTask},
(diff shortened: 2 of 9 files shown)


libsql-server/src/database/schema.rs
```diff
@@ -52,12 +52,15 @@ impl crate::connection::Connection for SchemaConnection {
...
             let builder = tokio::task::spawn_blocking({
                 let migration = migration.clone();
                 move || {
                     let res = connection.with_raw(|conn| -> crate::Result<_> {
+                        if disable_foreign_key {
+                            conn.execute("PRAGMA foreign_keys=off", ())?;
+                        }
                         let mut txn = conn
                             .transaction_with_behavior(rusqlite::TransactionBehavior::Immediate)
                             .map_err(|_| {
```
haaawk, line 66: If this fails with error we won't re-enable foreign_keys, will we?

libsql-server/src/schema/scheduler.rs
```diff
@@ -773,6 +789,9 @@ async fn step_job_run_success(
                         txn.pragma_update(None, "schema_version", job_id)?;
                         // update schema version to job_id?
                         txn.commit()?;
```
haaawk, line 791: There are 3 statements here that can return with error and I believe we won't re-enable foreign_keys in such case, will we?

libsql-server/src/schema/scheduler.rs
```diff
@@ -523,6 +531,10 @@ async fn try_step_task_inner(
             let (new_status, error) = step_task(&mut txn, job_id)?;
             txn.commit()?;
```
haaawk, line 533: There's a bunch of statements here that can return with error and we won't re-enable foreign_keys then, no?

libsql-server/src/schema/db.rs
```diff
@@ -328,15 +329,17 @@ pub(super) fn get_next_pending_migration_job(
             |row| {
                 let job_id = row.get::<_, i64>(0)?;
                 let status = MigrationJobStatus::from_int(row.get::<_, u64>(1)?);
-                let migration = serde_json::from_str(row.get_ref(2)?.as_str()?).unwrap();
+                let mut migration = serde_json::from_str(row.get_ref(2)?.as_str()?).unwrap();
```
haaawk, line 332: Why do we need migration to be mut?

libsql-server/src/schema/mod.rs
```diff
@@ -47,37 +47,47 @@ pub use scheduler::Scheduler;
...
-        while let Some(step) = migration.steps.last() {
-            if !matches!(step.query.stmt.kind, StmtKind::TxnEnd) {
-                break;
+    }
+
+    // skip all steps that are not tx items
+    while steps.next_if(|s| !s.query.stmt.kind.is_txn()).is_some() {}
+
+    // last stmt can be a tx commit
+    if let Some(step) = steps.next() {
```
haaawk, line 75: This is not a regression but while we're at this we might fix the transactional batch case when both commit and rollback are added at the end of hrana batch. This code currently expects only one transaction end statement at the end but for our transactional batches we actually have two so we should probably handle both 1 and 2 transaction end statements here.

libsql-server/src/database/schema.rs
MarinPostma: this is per-connection, the connection will be dropped on error, and foreign key check will be re-enable fo the next connection

libsql-server/src/schema/db.rs
MarinPostma: that's because validation takes a mut, but in this case nothing will change because we've stored it already validated

PR #1712, opened 2024-08-30
add libsql_enable_tracing to the libsql

sivukhin:
## Context

LibSQL is used from multiple SDK. In order to simplify debugging of issues with these SDKs we can expose `libsql_enable_tracing` method which will print rust logs to the stdout (if `RUST_LOG` env var set properly)

### DIFF
--- a/bindings/c/src/lib.rs
+++ b/bindings/c/src/lib.rs
@@ -32,7 +32,7 @@ unsafe fn set_err_msg(msg: String, output: *mut *const std::ffi::c_char) {
 }
 
 #[no_mangle]
-pub unsafe extern "C" fn libsql_enable_tracing() -> std::ffi::c_int {
+pub unsafe extern "C" fn libsql_enable_internal_tracing() -> std::ffi::c_int {
     if tracing_subscriber::fmt::try_init().is_ok() {
         1
     } else {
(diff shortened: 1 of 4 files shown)


bindings/c/src/lib.rs
```diff
@@ -31,6 +31,11 @@ unsafe fn set_err_msg(msg: String, output: *mut *const std::ffi::c_char) {
     }
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn libsql_enable_tracing() {
+    tracing_subscriber::fmt::init();
```
LucioFranco, line 36: I wonder if we wanna do `let _ = tracing_subscriber::fmt::try_init()` so that repeated calls of this in the same process won't panic which I think would be quite hard to recover from in C code?

sivukhin: Sure! I guess we can also propagate return code which can be useful if for some reason tracing can't be enabled:
```
if tracing_subscriber::fmt::try_init().is_ok() {
    1
} else {
    0
}
```

MarinPostma:
shouldn't we rather integrate with whatever tracer the wrapping application uses?

MarinPostma:
It feels strange to me that a lib would install a logger with no control

LucioFranco:
> shouldn't we rather integrate with whatever tracer the wrapping application uses?

We already enable tracing for our other libraries the wrap the rust code. This works decently for debugging purposes.

MarinPostma (approved):

PR #1721, opened 2024-09-05
Only add to last_frames_synced if frames did change.

jameswritescode:
Fixes an issue where a returned `Replicated#frames_synced` will continuously increment based off the `replicator.frames_synced()` even if no new frames have synced.

### DIFF
--- a/libsql/src/replication/mod.rs
+++ b/libsql/src/replication/mod.rs
@@ -256,13 +256,26 @@ impl EmbeddedReplicator {
             }
         }
 
-        let last_frames_synced = self.last_frames_synced.fetch_add(
-            replicator.frames_synced(),
-            std::sync::atomic::Ordering::Relaxed,
-        );
+        let current_frames_synced = replicator.frames_synced();
+
+        let mut last_frames_synced = self
+            .last_frames_synced
+            .load(std::sync::atomic::Ordering::Relaxed);
+
+        while current_frames_synced > last_frames_synced {
+            match self.last_frames_synced.compare_exchange(
+                last_frames_synced,
+                current_frames_synced,
+                std::sync::atomic::Ordering::Relaxed,
+                std::sync::atomic::Ordering::Relaxed,
+            ) {
+                Ok(_) => break,
+                Err(current_value) => last_frames_synced = current_value,
+            }
+        }
 
         let frames_synced =
-            ((replicator.frames_synced() as i64 - last_frames_synced as i64).abs()) as usize;
+            ((current_frames_synced as i64 - last_frames_synced as i64).abs()) as usize;
 
         let replicated = Replicated {
             frame_no: replicator.client_mut().committed_frame_no(),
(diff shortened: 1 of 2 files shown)


libsql/src/replication/mod.rs
```diff
@@ -149,13 +149,13 @@ impl EmbeddedReplicator {
         encryption_config: Option<EncryptionConfig>,
         perodic_sync: Option<Duration>,
     ) -> Result<Self> {
-        let mut replicator =
```
haaawk, line 152: Is this done by cargo fmt or you ide? If your ide then please revert this.

libsql/src/replication/mod.rs
```diff
@@ -256,13 +256,19 @@ impl EmbeddedReplicator {
...
-            std::sync::atomic::Ordering::Relaxed,
-        );
+        let current_frames_synced = replicator.frames_synced();
+
+        let last_frames_synced = self
+            .last_frames_synced
+            .load(std::sync::atomic::Ordering::Relaxed);
+
+        if current_frames_synced > last_frames_synced {
+            self.last_frames_synced
```
haaawk, line 266: This is incorrect for concurrent updates. You need to use compare and set

libsql/src/replication/mod.rs
jameswritescode: rust analyzer did this, but it's not doing it if I use cargo fmt outside of analyzer so I reverted it.

libsql/src/replication/mod.rs
jameswritescode: Switched it to `compare_and_exchange`, is that what you're looking for?

libsql/src/replication/mod.rs
```diff
@@ -256,13 +256,23 @@ impl EmbeddedReplicator {
...
-            replicator.frames_synced(),
-            std::sync::atomic::Ordering::Relaxed,
-        );
+        let current_frames_synced = replicator.frames_synced();
+
+        let last_frames_synced = self
+            .last_frames_synced
+            .load(std::sync::atomic::Ordering::Relaxed);
+
+        if current_frames_synced > last_frames_synced {
```
haaawk, line 265: This needs to be a while loop. You fetch last_frames_synced, compare with current_frames_synced and if it's smaller then compare_exchange and if it succeeds then break the loop.

libsql/src/replication/mod.rs
```diff
@@ -256,13 +256,26 @@ impl EmbeddedReplicator {
...
+
+        while current_frames_synced > last_frames_synced {
+            match self.last_frames_synced.compare_exchange(
+                last_frames_synced,
+                current_frames_synced,
+                std::sync::atomic::Ordering::Relaxed,
+                std::sync::atomic::Ordering::Relaxed,
+            ) {
+                Ok(_) => break,
+                Err(current_value) => last_frames_synced = current_value,
```
jameswritescode, line 273: I feel like this could result in some bad math in `frames_synced` following this if for some reason `current_value` is larger.

LucioFranco (approved):

PR #1749, opened 2024-09-19
sqld: error out when trying to sync from schema db

LucioFranco:


haaawk (approved):

LucioFranco:
After discussing we actually don't want to error but actually proxy the request, so going to close and reopen with a new fix

PR #1750, opened 2024-09-19
sqld: schema db and replica fixes

LucioFranco:
This PR contains two fixes one for preventing embedded replica's from being able to sync a schema db and another that cleans up the conn.is_primary panics/errors that we had before.

### DIFF
--- a/libsql-server/src/rpc/proxy.rs
+++ b/libsql-server/src/rpc/proxy.rs
@@ -635,7 +635,7 @@ impl Proxy for ProxyService {
                         Ok(conn) => {
                             if !conn.is_primary() {
                                 return Err(tonic::Status::failed_precondition(
-                                    "cannot run schema migration against a replica",
+                                    "cannot run schema migration against a replica from a replica",
                                 ));
                             }
 
@@ -710,7 +710,7 @@ impl Proxy for ProxyService {
                     Ok(conn) => {
                         if !conn.is_primary() {
                             return Err(tonic::Status::failed_precondition(
-                                "cannot run schema migration against a replica",
+                                "cannot run schema migration against a replica from a replica",
                             ));
                         }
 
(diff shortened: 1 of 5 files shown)


libsql-server/src/rpc/proxy.rs
```diff
@@ -633,12 +633,6 @@ impl Proxy for ProxyService {
                     tracing::debug!("connected: {client_id}");
                     match connection_maker.create().await {
                         Ok(conn) => {
-                            if !conn.is_primary() {
```
LucioFranco, line 636: I removed this since this one actually didn't make much sense, we allow this in the stream_exec. I added an error message for the describe one which should catch schema changes.

LucioFranco: Okay this allows embedded replica's to "write" proxied to the primary for schema changes but does not allow them to sync.

libsql-server/tests/cluster/schema_dbs.rs
```diff
@@ -129,3 +129,64 @@ fn schema_migration_basics() {
...
+
+        {
+            let db = Database::open_remote_with_connector(
+                "http://schema.replica0:8080",
+                "",
+                TurmoilConnector,
+            )
+            .unwrap();
+            let conn = db.connect().unwrap();
+            conn.execute("select * from sqlite_master;", ())
```
MarinPostma, line 181: this looks like exactly the same test as the one in `cluster/mod` no?

MarinPostma (approved):

libsql-server/tests/cluster/schema_dbs.rs
LucioFranco: I copied portions of it but I don't think we ran it through this path specifically?

PR #1751, opened 2024-09-19
libsql wal encryption

MarinPostma:
- **fix compactor shortest path when there are no segments**
- **POC encryption libsql-wal**


penberg:
Closing as obsolete

PR #1770, opened 2024-10-03
sqld: temp fix replica namespace doesnt exist

LucioFranco:
there is a bug where a primary will report that a valid
                           namespace doesn't exist when it does and causes the replicate to
                            destroy the namespace locally. For now the temp solution is to
                            ignore this error (still log it) and don't destroy the local
                          namespace data on the replica. This trades off storage space for
              returing 500s.

### DIFF
--- a/libsql-server/src/namespace/configurator/replica.rs
+++ b/libsql-server/src/namespace/configurator/replica.rs
@@ -123,10 +123,17 @@ impl ConfigureNamespace for ReplicaConfigurator {
                 loop {
                     match replicator.run().await {
                         err @ Error::Fatal(_) => Err(err)?,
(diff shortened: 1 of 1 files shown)


MarinPostma (approved):

haaawk (approved):

penberg:
@haaawk @LucioFranco Please make sure there is an issue about the root cause bug too so we remember to fix it.

haaawk:
https://github.com/tursodatabase/libsql/issues/1772

haaawk:
I was hoping we could fix the root issue on the primary but this is a good start to mitigate the problem

MarinPostma:
Maybe we should let the control plane eagerly create namespaces on replicas. That would make stuff easier for sqld

PR #1774, opened 2024-10-04
Worker version updated to latest 0.4.1

kunjee17:
I don't what more is required. I tried to go throw read me for contribution guideline but can't find one. I just needed to updated worker version to match latest. Build is working in `WSL` . Do let me know if any further checks or correction is required.


kunjee17:
I have updated worker version as it is not working for me with cloudflare worker project in rust. Once it is merged will try again. If still it doesn't work then will file an issue.

Issue #1778, opened 2024-10-05
Deserialization does not support newtypes


sjunepark:
I'm creating a issue corresponding to a PR Im making. (Relevant tests will be included in the PR)
In the current version, newtype fields aren't able to be deserialized as seen below.

```rust
#[cfg(feature = "serde")]
#[tokio::test]
async fn deserialize_row() {

// ... ommitted

    #[derive(Deserialize, Debug)]
    struct Data {
        id: i64,
        name: String,
        score: f64,
        data: Vec<u8>,
        age: Option<i64>,
        none: Option<()>,
        status: Status,
        wrapper: Wrapper,
        newtype: NewType,
    }

    #[derive(Deserialize, Debug, PartialEq)]
    struct NewType(String);

// ... ommitted

    let data: Data = libsql::de::from_row(&row).unwrap();

// ... ommitted

    // Fails !!!
    assert_eq!(data.newtype, NewType("Newtype".to_string())); 
}   
```

PR #1779, opened 2024-10-05
Enable deserialization for newtype field values

sjunepark:
PR in relevance to #1778 .

The original code used `deserialize_any` for newtypes.

Implemented a separate `deserialize_newtype_struct`.


sjunepark:
I wasn't to run tests for the whole workspace due to errors like below, which I wasn't able to fix.
I don't think this will break other test, but I'm not sure.

```shell
  = note: Undefined symbols for architecture arm64:
            "_sqlite3_unlock_notify", referenced from:
                rusqlite::unlock_notify::wait_for_unlock_notify::h68227f6013bf55ed in rusqlite-d113afe8a2e6076a.08udfb9kdnese01layhcpep3w.rcgu.o
          ld: symbol(s) not found for architecture arm64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Issue #1657, opened 2024-08-10
libsql-ffi build failed on Windows


wyhaya:
```bash
cargo new demo
cd demo
cargo add libsql
cargo check
```
### Output
```
error: failed to run custom build command for `libsql-ffi v0.4.0`

Caused by:
  process didn't exit successfully: `C:\Users\...\target\debug\build\libsql-ffi-845dd5bc7ddd4d05\build-script-build` (exit code: 101)
  --- stdout
  cargo:rerun-if-changed=bundled/src/sqlite3.c

  --- stderr
  thread 'main' panicked at C:\Users\...\.cargo\registry\src\index.crates.io-6f17d22bba15001f\libsql-ffi-0.4.0\build.rs:48:10:
  called `Result::unwrap()` on an `Err` value: Error { kind: NotFound, message: "program not found" }
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

https://github.com/tursodatabase/libsql/blob/807794891f2a3b9a60711bffdd0c9afdc0195135/libsql-ffi/build.rs#L42-L48

`cp` is sometimes not available on Windows.

---
`Windows 11` `rustc 1.80.0` `libsql = "0.5.0"`

milen-denev:
I have the same issue

matiassotomejias:
Hi, were you able to solve the problem? I have the same error

wyhaya:
> Hi, were you able to solve the problem? I have the same error


You can temporarily use the older version v0.4, which doesn't have this issue.
```
libsql = "0.4.0"
```

milen-denev:
> Hi, were you able to solve the problem? I have the same error

No. The only solution is to use libsql 0.4.0

milen-denev:
The reason this occurs is because in the build.rs script the developers have put the Linux `cp` command for copying. I created my own copycat of `cp`, [rust_cp](https://github.com/milen-denev/rust_cp). The repository has compiled binary, put this in any location you want on your windows system, and add Path environment variable, restart the system, and it should work.

![image](https://github.com/user-attachments/assets/be65c124-e7b2-4840-ac23-5fa3cc28f2df)

Cammisuli:
I'm hoping that this can be fixed soon. Otherwise I'm gonna have to use a patched version, which is quite inconvenient ☹

awakecoding:
[...]

PR #1791, opened 2024-10-22
fix libsql-ffi build.rs usage of cp command on Windows

awakecoding:
Fix for https://github.com/tursodatabase/libsql/issues/1657

### DIFF
--- a/libsql-ffi/build.rs
+++ b/libsql-ffi/build.rs
@@ -1,7 +1,7 @@
 use std::env;
 use std::ffi::OsString;
 use std::fs::{self, OpenOptions};
-use std::io::Write;
+use std::io::{self, Write};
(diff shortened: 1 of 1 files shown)


levydsa (approved):

GutZuFusss:
Can we expect this to be released anytime soon? Or should we look for a workaround, since this is already broken since 0.5.0?

LucioFranco:
This actually had to be reverted https://github.com/tursodatabase/libsql/commit/9499c3c40aa299b0ce5221b53f000f69462f8d8c so we need to get this fixed for all platforms before we can release that.

GutZuFusss:
Ok, thanks for the feedback! I'll work with a git dependency in the meantime then, thanks for your work!

PR #1793, opened 2024-10-23
Add Database::get_sync_usage_stats

haaawk:
Track detailed usage of synced bytes that will hopefully allow to identify why we're accounting so many bytes for sync.

Expose those detailed stats through Database::get_sync_usage_stats.

### DIFF
--- a/libsql/src/replication/mod.rs
+++ b/libsql/src/replication/mod.rs
@@ -65,6 +65,60 @@ pub enum Frames {
     Snapshot(SnapshotFile),
 }
 
+/// Detailed logs about bytes synced with primary
+pub struct SyncUsageStats {
+    prefetched_bytes: u64,
+    prefetched_bytes_discarded_due_to_new_session: u64,
+    prefetched_bytes_discarded_due_to_consecutive_handshake: u64,
+    prefetched_bytes_discarded_due_to_invalid_frame_header: u64,
+    synced_bytes_discarded_due_to_invalid_frame_header: u64,
+    prefetched_bytes_used: u64,
+    synced_bytes_used: u64,
+    snapshot_bytes: u64,
+}
+
+impl SyncUsageStats {
+    /// Number of bytes prefetched while doing handshake
+    pub fn prefetched_bytes(&self) -> u64 {
+        self.prefetched_bytes
+    }
+
+    /// Number of bytes prefetched and discarded due to the change of the client session
+    pub fn prefetched_bytes_discarded_due_to_new_session(&self) -> u64 {
+        self.prefetched_bytes_discarded_due_to_new_session
+    }
+
+    /// Number of bytes prefetched and discarded due to consecutive handshake with new prefetch
+    pub fn prefetched_bytes_discarded_due_to_consecutive_handshake(&self) -> u64 {
+        self.prefetched_bytes_discarded_due_to_consecutive_handshake
+    }
+
+    /// Number of bytes prefetched and discarded due to invalid frame header in received frames
+    pub fn prefetched_bytes_discarded_due_to_invalid_frame_header(&self) -> u64 {
+        self.prefetched_bytes_discarded_due_to_invalid_frame_header
+    }
+
+    /// Number of bytes synced and discarded due to invalid frame header in received frames
+    pub fn synced_bytes_discarded_due_to_invalid_frame_header(&self) -> u64 {
+        self.synced_bytes_discarded_due_to_invalid_frame_header
+    }
+
+    /// Number of bytes prefetched and used
+    pub fn prefetched_bytes_used(&self) -> u64 {
+        self.prefetched_bytes_used
+    }
+
+    /// Number of bytes synced and used
+    pub fn synced_bytes_used(&self) -> u64 {
+        self.synced_bytes_used
+    }
+
+    /// Number of bytes downloaded as snapshots
+    pub fn snapshot_bytes(&self) -> u64 {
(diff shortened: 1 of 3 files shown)


penberg (approved):

haaawk:
What we should do next is:
1. Expose this in JS and Go SDKs
2. Ask users to upgrade and share their numbers after a while
3. Setup our own reproducer on fly (primary, some loader machine that keeps writing ideally updates mostly, one or two embedded replicas clients that sync periodically)

libsql/src/replication/mod.rs
```diff
@@ -65,6 +65,19 @@ pub enum Frames {
...
+#[derive(Debug)]
+pub struct SyncUsageStats {
+    pub prefetched_bytes: u64,
+    pub prefetched_bytes_discarded_due_to_new_session: u64,
+    pub prefetched_bytes_discarded_due_to_consecutive_handshake: u64,
+    pub prefetched_bytes_discarded_due_to_invalid_frame_header: u64,
+    pub synced_bytes_discarded_due_to_invalid_frame_header: u64,
+    pub prefetched_bytes_used: u64,
+    pub synced_bytes_used: u64,
+    pub snapshot_bytes: u64,
```
LucioFranco, line 78: We should convert these to getters rather than public fields on a struct. More so for public structs that we expose to users and we should follow the style in the rest of the crate for that. If these are pub for internal reasons then they should be `pub(crate)` etc.

libsql/src/replication/mod.rs
```diff
@@ -210,6 +223,37 @@ impl EmbeddedReplicator {
...
+                        .prefetched_bytes_discarded_due_to_consecutive_handshake.load(std::sync::atomic::Ordering::SeqCst),
+                    prefetched_bytes_discarded_due_to_invalid_frame_header: stats
+                        .prefetched_bytes_discarded_due_to_invalid_frame_header.load(std::sync::atomic::Ordering::SeqCst),
+                    synced_bytes_discarded_due_to_invalid_frame_header: stats
+                        .synced_bytes_discarded_due_to_invalid_frame_header.load(std::sync::atomic::Ordering::SeqCst),
+                    prefetched_bytes_used: stats.prefetched_bytes_used.load(std::sync::atomic::Ordering::SeqCst),
+                    synced_bytes_used: stats.synced_bytes_used.load(std::sync::atomic::Ordering::SeqCst),
+                    snapshot_bytes: stats.snapshot_bytes.load(std::sync::atomic::Ordering::SeqCst),
+                };
+                tracing::info!("Sync usage stats: {:?}", result);
```
LucioFranco, line 250: I don't think we want this to be info? I think this should at least be debug or else it will be spammed in the logs often? I think we should let users choose if they want this to be info or not by having their own log statement.

libsql/src/replication/remote_client.rs
```diff
@@ -107,8 +172,12 @@ impl RemoteClient {
         Ok(new_session)
     }
 
-    async fn do_handshake_with_prefetch(&mut self) -> (Result<bool, Error>, Duration) {
+    async fn do_handshake_with_prefetch(&mut self) -> (Result<(), Error>, Duration) {
         tracing::info!("Attempting to perform handshake with primary.");
+        if let Some((Ok(frames), _)) = &self.prefetched_batch_log_entries {
+            let bytes = frames.get_ref().frames.iter().map(|f| f.data.len() as u64).sum();
```
LucioFranco, line 178: do we even need to sum the length, can we just assume that each frame is 4kb and just take frames.len() * 4kb?

LucioFranco (changes requested):

libsql/src/replication/mod.rs
haaawk: This will print only as often as user calls get_sync_usage_stats. This API is not meant to be called often.
What I want to achieve with this info is to make it easy for us to work with users by saying:
1. Call get_sync_usage_stats
2. Find "Sync usage stats:" log line and send to us.

libsql/src/replication/remote_client.rs
haaawk: I'm not sure but it is a good question.

libsql/src/replication/mod.rs
haaawk: Removing the tracing since it's not crucial to the PR and I don't want to spend time discussing this.

libsql/src/replication/remote_client.rs
haaawk: Added a TODO to check if it's correct.

libsql/src/replication/mod.rs
haaawk: Done.

LucioFranco (approved):

libsql/src/replication/mod.rs
```diff
@@ -65,6 +65,52 @@ pub enum Frames {
...
+    prefetched_bytes_discarded_due_to_consecutive_handshake: u64,
+    prefetched_bytes_discarded_due_to_invalid_frame_header: u64,
+    synced_bytes_discarded_due_to_invalid_frame_header: u64,
+    prefetched_bytes_used: u64,
+    synced_bytes_used: u64,
+    snapshot_bytes: u64,
+}
+
+impl SyncUsageStats {
+    pub fn prefetched_bytes(&self) -> u64 {
```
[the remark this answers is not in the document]
haaawk: Done.

PR #1801, opened 2024-10-30
Initial offline write support

penberg:
This pull request adds a new type of database, synced database, which
allows local writes that can sync to a remote server.

The protocol on server side is simple:

  **POST** `/sync/<generation>/<start_frame>/<end_frame>`

where request body is the WAL frames.

### DIFF
--- a/libsql/src/local/connection.rs
+++ b/libsql/src/local/connection.rs
@@ -57,13 +57,23 @@ impl Connection {
                 )));
             }
         }
-
-        Ok(Connection {
+        let conn = Connection {
             raw,
             drop_ref: Arc::new(()),
             #[cfg(feature = "replication")]
             writer: db.writer()?,
-        })
+        };
+        #[cfg(feature = "sync")]
+        if let Some(_) = db.sync_ctx {
+            // We need to make sure database is in WAL mode with checkpointing
+            // disabled so that we can sync our changes back to a remote
+            // server.
+            conn.query("PRAGMA journal_mode = WAL", Params::None)?;
+            unsafe {
+                ffi::libsql_wal_disable_checkpoint(conn.raw);
+            }
+        }
+        Ok(conn)
     }
 
     /// Get a raw handle to the underlying libSQL connection
--- a/libsql/src/local/database.rs
+++ b/libsql/src/local/database.rs
@@ -2,8 +2,7 @@ use std::sync::Once;
 
 cfg_replication!(
     use http::uri::InvalidUri;
-    use crate::database::EncryptionConfig;
-    use libsql_replication::frame::FrameNo;
+    use crate::database::{EncryptionConfig, FrameNo};
 
     use crate::replication::client::Client;
     use crate::replication::local_client::LocalClient;
@@ -19,6 +18,10 @@ cfg_replication!(
     }
 );
 
+cfg_sync! {
+    use crate::sync::SyncContext;
+}
+
 use crate::{database::OpenFlags, local::connection::Connection};
 use crate::{Error::ConnectionFailed, Result};
 use libsql_sys::ffi;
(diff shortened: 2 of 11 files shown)


libsql/src/local/database.rs
```diff
@@ -351,6 +377,76 @@ impl Database {
...
+        let uri = format!("{}/sync/{}/{}/{}", sync_ctx.sync_url, generation, frame_no, frame_no+1);
+        self.push_with_retry(uri, &sync_ctx.auth_token, frame.to_vec(), sync_ctx.max_retries).await?;
+        Ok(())
+    }
+
+    #[cfg(feature = "sync")]
+    async fn push_with_retry(&self, uri: String, auth_token: &Option<String>, frame: Vec<u8>, max_retries: usize) -> Result<()> {
+        let mut nr_retries = 0;
+        loop {
+            let client = reqwest::Client::new();
```
LucioFranco, line 430: Whats the reason for using reqwest instead of just hyper directly? We also probably want to cache this client so we can re-use its dns resolution/tls handshake.

libsql/src/local/database.rs
```diff
@@ -351,6 +377,76 @@ impl Database {
...
+                None => {}
+            }
+            let res = builder.body(frame.to_vec()).send().await.unwrap();
+            if res.status().is_success() {
+                return Ok(());
+            }
+            if nr_retries > max_retries {
+                return Err(crate::errors::Error::ConnectionFailed(format!("Failed to push frame: {}", res.status())));
+            }
+            std::thread::sleep(std::time::Duration::from_millis(100 * (1 << nr_retries)));
```
LucioFranco, line 445: this should be an async sleep to avoid blocking the workers.

libsql/src/replication/mod.rs
```diff
@@ -34,27 +34,7 @@ mod connection;
...
-        self.frame_no
-    }
-
-    /// The count of frames synced during this call of `sync`. A frame is a 4kB frame from the
-    /// libsql write ahead log.
-    pub fn frames_synced(&self) -> usize {
-        self.frames_synced
-    }
-}
+use crate::database::Replicated;
```
LucioFranco, line 37:
```suggestion
pub use crate::database::Replicated;
```

Need to re-export to make code backwards compat.

libsql/src/sync.rs
```diff
@@ -0,0 +1,18 @@
+const DEFAULT_MAX_RETRIES: usize = 5;
+pub struct SyncContext {
+    pub sync_url: String,
+    pub auth_token: Option<String>,
+    pub max_retries: usize,
+    pub durable_frame_num: u32,
```
LucioFranco, line 6: We should provide getters for these rather than expose them. I can do this in a follow up.

libsql/src/local/connection.rs
```diff
@@ -57,13 +57,23 @@ impl Connection {
...
-        })
+        };
+        #[cfg(feature = "sync")]
+        if let Some(_) = db.sync_ctx {
+            // We need to make sure database is in WAL mode with checkpointing
+            // disabled so that we can sync our changes back to a remote
+            // server.
+            conn.query("PRAGMA journal_mode = WAL", Params::None)?;
+            unsafe {
+                ffi::libsql_wal_disable_checkpoint(conn.raw);
```
LucioFranco, line 73: Do we need to adding parsing to prevent users from check pointing on their own?

LucioFranco (approved):
left some feedback but overall LGTM

libsql/src/local/connection.rs
penberg: Added a note of that in https://gist.github.com/penberg/c8082b450fa8ba56a96c2d9c0d45ad8f

PR #1820, opened 2024-11-14
libsql: rework sync v2 structure

LucioFranco:


libsql/src/local/database.rs
```diff
@@ -380,84 +387,54 @@ impl Database {
...
-        
+        let max_frame_no = conn.wal_frame_count();
+
         let generation = 1; // TODO: Probe from WAL.
-        let start_frame_no = sync_ctx.durable_frame_num + 1;
+        let start_frame_no = ctx.durable_frame_num() + 1;
         let end_frame_no = max_frame_no;
 
+        // TODO: figure out relation to durable_frame_num
+        // let max_frame_no = ctx.max_frame_no();
```
LucioFranco, line 414: @penberg before we can merge this would be helpful to clairify the difference between durable_frame_ num and max_frame_no. What I understand is we want to write the durable_frame_num here (which I guess is the max_frame_no from the server) to disk to ensure we start at a more advanced wal frame than 0.

libsql/src/sync.rs
```diff
@@ -1,18 +1,149 @@
...
+
+                let resp = serde_json::from_slice::<serde_json::Value>(&body[..]).unwrap();
+
+                let max_frame_no = resp.get("max_frame_no").unwrap().as_u64().unwrap() as u32;
+
+                // Update our best known max_frame_no from the server and write it to disk.
+                self.set_max_frame_no(max_frame_no).await.unwrap();
+
+                return Ok(max_frame_no);
+            } else if res.status().is_server_error() || attempts < self.max_retries {
```
LucioFranco, line 88: @penberg small change compared to what you had, I just retry on server failures rather than all failures like auth_token being invalid.

libsql/src/local/database.rs
penberg: You always have two different max frame numbers: one on client, one on server. I tend to call the latter `durable_frame_num`.

libsql/src/local/database.rs
```diff
@@ -320,7 +324,10 @@ impl Database {
 
     #[cfg(feature = "replication")]
     /// Sync with primary at least to a given replication index
-    pub async fn sync_until(&self, replication_index: FrameNo) -> Result<crate::database::Replicated> {
+    pub async fn sync_until(
+        &self,
+        replication_index: FrameNo,
+    ) -> Result<crate::database::Replicated> {
```
haaawk, line 330: Please avoid reformattings like that in PRs that are supposed to do something non-trivial.

libsql/src/local/database.rs
```diff
@@ -380,84 +387,54 @@ impl Database {
...
+            Some(ctx) => ctx.lock().await,
+            None => panic!("sync context not set"),
+        };
+
         let conn = self.connect()?;
 
+        // TODO: can this be cached?
         let page_size = {
-            let rows = conn.query("PRAGMA page_size", crate::params::Params::None)?.unwrap();
+            let rows = conn
```
haaawk, line 399: Again - please avoid noise like that in PRs

libsql/src/local/connection.rs
```diff
@@ -445,6 +446,39 @@ impl Connection {
...
+        let mut max_frame_no: std::os::raw::c_uint = 0;
+        unsafe { libsql_sys::ffi::libsql_wal_frame_count(self.handle(), &mut max_frame_no) };
+
+        max_frame_no
+    }
+
+    pub fn wal_get_frame(&self, frame_no: u32, page_size: u32) -> Result<BytesMut> {
+        let frame_size: usize = 24 + page_size as usize;
+
+        let mut buf = BytesMut::with_capacity(frame_size);
```
haaawk, line 460: Why is this better than just vec![0; frame_size]?

libsql/src/sync.rs
```diff
@@ -1,18 +1,149 @@
...
+        let path = format!("{}-info", self.db_path);
+
+        let contents = tokio::fs::read(&path).await.unwrap();
+
+        let metadata = serde_json::from_slice::<MetadataJson>(&contents[..]).unwrap();
+
+        self.max_frame_no = metadata.max_frame_no;
+
+        Ok(())
+    }
```
haaawk, line 143: What's all of this? Can we have one PR that just switches to hyper and all other changes in another PR please?

haaawk: But also what's this?

libsql/src/local/database.rs
LucioFranco: Seems like rustfmt wasn't run on this file (I have it run on save), what we should have is CI enforcing that we don't merge PRs without it passing rustfmt checks so I will add that.

haaawk:
Yes but since then please post stuff like that in a separate PRs - it costs
little, gets merged sooner and makes the main PR simpler.

libsql/src/local/database.rs
LucioFranco: Really odd, my editor is auto formatting this but with `cargo fmt` which we use to check on CI so we avoid these changes its not formatting lines with `#[cfg(feature = "")]`. Gonna take a look into this, seems really odd that cargo fmt isn't picking this up.

haaawk: I wouldn't invest time into that. Just when you see stuff like this put it on a separate PR and even merge it without review.

PR #1825, opened 2024-11-18
libsql: Switch to hyper for local offline

LucioFranco:
This commit switches our usage of reqwest to hyper. In the future, this will allow us to pass in configuration for the connector to support both custom TLS implementations (aka using the system tls impl over rustls) and connecting to the sqld network simulation for future testing. This also removes the extra dependency on reqwest which is not needed since we already pull in the dependencies needed to implement the new protocol for local offline writes.

### DIFF
--- a/libsql/Cargo.toml
+++ b/libsql/Cargo.toml
@@ -42,7 +42,6 @@ fallible-iterator = { version = "0.3", optional = true }
 
 libsql_replication = { version = "0.6", path = "../libsql-replication", optional = true }
 async-stream = { version = "0.3.5", optional = true }
-reqwest = { version = "0.12.9", default-features = false, features = [ "rustls-tls", "json" ], optional = true }
 
 [dev-dependencies]
 criterion = { version = "0.5", features = ["html_reports", "async", "async_futures", "async_tokio"] }
@@ -105,7 +104,6 @@ sync = [
   "dep:bytes",
   "dep:tokio",
   "dep:futures",
-  "dep:reqwest",
   "dep:serde_json",
 ]
 hrana = [
--- a/libsql/src/local/database.rs
+++ b/libsql/src/local/database.rs
@@ -476,21 +476,40 @@ impl Database {
     ) -> Result<u32> {
         let mut nr_retries = 0;
         loop {
-            let client = reqwest::Client::new();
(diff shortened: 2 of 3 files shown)


libsql/src/local/database.rs
```diff
@@ -476,21 +476,40 @@ impl Database {
...
+                    req.headers_mut()
+                        .expect("valid http request")
+                        .insert("Authorization", auth_header);
                 }
                 None => {}
             }
-            let res = builder.body(frame.to_vec()).send().await.unwrap();
+
+            // TODO(lucio): convert this to use bytes to make this clone cheap, it should be
+            // possible use BytesMut when reading frames from the WAL and efficiently use Bytes
```
haaawk, line 498: possible to use?

libsql/Cargo.toml
```diff
@@ -42,7 +42,6 @@ fallible-iterator = { version = "0.3", optional = true }
 
 libsql_replication = { version = "0.6", path = "../libsql-replication", optional = true }
 async-stream = { version = "0.3.5", optional = true }
-reqwest = { version = "0.12.9", default-features = false, features = [ "rustls-tls", "json" ], optional = true }
```
haaawk, line 45: I was going to ask why we're not removing reqwest from Cargo.toml but now we do :)

libsql/src/local/database.rs
```diff
@@ -476,21 +476,40 @@ impl Database {
...
+                        .insert("Authorization", auth_header);
                 }
                 None => {}
             }
-            let res = builder.body(frame.to_vec()).send().await.unwrap();
+
+            // TODO(lucio): convert this to use bytes to make this clone cheap, it should be
+            // to possible use BytesMut when reading frames from the WAL and efficiently use Bytes
+            // from that.
+            let req = req.body(frame.clone().into()).expect("valid body");
```
haaawk, line 500: body wouldn't accept frame.to_vec()?

haaawk (approved):

libsql/src/local/database.rs
LucioFranco: I found `to_vec` to not be clear since, its already a `Vec<u8>` so this instead reads copy the frame and convert it into the type accepted by body (which in this case is hyper::Body). This should still compile to the same thing.

PR #1827, opened 2024-11-18
libsql: refactor push and sync code

LucioFranco:
This PR has multiple commits each broken up into their own refactor piece. Broken down is the list of refactors and why:

- Move sync ctx behind a mutex, this in general makes sense since we only want one push to be active at a time but also will enable the future durable_frame_no caching by allow us to mutate the struct on push. It is an async lock since we want to sync access to push over long async periods.
- move push_with_retry into sync.rs
- move push_one_frame into sync.rs these both make sense since the data they depend on is inside the sync_ctx struct so this allows us to make those fields private.
- Following up on those two, the next commit makes those fields private to improve code encapsulation.
- Small refactor now to use the auth_token field from the struct.

This PR just moves some code around but doesn't actually touch any of the logic in the code.

### DIFF
--- a/libsql/src/sync.rs
+++ b/libsql/src/sync.rs
@@ -2,10 +2,10 @@ use crate::Result;
 
 const DEFAULT_MAX_RETRIES: usize = 5;
 pub struct SyncContext {
-    pub sync_url: String,
-    pub auth_token: Option<String>,
-    pub max_retries: usize,
-    pub durable_frame_num: u32,
+    sync_url: String,
+    auth_token: Option<String>,
+    max_retries: usize,
+    durable_frame_num: u32,
 }
 
 impl SyncContext {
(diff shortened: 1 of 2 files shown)


libsql/src/sync.rs
```diff
@@ -18,7 +18,27 @@ impl SyncContext {
         }
     }
 
-    pub(crate) async fn push_with_retry(
+    pub(crate) async fn push_one_frame(
```
haaawk, line 21: Is this still guarded with `#[cfg(feature = "sync")]` like previously?

LucioFranco:
Last commit contains a bit more since for some reason an old hyper commit got layered into the commit range.

LucioFranco:
Should now not contain any hyper changes and should just be a refactor of the old functions.

libsql/src/sync.rs
LucioFranco: Not needed since `sync.rs` is behind that feature flag.

haaawk (approved):

PR #1830, opened 2024-11-19
libsql: add `durable_frame_num` caching and metadata file

### LINKED ISSUE #1818: Cache server-side frame number in client
The WAL push protocol allows the client to send frames that are already applied on the server WAL (they will just be ignored). However, this is obviously slow, which is why the client uses server hints to skip frames: https://github.com/tursodatabase/libsql/pull/1817

We can do better, though. Let's cache the server-side max frame number on client side in a `<database>-info` file (similar to replication index). Let's make it use JSON so we can add more metadata later. The file should be lazily populated when we get server-side hints.

LucioFranco:
This PR adds both a metadata file that is used to cache the `durable_frame_num` to allow us to only send the frames the server needs. This also adds some basic logging to allow us to track what frames and generations are being pushed to the server.

Closes #1818 

```
❯ RUST_LOG=libsql::sync=debug cargo run --example offline_writes
    Blocking waiting for file lock on build directory
   Compiling libsql v0.6.0 (/home/lucio/code/libsql/libsql)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 10.14s
     Running `target/debug/examples/offline_writes`
2024-11-19T19:57:20.054954Z DEBUG libsql::sync: no metadata info file found
Please write your entry to the guestbook:
hello
You entered: hello

Guest book entries:
  cargo run --example offline_writeshello

  hello

  hello

  hello

  hello

Syncing database to remote...
2024-11-19T19:57:23.677365Z DEBUG push_one_frame{generation=1 frame_no=1}: libsql::sync: pushing frame
2024-11-19T19:57:23.781639Z DEBUG push_one_frame{generation=1 frame_no=1}: libsql::sync: frame successfully pushed durable_frame_num=6
2024-11-19T19:57:23.781975Z DEBUG push_one_frame{generation=1 frame_no=7}: libsql::sync: pushing frame
2024-11-19T19:57:23.837472Z DEBUG push_one_frame{generation=1 frame_no=7}: libsql::sync: frame successfully pushed durable_frame_num=6
Done!
libsql on  lucio/cache-server-frame [$!?] via 🦀 v1.81.0
❯ RUST_LOG=libsql::sync=debug cargo run --example offline_writes
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.23s
     Running `target/debug/examples/offline_writes`
Please write your entry to the guestbook:
bye
You entered: bye

Guest book entries:
  cargo run --example offline_writeshello

  hello

  hello

  hello

  hello

  bye

Syncing database to remote...
2024-11-19T19:57:42.074752Z DEBUG push_one_frame{generation=1 frame_no=7}: libsql::sync: pushing frame
2024-11-19T19:57:42.118522Z DEBUG push_one_frame{generation=1 frame_no=7}: libsql::sync: frame successfully pushed durable_frame_num=7
2024-11-19T19:57:42.118936Z DEBUG push_one_frame{generation=1 frame_no=8}: libsql::sync: pushing frame
2024-11-19T19:57:42.162425Z DEBUG push_one_frame{generation=1 frame_no=8}: libsql::sync: frame successfully pushed durable_frame_num=7
Done!
```


LucioFranco:
@penberg in the logs above I see it is re-sending frame 7 because the server on the previous call returned max_frame_no 6 when we sent frame_no 7. Is there an off by one on the server side or is that to be expected with the protocol?

libsql/src/sync.rs
```diff
@@ -75,4 +99,46 @@ impl SyncContext {
...
+        self.durable_frame_num = metadata.durable_frame_num;
+
+        Ok(())
+    }
+}
+
+#[derive(serde::Serialize, serde::Deserialize)]
+struct MetadataJson {
+    version: u32,
+    durable_frame_num: u32,
```
haaawk, line 143: Should we make it u64 or even u128 to make it more future proof?

libsql/src/sync.rs
```diff
@@ -75,4 +99,46 @@ impl SyncContext {
...
+        );
+
+        self.durable_frame_num = metadata.durable_frame_num;
+
+        Ok(())
+    }
+}
+
+#[derive(serde::Serialize, serde::Deserialize)]
+struct MetadataJson {
```
haaawk, line 141: Maybe we could add a field with hash of other fields and verify on read that file is not corrupted?

libsql/src/sync.rs
```diff
@@ -75,4 +99,46 @@ impl SyncContext {
...
+    async fn write_metadata(&mut self) -> Result<()> {
+        let path = format!("{}-info", self.db_path);
+
+        let contents = serde_json::to_vec(&MetadataJson {
+            version: METADATA_VERSION,
+            durable_frame_num: self.durable_frame_num,
+        })
+        .unwrap();
+
+        tokio::fs::write(path, contents).await.unwrap();
```
haaawk, line 112: Wouldn't it be safer to write to tmp file and then atomically rename to avoid situation when process crashes in the middle of writting?

libsql/src/sync.rs
```diff
@@ -31,11 +45,21 @@ impl SyncContext {
...
             .push_with_retry(uri, frame.to_vec(), self.max_retries)
             .await?;
 
-        Ok(max_frame_no)
+        tracing::debug!(?durable_frame_num, "frame successfully pushed");
+
+        // Update our last known max_frame_no from the server.
+        self.durable_frame_num = durable_frame_num;
+
+        self.write_metadata().await?;
```
haaawk, line 60: Do we have to write metadata on every frame? Maybe we could do that after each batch?

libsql/src/sync.rs
```diff
@@ -1,25 +1,39 @@
...
             max_retries: DEFAULT_MAX_RETRIES,
-        }
+        };
+
+        me.read_metadata().await?;
+
+        Ok(me)
     }
 
+    #[tracing::instrument(skip(self, frame))]
```
haaawk, line 34: What does this do?

libsql/src/sync.rs
LucioFranco: These were selected by @penberg so he should comment on making those changes.

penberg: We should add checkpoint generation there

penberg:
@LucioFranco What happens if the metadata file is corrupted? The code should recover from that by trying to push the whole log.

libsql/src/sync.rs
LucioFranco: This attaches a tracing span to each call. So this adds a span that includes the frame_no and generation being sent and any log that is produced within this fn (and span) will have that data attached allowing us to simplify the actual log lines.

libsql/src/sync.rs
LucioFranco: I am adding the tracking code for this but atm we only have generation set to 1.

LucioFranco: https://github.com/tursodatabase/libsql/pull/1830/commits/3575914dbaa856eaaa59f9fe070c48eca206e250

libsql/src/sync.rs
LucioFranco: https://github.com/tursodatabase/libsql/pull/1830/commits/de24d8211e4dd42833f962de0188087f896206f1

libsql/src/sync.rs
```diff
@@ -93,4 +121,175 @@ impl SyncContext {
...
+        }
+
+        let contents = tokio::fs::read(&path).await.unwrap();
+
+        let metadata = serde_json::from_slice::<MetadataJson>(&contents[..]).unwrap();
+
+        metadata.verify_hash()?;
+
+        // TODO(lucio): convert this into a proper error
+        assert_eq!(
```
haaawk, line 163: debug assert maybe?

libsql/src/sync.rs
haaawk: What does this mean?

haaawk: I don't see this being fixed in commit de24d82

libsql/src/sync.rs
LucioFranco: I think all in all we want this to panic in every release mode but I expect this to go away very shortly in a follow up.

libsql/src/sync.rs
```diff
@@ -93,4 +121,175 @@ impl SyncContext {
...
+        if !std::fs::exists(&path).unwrap() {
+            tracing::debug!("no metadata info file found");
+            return Ok(());
+        }
+
+        let contents = tokio::fs::read(&path).await.unwrap();
+
+        let metadata = serde_json::from_slice::<MetadataJson>(&contents[..]).unwrap();
+
+        metadata.verify_hash()?;
```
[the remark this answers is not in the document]
LucioFranco: I think we want to basically reset these values anytime this read_metadata function fails?

libsql/src/sync.rs
LucioFranco: https://github.com/tursodatabase/libsql/pull/1830/commits/de24d8211e4dd42833f962de0188087f896206f1

penberg:
@LucioFranco I corrupted the info file as follows:

```
{"hash":3168545583"version":0,"durable_frame_num":120,"generation":1}
```

and now I see this error:

```
Running `/Users/penberg/src/tursodatabase/libsql/target/debug/examples/offline_writes`
thread 'main' panicked at /Users/penberg/src/tursodatabase/libsql/libsql/src/sync.rs:161:78:
called `Result::unwrap()` on an `Err` value: Error("expected `,` or `}`", line: 1, column: 19)
```

As I said before, the info file is for _caching_ purposes. If we fail to read it, we should just discard it and regenerate it after chatting with the server.

libsql/src/sync.rs
LucioFranco: sorry I made a mistake it wasn't done in that commit but its done here https://github.com/tursodatabase/libsql/pull/1830/commits/26ac07e35e549eb8dfd77ccfde9960de85079800

LucioFranco:
@penberg that has been fixed/implemented and there are a bunch of tests including a corrupted metadata file.

haaawk (approved):

libsql/src/local/database.rs
```diff
@@ -402,7 +402,7 @@ impl Database {
 
         let max_frame_no = conn.wal_frame_count();
 
-        let generation = 1; // TODO: Probe from WAL.
+        let generation = sync_ctx.generation(); // TODO: Probe from WAL.
         let start_frame_no = sync_ctx.durable_frame_num() + 1;
```
penberg, line 406: Now that we cache durable frame number, we need to be prepared for the scenario where remote database got _deleted_. IOW. the durable frame number we have here might be higher than what is actually on server side.

PR #1840, opened 2024-11-25
Make it possible to use offline writes in C bindings

haaawk:
This is to make it possible to test offline writes in opsqlite.
To make the integration easy we don't create a new factory function but modify `libsql_open_sync_with_config` to create offline writes database if primary url contains `offline` query parameter.

### DIFF
--- a/bindings/c/src/lib.rs
+++ b/bindings/c/src/lib.rs
@@ -6,7 +6,7 @@ extern crate lazy_static;
 mod types;
 
 use crate::types::libsql_config;
-use libsql::{errors, LoadExtensionGuard};
+use libsql::{errors, Builder, LoadExtensionGuard};
 use tokio::runtime::Runtime;
 use types::{
     blob, libsql_connection, libsql_connection_t, libsql_database, libsql_database_t, libsql_row,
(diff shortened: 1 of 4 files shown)


penberg (approved):

libsql/src/database/builder.rs
```diff
@@ -408,6 +413,13 @@ cfg_sync! {
             self
         }
 
+        /// Provide a custom http connector that will be used to create http connections.
+        pub fn connector(mut self, connector: HttpsConnector<HttpConnector>) -> Self
```
LucioFranco, line 417: this should be generic over the connector like the other fns in this builder https://github.com/tursodatabase/libsql/blob/7eb42d44712bb27235f2deaf200b289d968dab41/libsql/src/database/builder.rs#L214

LucioFranco (changes requested):

libsql/src/database/builder.rs
haaawk: What's the value of this? I started with generic but then came to the conclusion that this is over-engineering until I really need it.

bindings/c/src/lib.rs
```diff
@@ -152,6 +152,40 @@ pub unsafe extern "C" fn libsql_open_sync_with_config(
             return 3;
         }
     };
+    let uri: http::Uri = match primary_url.try_into(){
+        Ok(uri) => uri,
+        Err(e) => {
+            set_err_msg(format!("Wrong primary URL: {e}"), out_err_msg);
+            return 100;
+        }
```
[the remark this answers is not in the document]
haaawk: This is C bindings specific logic. core crate does not need it. It already has different builders for different versions of database.

libsql/src/database/builder.rs
LucioFranco: It should be a simple copy and paste since we already have the code. But for motivation, 1) our API already follows this path so we should keep them all aligned, 2) Its not really over engineered more so it just exposes the generc bounds that is required for a user to pass in any type of connector. For example, we need this to pass in turmoil connector because if you force this API then we must use the HttpConnector from hyper which binds into tokio::net::TcpStream rather than turmoil::net::TcpStream. This API is too strict and since this is a public API we should keep it generic since changing it can affect users. As I said, it should be a simple copy paste to what the other APIs do but if its not let me know and I can fix it.

LucioFranco: Oh and copying this code should fix the feature flag issues since likely its failing because of unused code during certain feature flag combinations.

bindings/c/src/lib.rs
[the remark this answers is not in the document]
LucioFranco: if I understand this code correctly, its converting some String to a Uri which we already do this logic for other config types https://docs.rs/libsql/latest/libsql/struct.Builder.html#method.new_remote as you can see here we accept a string. We should align them so that the bindings crate doesn't need to import http. I think that doesn't make much sense a consumer of `libsql` shouldn't have to import `http`, this may have been my mistake when I wrote the config.

libsql/src/database/builder.rs
haaawk: It is a public API so ok - I will make it more generic

haaawk: Done.

bindings/c/src/lib.rs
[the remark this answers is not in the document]
LucioFranco: Ah yes, you're right I misread the code, I just went and checked the call site I was worried about. This is fine then thanks for the explanation.

LucioFranco (approved):

PR #1841, opened 2024-11-25
sqld: reject `ATTACH` statements

LucioFranco:
This adds support for rejecting `ATTACH` statements from dumps. `ATTACH` statements should be considered dangerous in a multi-tenant setting. With this commit any `ATTACH` statement inside a dump will be rejected with a 400 status code and a message. In addition, any other sql errors returned by sqlite will be returned as a 400. All other dump errors through this code path (conn.with_raw) will return a 500 like before.

### DIFF
--- a/libsql-server/src/namespace/configurator/helpers.rs
+++ b/libsql-server/src/namespace/configurator/helpers.rs
@@ -11,6 +11,7 @@ use futures::Stream;
 use libsql_sys::EncryptionConfig;
 use libsql_wal::io::StdIO;
 use libsql_wal::registry::WalRegistry;
+use rusqlite::hooks::{AuthAction, AuthContext, Authorization};
 use tokio::io::AsyncBufReadExt as _;
 use tokio::task::JoinSet;
 use tokio_util::io::StreamReader;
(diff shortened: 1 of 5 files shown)


libsql-server/src/namespace/configurator/helpers.rs
```diff
@@ -328,8 +329,31 @@ where
...
+                            AuthAction::Attach { filename: _ } => Authorization::Deny,
+                            _ => Authorization::Allow,
+                        }));
+                        conn.execute(&line, ())
+                    })
+                    .map_err(|e| match e {
+                        rusqlite::Error::SqlInputError {
+                            msg, sql, offset, ..
+                        } => {
+                            let msg = if sql.to_lowercase().contains("attach") {
```
sivukhin, line 343: Can we instead check error code from `SqlInputError` and compare it with something like `AuthorizationForStatementDenied`?

haaawk (commented):
LGTM but I think Nikita's suggestions are good

libsql-server/src/namespace/configurator/helpers.rs
```diff
@@ -328,8 +329,31 @@ where
             line = tokio::task::spawn_blocking({
                 let conn = conn.clone();
                 move || -> crate::Result<String, LoadDumpError> {
-                    conn.with_raw(|conn| conn.execute(&line, ())).map_err(|e| {
-                        LoadDumpError::Internal(format!("line: {}, error: {}", line_id, e))
+                    conn.with_raw(|conn| {
+                        conn.authorizer(Some(|auth: AuthContext<'_>| match auth.action {
```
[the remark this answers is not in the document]
LucioFranco: I want to say we don't re-use libsql connections unless a user does it via hrana etc. So its controlled by the user of the connection api rather than internally.

libsql-server/src/namespace/configurator/helpers.rs
LucioFranco: I was thinking that but when I call https://docs.rs/rusqlite/latest/rusqlite/enum.Error.html#method.sqlite_error  and the other one on `Error` it returns both `None` so I am not totally sure what is going on.

libsql-server/src/namespace/configurator/helpers.rs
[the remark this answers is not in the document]
sivukhin: Ok, just wanted to clarify this as we install authorizer and if connection will be reused - user will be unable to `ATTACH` db (which may be valid action)

libsql-server/src/namespace/configurator/helpers.rs
sivukhin: Ah, sad situation :(

Ok, let's leave it as is then!

PR #1853, opened 2024-11-28
Remove  query parameter from embedded replica url

haaawk:
Otherwise replica fails with:

```
Error syncing database: sync error: failed to push frame: status=400 Bad Request, error={"error":"Protocol error: route not found: [\"?offline\", \"sync\", \"1\", \"1\", \"2\"]"}]
```


penberg (approved):

haaawk:
It turns out offline sync works with url like `libsql://testd-haaawk.turso.io` but it does not work with a trailing slash like `libsql://testd-haaawk.turso.io/`:

```
╰──> export LIBSQL_SYNC_URL=libsql://testd-haaawk.turso.io/ 
╰──> cargo run --example offline_writes                    
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.60s
     Running `target/debug/examples/offline_writes`
Please write your entry to the guestbook:
test6
You entered: test6

Guest book entries:
  test

  tent

  test2

  test3

  test4

  test5

  test6

Syncing database to remote...
thread 'main' panicked at libsql/examples/offline_writes.rs:77:21:
called `Result::unwrap()` on an `Err` value: Sync(PushFrame(404, "{\"error\":\"route not found: [\\\"\\\", \\\"sync\\\", \\\"1\\\", \\\"9\\\", \\\"10\\\"]\"}"))

```

I will have to make sure there's no trailing `/` in the url

Issue #1856, opened 2024-11-29
Recover client from server restarts when handling batons?


penberg:
Currently, I think that when a client holds a server-issued baton and attempts to send it back after a server restart, the request fails with a BatonInvalid error. This error breaks the client's workflow and requires manual intervention.

If this turns out to be the case, let's fix that so thatif a BatonInvalid error occurs and no transaction is in progress, the client should:

* Automatically restart the stream
* Obtain a new baton from the server
* Continue normal operation

penberg:
I was testing the Rust SDK and this only happens if restart happens in the middle of an interactive transaction. We send the following error:

```
thread 'main' panicked at libsql/examples/example.rs:31:23:
called `Result::unwrap()` on an `Err` value: Hrana(Api("Received an invalid baton"))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Which we should improve to tell that the transaction was aborted.

penberg:
I suspected that this would trigger with cursors, but in testing it looks like we recover from restarts fine. Or perhaps I was not trying hard enough to figure out a case where this can happen.

PR #1860, opened 2024-12-01
add "Transaction rolled back" for invalid batons inside transactions

levydsa:
Warn user when a stream is dropped by the server while running a transaction #1856

### DIFF
--- a/libsql/src/hrana/stream.rs
+++ b/libsql/src/hrana/stream.rs
@@ -81,7 +81,12 @@ where
             tracing::trace!("baton not found - skipping finalize for {:?}", req);
             return Ok(false);
         }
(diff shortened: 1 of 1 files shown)


penberg (approved):
Looks good to me. @LucioFranco please review too.

LucioFranco (changes requested):
The PR needs a description with motivation for this change, we should add that and ensure its in the commit messages.

That said, I think it would be smarter to emit this error from the server rather than the client? That way users don't need to update their client to get this improvement?

MrG9090 (approved):

penberg:
Unfortunately there is no trace of us discussing this, but I think we agreed not to do this. Closing.

PR #1864, opened 2024-12-04
Fix Windows build

penberg:


LucioFranco (approved):
LGTM though we should add CI

PR #1870, opened 2024-12-06
libsql: Fix build on older Rust toolchains

penberg:
The JavaScript SDK is unfortunately stuck with Rust 1.78:

https://github.com/tursodatabase/libsql-js/issues/104

We should fix that, but since that's hard, let's just once again paper over it by fixing build on older Rust toolchains.

### DIFF
--- a/libsql/src/sync.rs
+++ b/libsql/src/sync.rs
@@ -309,7 +309,7 @@ impl SyncContext {
     async fn read_metadata(&mut self) -> Result<()> {
         let path = format!("{}-info", self.db_path);
 
-        if !std::fs::exists(&path).map_err(SyncError::io("metadata file exists"))? {
+        if !Path::new(&path).try_exists().map_err(SyncError::io("metadata file exists"))? {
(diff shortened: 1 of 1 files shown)


haaawk (approved):

libsql/src/sync.rs
```diff
@@ -309,7 +309,7 @@ impl SyncContext {
     async fn read_metadata(&mut self) -> Result<()> {
         let path = format!("{}-info", self.db_path);
 
-        if !std::fs::exists(&path).map_err(SyncError::io("metadata file exists"))? {
+        if !Path::new(&path).exists() {
```
haaawk, line 312: This hides potential errors. Is there an API that does not hide them and works for older build?

haaawk: [try_exists](https://doc.rust-lang.org/std/path/struct.Path.html#method.try_exists) seems to do the job and is there since 1.63

haaawk (approved):

PR #1877, opened 2024-12-11
libsql: Fix WAL pull logic on last frame

penberg:
The WAL sync protocol only allows you to fetch frames for a range. While the protocol does not have mechanism to determine what is the latest frame on the server for pull, it does signal with HTTP status code 400 when you attempt to read a frame that does not exists.

Let's use this signal to fix WAL pull logic not to fail always. In the future, we might want to consider extending the protocol to allow clients to proble for latest frame. That, however, will make the protocol a bit more chatty, so I am trying to avoid that as long as I can.

### DIFF
--- a/libsql/src/sync.rs
+++ b/libsql/src/sync.rs
@@ -107,7 +107,7 @@ impl SyncContext {
     }
 
     #[tracing::instrument(skip(self))]
-    pub(crate) async fn pull_one_frame(&mut self, generation: u32, frame_no: u32) -> Result<Bytes> {
+    pub(crate) async fn pull_one_frame(&mut self, generation: u32, frame_no: u32) -> Result<Option<Bytes>> {
         let uri = format!(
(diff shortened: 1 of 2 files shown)


libsql/src/sync.rs
```diff
@@ -256,7 +260,10 @@ impl SyncContext {
                 let frame = hyper::body::to_bytes(res.into_body())
                     .await
                     .map_err(SyncError::HttpBody)?;
-                return Ok(frame);
+                return Ok(Some(frame));
+            }
+            if res.status() == StatusCode::BAD_REQUEST {
+                return Ok(None);
             }
```
LucioFranco, line 267: I am not a big fan of catch all BAD_REQUEST here since there are other areas that can produce this code and we end up eating the response. Should the server return 404 if the frame doesn't exist? Or maybe it should return 200 + an error code.

PR #1887, opened 2024-12-19
Removes sqlite3_config and sqlite3_initialize from local db new

jeremywrowe:
This change removes ffi::sqlite3_config() and ffi::sqlite3_initialize() from local/database.new and moves the configuration down to local/connection#connect. This is done because a SQLite connection can not be initialized multiple times, this change frees users of the library to intialize SQLite connections at a different level. initialize specifically presents a problem when libsql is being included in a shared library where there are other shared libraries that interact with the SQLite C API. This change *should* be relatively safe, since connections can change the threading mode on a per connection basis so long that sqlite was compiled with SQLITE_THREADSAFE=1 or SQLITE_THREADSAFE=2.

Refs:

* https://sqlite.org/threadsafe.html
* https://sqlite.org/c3ref/c_config_covering_index_scan.html#sqliteconfigserialized.

### DIFF
--- a/libsql/src/local/connection.rs
+++ b/libsql/src/local/connection.rs
@@ -38,14 +38,28 @@ impl Connection {
     pub(crate) fn connect(db: &Database) -> Result<Connection> {
         let mut raw = std::ptr::null_mut();
         let db_path = db.db_path.clone();
+        let flags: c_int = ffi::SQLITE_OPEN_FULLMUTEX | db.flags.bits();
+
         let err = unsafe {
+            if ffi::sqlite3_threadsafe() == 0 {
+                // https://sqlite.org/threadsafe.html
+                return Err(Error::ConnectionFailed(String::from(
+                    "SQLite thread safety check failed: SQLite was compiled without thread-safety support \
+                    (SQLITE_THREADSAFE=0). For libsql to operate correctly, SQLite must be compiled with either \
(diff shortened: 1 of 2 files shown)


josh-degraw (approved):

libsql/src/local/connection.rs
```diff
@@ -38,14 +38,28 @@ impl Connection {
     pub(crate) fn connect(db: &Database) -> Result<Connection> {
         let mut raw = std::ptr::null_mut();
         let db_path = db.db_path.clone();
+        let flags: c_int = ffi::SQLITE_OPEN_FULLMUTEX | db.flags.bits();
+
         let err = unsafe {
+            if ffi::sqlite3_threadsafe() == 0 {
```
LucioFranco, line 44: I don't think this is actually safe, because accroding to the docs a user can still enable the threading model in such a way that it invalidates the assumptions the rust code has via `sqlite_confgi`

>  The return value of the sqlite3_threadsafe() function shows only the compile-time setting of thread safety, not any run-time changes to that setting made by sqlite3_config(). 

So this to me isn't strong enough to enforce 100% (esp with sqlite having global config) that the send/sync bounds we have in rust are safely implemented.

LucioFranco (changes requested):

LucioFranco:
closing in favor of https://github.com/tursodatabase/libsql/pull/1891

PR #1898, opened 2024-12-31
sqld: disable checkpoint on primary conn create

LucioFranco:
This commit changes our primary connection initialization code in two ways to achieve the ability to disable checkpointing the wal.

1) We ignore the initial checkpoint that we call directly into sqlite3
   before we restore from bottomless. There is a fixme above that
   explains why we need this but to me right now its not totally clear
   why without digging deeper into the internals of bottomless. We
   should do this but for the moment this unblocks us and from the fixme
   comment it does not sound unsafe rather doing extra work potentially.

2) When bottomless needs to get the local change counter is creates a
   sqlite connection. When this connection drops it seems like it
   checkpoints the wal. I took a brief look at the `sqlite3_close` code
   and did not find anything obvious, I have been told in the past that
   sqlite3 likes to checkpoint at weird points so this could be one of
   those. For the moment, the temporary fix like above is to
   `std::mem::forget` the connection so that `Drop` never gets called
   and thus `sqlite3_close` never gets called.

With both of these changes we now don't checkpoint the wal unless we hit the max size or interval (which for testing I have set very high).

### DIFF
--- a/bottomless/src/replicator.rs
+++ b/bottomless/src/replicator.rs
@@ -935,6 +935,14 @@ impl Replicator {
         })?;
         tracing::trace!("Local change counter: {change_counter}");
 
+        // TODO: we shouldn't leak the connection here but for some reason when this connection get
+        // dropped it seems to checkpoint the database
+        if std::env::var("LIBSQL_BOTTOMLESS_DISABLE_INIT_CHECKPOINTING").is_ok()
+            || std::env::var("LIBSQL_DISABLE_INIT_CHECKPOINTING").is_ok()
+        {
+            std::mem::forget(conn);
+        }
+
         Ok(change_counter.to_be_bytes())
     }
 
--- a/libsql-server/src/namespace/configurator/helpers.rs
+++ b/libsql-server/src/namespace/configurator/helpers.rs
@@ -84,9 +84,16 @@ pub(super) async fn make_primary_connection_maker(
 
     let bottomless_replicator = match primary_config.bottomless_replication {
         Some(ref options) => {
-            tracing::debug!("Checkpointing before initializing bottomless");
-            crate::replication::primary::logger::checkpoint_db(&db_path.join("data"))?;
-            tracing::debug!("Checkpointed before initializing bottomless");
+            // TODO: figure out why we really need this the fixme above is not clear enough but
+            // disabling this allows us to prevent checkpointing of the wal file.
+            if !std::env::var("LIBSQL_DISABLE_INIT_CHECKPOINTING").is_ok() {
+                tracing::debug!("Checkpointing before initializing bottomless");
+                crate::replication::primary::logger::checkpoint_db(&db_path.join("data"))?;
+                tracing::debug!("Checkpointed before initializing bottomless");
+            } else {
+                tracing::warn!("Disabling initial checkpoint before bottomless");
+            }
+
             let options = make_bottomless_options(options, bottomless_db_id, name.clone());
             let (replicator, did_recover) =
                 init_bottomless_replicator(db_path.join("data"), options, &restore_option).await?;


bottomless/src/replicator.rs
```diff
@@ -935,6 +935,12 @@ impl Replicator {
         })?;
         tracing::trace!("Local change counter: {change_counter}");
 
+        // TODO: we shouldn't leak the connection here but for some reason when this connection get
+        // dropped it seems to checkpoint the database
+        if std::env::var("LIBSQL_DISABLE_INIT_CHECKPOINTING").is_ok() {
+            std::mem::forget(conn);
```
haaawk, line 941: Is this safe? @penberg @sivukhin May I borrow your brains here? Isn't leaking sqlite connection leading to some disaster over time?

bottomless/src/replicator.rs
```diff
@@ -935,6 +935,12 @@ impl Replicator {
         })?;
         tracing::trace!("Local change counter: {change_counter}");
 
+        // TODO: we shouldn't leak the connection here but for some reason when this connection get
+        // dropped it seems to checkpoint the database
+        if std::env::var("LIBSQL_DISABLE_INIT_CHECKPOINTING").is_ok() {
```
haaawk, line 940: Let's maybe call it `LIBSQL_DISABLE_BOTTOMLESS_CHECKPOINTING` ? With this change it does not affect only init checkpointing, does it?

libsql-server/src/namespace/configurator/helpers.rs
```diff
@@ -84,9 +84,14 @@ pub(super) async fn make_primary_connection_maker(
 
     let bottomless_replicator = match primary_config.bottomless_replication {
         Some(ref options) => {
-            tracing::debug!("Checkpointing before initializing bottomless");
-            crate::replication::primary::logger::checkpoint_db(&db_path.join("data"))?;
-            tracing::debug!("Checkpointed before initializing bottomless");
+            // TODO: figure out why we really need this the fixme above is not clear enough but
+            // disabling this allows us to prevent checkpointing of the wal file.
+            if std::env::var("LIBSQL_DISABLE_INIT_CHECKPOINTING").is_ok() {
```
haaawk, line 89: I'm not sure whether I understand this. Isn't the code now doing the initial checkpoint only when `LIBSQL_DISABLE_INIT_CHECKPOINTING` is set to anything? That would mean the initial checkpoint is disabled by default and only enabled when [...]_DISABLE_[...] env variable is set, no?

bottomless/src/replicator.rs
LucioFranco: Only reason I didn't call it that was because it felt better to have a single env var for this change and some of the code we feature flagged was in sqld.

libsql-server/src/namespace/configurator/helpers.rs
LucioFranco: Yeah, this has been fixed and will also produce a warn message when this flag is set.

bottomless/src/replicator.rs
LucioFranco: Ok I gave the bottomless code its own feature flag with `LIBSQL_BOTTOMLESS_`, this fits the style we are already using in bottomless.

haaawk (commented):
So now we have both `LIBSQL_BOTTOMLESS_DISABLE_INIT_CHECKPOINTING` and `LIBSQL_DISABLE_INIT_CHECKPOINTING`?

libsql-server/src/namespace/configurator/helpers.rs
```diff
@@ -84,9 +84,16 @@ pub(super) async fn make_primary_connection_maker(
 
     let bottomless_replicator = match primary_config.bottomless_replication {
         Some(ref options) => {
-            tracing::debug!("Checkpointing before initializing bottomless");
-            crate::replication::primary::logger::checkpoint_db(&db_path.join("data"))?;
-            tracing::debug!("Checkpointed before initializing bottomless");
+            // TODO: figure out why we really need this the fixme above is not clear enough but
+            // disabling this allows us to prevent checkpointing of the wal file.
+            if std::env::var("LIBSQL_DISABLE_INIT_CHECKPOINTING").is_err() {
```
haaawk, line 89: Set `LIBSQL_BOTTOMLESS_DISABLE_INIT_CHECKPOINTING` automatically here

LucioFranco:
Merging and releasing this since its behind a feature flag. Platform testing will follow.

PR #1901, opened 2025-01-03
Fix a Y2038 bug by replacing `Int32x32To64` with regular multiplication

CookiePLMonster:
This issue has also been reported in "upstream" SQLite, currently awaiting approval: <https://sqlite.org/forum/forumpost/51e004d479>

`Int32x32To64` macro internally truncates the arguments to int32, while `time_t` is 64-bit on most/all modern platforms. Therefore, usage of this macro creates a Year 2038 bug.

I detailed this issue a while ago in a writeup, and spotted the same issue in this repository when updating the list of affected repositories:
<https://cookieplmonster.github.io/2022/02/17/year-2038-problem/>


CookiePLMonster:
"Upstream" SQLite fixed this bug in https://github.com/sqlite/sqlite/commit/8d6e3f513c049a07d34f77ab526259c916418af6. They are missing a `LL` suffix but with a 64-bit `time_t` this doesn't matter anyway.

Cheaterman:
Please merge Turso Silent Patch, the most legendary collab since PB & J! :-)

PR #1906, opened 2025-01-06
libsql: add `skip_safety_assert` builder option

LucioFranco:
This adds a `skip_saftey_assert` builder option that will allow uses to unsafely opt-out of the safety checks that libsql has built in. The inline doc comments for this function explain more.

### DIFF
--- a/libsql/src/database.rs
+++ b/libsql/src/database.rs
@@ -74,6 +74,7 @@ enum DbType {
         path: String,
         flags: OpenFlags,
         encryption_config: Option<EncryptionConfig>,
+        skip_saftey_assert: bool,
     },
     #[cfg(feature = "replication")]
     Sync {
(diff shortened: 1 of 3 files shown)


jeremywrowe:
Looks good to me!

libsql/src/database.rs
```diff
@@ -74,6 +74,7 @@ enum DbType {
         path: String,
         flags: OpenFlags,
         encryption_config: Option<EncryptionConfig>,
+        skip_saftey_assert: bool,
```
Laerte, line 77: Is this a typo? Title says `skip_safety_assert` but I see references to `skip_saftey_assert`.

LucioFranco: yeah this is def a typo, but its a private type so we can fix it without worry.

PR #1908, opened 2025-01-07
libsql: add per-sync-url metadata

### LINKED ISSUE #1837: Cache durable frame number per sync URL
You can sync a local database to multiple remote databases. Therefore, cache durable frame number per sync URL, not per local database.

LucioFranco:
This commit changes the filename of the metadata file to contain `<db_path>-<sync_url>-info`. Where `db_path` is a full file path + filename and `sync_url` is the host of the authority of the URI. This also includes upfront uri parsing and two new errors that can be produced when trying to extract the host.

This approach trades off creating multiple files in exchange for allowing multiple sync context's to operate at the same time concurrently. Originally, I had approached updating the metadata to include a hashmap of endpoints and their metadata but this approach falls flat if you have multiple `Database` and thus `SyncContext` in the same process (and beyond).

This commit does NOT include updating from the original v0 metadata format but instead will force a full re-sync and a new meatadata file will be produced with verison set to `1`. Not implementing this mean't simpler code with less space to produce errors and since this only runs once when a user upgrades that cost is okay.

Closes #1837

### DIFF
--- a/libsql/src/sync.rs
+++ b/libsql/src/sync.rs
@@ -1,10 +1,10 @@
 use crate::{util::ConnectorService, Result};
 
-use std::path::Path;
+use std::{path::Path, str::FromStr};
 
 use bytes::Bytes;
 use chrono::Utc;
-use http::{HeaderValue, StatusCode};
+use http::{uri::InvalidUri, HeaderValue, StatusCode, Uri};
 use hyper::Body;
 use tokio::io::AsyncWriteExt as _;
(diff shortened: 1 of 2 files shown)


LucioFranco:
This is ready for review but is in draft since I want to do some manual testing before we ship it.

penberg:
Let's instead change the metadata JSON format to support multiple URLs and have a single file.

PR #1911, opened 2025-01-10
Fix cross compilation and build for sandboxes enviroments.

levydsa:
Build scripts shouldn't write to anything but `OUT_DIR`. So, instead of copying `sqlite3.c` to sqlite3mc in source, copy it to `OUT_DIR` and build from there.

https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script
> Build scripts may save any output files or intermediate artifacts in the directory specified in the [OUT_DIR environment variable](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts). Scripts should not modify any files outside of that directory.

On some environments, Nix derivations, the permissions of source files are limited, and get copied over using the standard `fs::copy`. To prevent this, `cp --no-preserve=mode,ownership` is needed. Inside a Nix derivation, some coreutils are overridden and this is fine to use even on a MacOS device. To make this more reliable, if the `cp` command fails, do a normal copy with `fs` functions.

Some CMake configuration was added to allow cross-compilation of sqlite3mc for more targets.

### DIFF
--- a/libsql-ffi/build.rs
+++ b/libsql-ffi/build.rs
@@ -362,19 +362,28 @@ fn build_multiple_ciphers(target: &str, out_path: &Path) {
         .unwrap();
 
     if let Some(ref cc) = cross_cc {
-        if cc.contains("aarch64") && cc.contains("linux") {
-            cmake_opts.push(&cmake_toolchain_opt);
-            writeln!(toolchain_file, "set(CMAKE_SYSTEM_NAME \"Linux\")").unwrap();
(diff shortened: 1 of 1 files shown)


levydsa:
The `copy_with_cp` name is a bit misleading, but should be fine. Maybe changing it to just `copy` is fine?

penberg:
Hey @levydsa, I opened a draft pull request to update `libsql-js.git` to this, which is a good test if build breaks on some platform: https://github.com/tursodatabase/libsql-js/pull/173

PR #1916, opened 2025-01-12
fix: offline writes

john-kohout-e2g:


penberg:
Fixed by https://github.com/tursodatabase/libsql/commit/111c27762a62c71a17c48eff8e447df1878a19fe as far as I can tell.

PR #1928, opened 2025-01-22
WAL sync checkpointing support

penberg:
This pull request adds WAL sync checkpointing support.

### DIFF
--- a/libsql/src/sync.rs
+++ b/libsql/src/sync.rs
@@ -61,16 +61,23 @@ impl SyncError {
     }
 }
 
+pub enum PullResult {
+    /// A frame was successfully pulled.
+    Frame(Bytes),
+    /// We've reached the end of the generation.
+    EndOfGeneration { max_generation: u32 },
+}
+
 pub struct SyncContext {
     db_path: String,
     client: hyper::Client<ConnectorService, Body>,
     sync_url: String,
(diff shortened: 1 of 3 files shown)


libsql/src/sync.rs
```diff
@@ -532,29 +568,60 @@ async fn try_pull(
...
     loop {
+        let generation = sync_ctx.durable_generation();
+        let frame_no = sync_ctx.durable_frame_num() + 1;
         match sync_ctx.pull_one_frame(generation, frame_no).await {
-            Ok(Some(frame)) => {
+            Ok(PullResult::Frame(frame)) => {
                 insert_handle.insert(&frame)?;
-                frame_no += 1;
+                sync_ctx.durable_frame_num = frame_no;
             }
```
levydsa, line 582: We only need to insert here, apparently. Can we instead of maintaining the invariants of `WalInsertHandle` with a `RefCell` and check at runtime, use inherited mutability at comptime do the job by creating and dropping the handle on each frame? I introduced `WalInsertHandle` in a previous PR to be able to do that, but I might have encoded the wrong semantics. What you think?

penberg: I am not sure I understand what you mean. If you want, please submit a PR to do what you propose and happy to merge if it's cleaner.

PR #1950, opened 2025-02-08
Push up to 128 frames in sync

haaawk:
In my tests, this speeds up push 4-5 times.
Making push of 300 frames take 4-5s instead of 20s.

### DIFF
--- a/libsql/src/sync.rs
+++ b/libsql/src/sync.rs
@@ -120,6 +120,10 @@ impl SyncContext {
         Ok(me)
     }
 
+    pub fn set_push_batch_size(&mut self, push_batch_size: u32) {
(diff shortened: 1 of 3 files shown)


libsql/src/sync.rs
```diff
@@ -537,19 +538,25 @@ async fn try_push(
 
     let mut frame_no = start_frame_no;
     while frame_no <= end_frame_no {
-        let frame = conn.wal_get_frame(frame_no, page_size)?;
+        let batch_size = 128.min(end_frame_no - frame_no + 1);
```
penberg, line 541: We need to make this configurable so that clients with really crappy networking can make it much smaller.

haaawk: Done.

PR #1954, opened 2025-02-13
Fix busy timeouts

sivukhin:
This PR setup properly setup busy timeout for `libsql-server` connections:
1. It uses 100ms busy timeout by default
2. It uses 5s busy timeout for connection which execute `wal_checkpoint(TRUNCATE)`

### DIFF
--- a/libsql-server/src/connection/connection_core.rs
+++ b/libsql-server/src/connection/connection_core.rs
@@ -1,4 +1,3 @@
-use std::ffi::{c_int, c_void};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


libsql-server/src/connection/connection_core.rs
```diff
@@ -88,18 +87,6 @@ impl<W: Wal + Send + 'static> CoreConnection<W> {
             config.max_row_size as i32,
         );
 
-        unsafe {
```
haaawk, line 91: What's the effect of this being removed?

sivukhin: We will have default busy handler which will abort after busy timeout and internally sleep and wake up with some intervals

penberg (approved):

PR #1966, opened 2025-02-25
Make client work with libsql:// schema again

haaawk:
This fixes the error like:
```
Error opening db path test.db, primary url libsql://testd.turso.io: sync error: error trying to connect: unsupported scheme libsql
```
when using libsql:// in the url


levydsa (approved):
LGTM

PR #1968, opened 2025-02-26
make wal checkpoint not fallback to passive mode if busy lock not acq…

pereman2:
…uired

Wal truncate checkpoint may not acquire write lock on WAL and it will fallback to passive mode making it proceed with checkpoint even though we wanted to skip checkpoint altogether.

### DIFF
--- a/libsql-sqlite3/src/wal.c
+++ b/libsql-sqlite3/src/wal.c
@@ -4131,12 +4131,9 @@ static int sqlite3WalCheckpoint(
     */
     if( eMode!=SQLITE_CHECKPOINT_PASSIVE ){
       rc = walBusyLock(pWal, xBusy2, pBusyArg, WAL_WRITE_LOCK, 1);
-      if( rc!=SQLITE_OK ){
-        walUnlockExclusive(pWal, WAL_CKPT_LOCK, 1);
-        pWal->ckptLock = 0;
-        return rc;
+      if( rc==SQLITE_OK ){
+        pWal->writeLock = 1;
       }
-      pWal->writeLock = 1;
     }
   }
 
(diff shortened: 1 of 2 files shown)


libsql-sqlite3/src/wal.c
```diff
@@ -4129,13 +4129,12 @@ static int sqlite3WalCheckpoint(
...
-      if( rc==SQLITE_OK ){
-        pWal->writeLock = 1;
-      }else if( rc==SQLITE_BUSY ){
-        eMode2 = SQLITE_CHECKPOINT_PASSIVE;
-        xBusy2 = 0;
-        rc = SQLITE_OK;
+      if( rc!=SQLITE_OK ){
+        walUnlockExclusive(pWal, WAL_CKPT_LOCK, 1);
+        pWal->ckptLock = 0;
+        return rc;
```
jussisaurio: If on line 4118 already `rc != SQLITE_OK` then the code continues to do the following things:

```
  walDisableBlocking(pWal);
  sqlite3WalDb(pWal, 0);

  /* Release the locks. */
  sqlite3WalEndWriteTransaction(pWal);
  if( pWal->ckptLock ){
    walUnlockExclusive(pWal, WAL_CKPT_LOCK, 1);
    pWal->ckptLock = 0;
  }
```

Are you sure we don't just want to omit the `else if` and let the code fall through?

pereman2: I changed to fallback already :)

jussisaurio: ok i am slowpoke

avinassh (approved):

pereman2:
We found a workaround for this :)

PR #1971, opened 2025-02-27
fix(libsql-sys): support non-Unix when rusqlite feature is disabled

CBenoit:
Currently the codepath used when rusqlite is disabled, is Unix-platform-dependent.

For instance, here is the error we compiling from Windows:

```
error[E0433]: failed to resolve: could not find `unix` in `os`
   --> C:\Users\runneradmin\.cargo\git\checkouts\libsql-311658d335deb3b1\9b04f1d\libsql-sys\src\connection.rs:276:26
    |
276 |             use std::os::unix::ffi::OsStrExt;
    |                          ^^^^ could not find `unix` in `os`
    |
note: found an item that was configured out
   --> /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14\library\std\src\os\mod.rs:27:9
note: the item is gated here
   --> /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14\library\std\src\os\mod.rs:19:1
note: found an item that was configured out
   --> /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14\library\std\src\os\mod.rs:65:9
note: the item is gated here
   --> /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14\library\std\src\os\mod.rs:64:1

error[E0599]: no method named `as_bytes` found for reference `&OsStr` in the current scope
   --> C:\Users\runneradmin\.cargo\git\checkouts\libsql-311658d335deb3b1\9b04f1d\libsql-sys\src\connection.rs:277:73
    |
277 |             let path = std::ffi::CString::new(path.as_ref().as_os_str().as_bytes())
    |                                                                         ^^^^^^^^
    |
help: there is a method `as_encoded_bytes` with a similar name
    |
277 |             let path = std::ffi::CString::new(path.as_ref().as_os_str().as_encoded_bytes())
    |                                                                         ~~~~~~~~~~~~~~~~

Some errors have detailed explanations: E0433, E0599.
For more information about an error, try `rustc --explain E0433`.
error: could not compile `libsql-sys` (lib) due to 2 previous errors
```

This patch is adding a platform-independent branch.

The database path is expected to be UTF-8 by the libsql native library. However, in the real world, all Unix paths are not necessarily UTF-8. For this reason, I understand why the Unix codepath is attempting a direct conversion from OsString to CString.
However, it’s not possible to do something similar on other platforms. For these platforms, we are enforcing correct UTF-8 using `to_str`. At least, it should work reasonably well in most cases.

### DIFF
--- a/libsql-sys/src/connection.rs
+++ b/libsql-sys/src/connection.rs
@@ -273,9 +273,27 @@ impl<W: Wal> Connection<W> {
 
         #[cfg(not(feature = "rusqlite"))]
         let conn = unsafe {
-            use std::os::unix::ffi::OsStrExt;
-            let path = std::ffi::CString::new(path.as_ref().as_os_str().as_bytes())
-                .map_err(|_| crate::error::Error::Bug("invalid database path"))?;
+            #[cfg(unix)]
+            let path = {
+                use std::os::unix::ffi::OsStrExt;
+                std::ffi::CString::new(path.as_ref().as_os_str().as_bytes()).map_err(|_| {
+                    crate::error::Error::Bug(
+                        "invalid database path containing an internal nul byte",
+                    )
+                })?
+            };
+            #[cfg(not(unix))]
+            let path = path
+                .to_str()
+                .ok_or_else(|| crate::error::Error::Bug("database path is not valid unicode"))
+                .and_then(|x| {
+                    std::ffi::CString::new(x).map_err(|_| {
+                        crate::error::Error::Bug(
+                            "invalid database path containing an internal nul byte",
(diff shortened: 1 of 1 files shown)


CBenoit:
Hello @penberg
I’m not sure how my change is impacting the "Go bindings tests" job.

penberg:
@CBenoit Please rebase on top of `main`, you're just getting hit by unrelated bug that is fixed in `main`.

CBenoit:
@penberg I see! Done!

PR #1981, opened 2025-03-11
cmake(libsql-ffi): don't use `-msse4.2` `-maes` when compiling for ARM hosts

bengsparks:
The`libsql-ffi/bundled/SQLite3MultipleCiphers/CMakeLists.txt` intentionally avoids adding `-msse4.2` and `-maes`, only to unconditionally add the flags when compiling using Clang, regardless of architecture.
This PR also updates a similar check that appears in a conditional branch when compiling with GCC.


penberg:
@bengsparks Can you rebase on top of `main`? The test failure is unrelated to your PR and just recently fixed.

bengsparks:
@penberg done :)

PR #1994, opened 2025-03-20
libsql: Improve sync protocol probe tracing

penberg:


haaawk (approved):

haaawk:
It seems that you need to guard the import with sync feature.

PR #1998, opened 2025-03-21
libsql: Check integrity during WAL pull

penberg:
The check is expensive, but let's add it to catch errors early.

### DIFF
--- a/libsql/src/local/connection.rs
+++ b/libsql/src/local/connection.rs
@@ -3,6 +3,7 @@
 use crate::local::rows::BatchedRows;
 use crate::params::Params;
 use crate::{connection::BatchRows, errors};
+use crate::value::Value;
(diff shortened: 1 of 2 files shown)


libsql/src/local/connection.rs
```diff
@@ -568,6 +569,18 @@ impl Connection {
...
+        let stmt = self.prepare("PRAGMA integrity_check")?;
+        let rows = stmt.query(&Params::None)?;
+        let mut result = true;
+        if let Some(row) = rows.next()? {
+            if row.get_value(0)? != Value::Text("ok".to_string()) {
+                result = false;
+            }
+        }
+        Ok(result)
+    }
```
avinassh, line 583: i would prefer returning an error instead of bools

PR #2021, opened 2025-04-09
Bootstrap database before opening connection

avinassh:
This patch is a follow up on https://github.com/tursodatabase/libsql/pull/2017 It changes the heuristics for bootstrapping the DB. 

One more change is, it also does bootstrap before any connection to the local SQLite is created 

Earlier, we bootstrapped whenever we noticed a change in the generation. However, this patch changes this behaviour and does bootstrap only if local files don't exist:

1. if no db file or the metadata file exists, then user is starting from scratch
and we will do the sync
2. if the db file exists, but the metadata file does not exist (or other way around),
then local db is in an incorrect state. we stop and return with an error
3. if the db file exists and the metadata file exists, then we don't need to do the
sync

### DIFF
--- a/libsql/src/local/database.rs
+++ b/libsql/src/local/database.rs
@@ -465,8 +465,8 @@ impl Database {
     /// Sync WAL frames to remote.
     pub async fn sync_offline(&self) -> Result<crate::database::Replicated> {
         let mut sync_ctx = self.sync_ctx.as_ref().unwrap().lock().await;
+        crate::sync::bootstrap_db(&mut sync_ctx).await?;
         let conn = self.connect()?;
-
         crate::sync::sync_offline(&mut sync_ctx, &conn).await
     }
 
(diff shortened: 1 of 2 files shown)


libsql/examples/remote_sync.rs
```diff
@@ -48,7 +48,13 @@ async fn main() {
...
     db.sync().await.unwrap();
     println!(" done");
-
+    {
+        let mut stmt = conn.prepare("PRAGMA integrity_check").await.unwrap();
+        let mut rows = stmt.query(()).await.unwrap();
+        let result: String = rows.next().await.unwrap().unwrap().get(0).unwrap();
+        assert_eq!(result, "ok", "integrity check failed after sync");
+        println!("Integrity check passed after sync");
+    }
```
penberg, line 57: Please drop this and let's instead work on #1998 to do this internally.

penberg:
I think we should just make `connect()` bootstrap if needed to avoid people having to call `sync()`. If there's some problem with that, then `connect()` at minimum needs to return an error telling people to sync first for empty database.

avinassh:
I have opened an issue to track integrity check and bootstrapping on connect

https://github.com/tursodatabase/libsql/issues/2024 
https://github.com/tursodatabase/libsql/issues/2023

PR #2033, opened 2025-04-16
Fix `block_on` usage in `connect`

### LINKED ISSUE #2032: `db.connect` panics when used in a single threaded environment
Consider this code (introduced via this PR: https://github.com/tursodatabase/libsql/pull/2029)

https://github.com/tursodatabase/libsql/blob/10f5dd836f93a673ca95d5134dd06110dcb2c9b2/libsql/src/database.rs#L685-L695

If libsql library is used in a single threaded environment, then it fails with following error:

> can call blocking only when running on the multi-threaded runtime

This is happening due to incorrect use of `block_in_place` which should be used only in multiple threaded environments.

avinassh:
fixes #2032

### DIFF
--- a/libsql/src/database.rs
+++ b/libsql/src/database.rs
@@ -683,35 +683,18 @@ impl Database {
                 use tokio::sync::Mutex;
 
                 if let Some(sync_ctx) = &db.sync_ctx {
-                    let sync_ctx = sync_ctx.clone();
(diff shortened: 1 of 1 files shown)


avinassh:
Okay, this might not be the proper way to handle this. 

https://greptime.com/blogs/2023-03-09-bridging-async-and-sync-rust

https://stackoverflow.com/a/66280983

penberg (approved):

Issue #1384, opened 2024-05-12
iOS/Android support


ospfranco:
I'm trying to build static library (.a) that I can use on iOS and Android. I'm trying to compile [the C experimental](https://github.com/tursodatabase/libsql/tree/main/bindings/c) bindings, doing `cargo build --target [TARGET] --release` works and compiles the library correctly. 

The problem however lies when I try to create all three iOS archs (`x86_64-apple-ios`, `aarch64-apple-ios`, `aarch64-apple-ios-sim`) and then generate a single fat binary for `x86_64-apple-ios` and `aarch64-apple-ios-sim`, which is needed to package all of them in a single `.xcframework`

I get the following error:

```
lipo -create ../../target/x86_64-apple-ios/release/libsql_experimental.a ../../target/aarch64-apple-ios-sim/release/libsql_experimental.a -output generated/simulator_fat/libsql_experimental.a
fatal error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: archive member ../../target/x86_64-apple-ios/release/libsql_experimental.a(sqlite3mc.c.o) cputype (16777228) and cpusubtype (0) does not match previous archive members cputype (16777223) and cpusubtype (3) (all members must match)
make: *** [libsql.xcframework] Error 1
```

Which seems to indicate the sqlite3mc.c object file has not been compiled for the correct architecture.

Are there some instructions to compile the library to iOS/Android? I've seen at least one person that has managed to compile it for Android.

haaawk:
I successfully ran libsql on Android. I don't think I had to do anything special. What I do though is having a new Rust lib project that depends on libsql.

ospfranco:
Running lipo on the generated `.a` for `x86_64` allowed me to pin point the problem:

```
lipo -info ../../target/x86_64-apple-ios/release/libsql_experimental.a                                                                                   
fatal error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: archive member ../../target/x86_64-apple-ios/release/libsql_experimental.a(sqlite3mc.c.o) cputype (16777228) and cpusubtype (0) does not match previous archive members cputype (16777223) and cpusubtype (3) (all members must match)
```

[...]

PR #2070, opened 2025-05-18
Enable encryption for mobile targets

winghouchan:
### Why?

#1423 introduced support for React Native however encryption was disabled due to the difficulty in getting successful builds with encryption enabled:

> - Encryption seems to be compiling sqlcipher (?) via cmake, which also includes OpenSSL flags, the problem the compilation step is hard to integrate with Android (and that's why I had to disable it). Someone else that knows the internals of libsql should give it a try.

– https://github.com/tursodatabase/libsql/pull/1423#issuecomment-2186259989

With encryption disabled, libSQL on React Native does not have feature parity with other platforms.

### What?

This PR enables encryption for builds against mobile targets. However, while the encryption feature is enabled, there is additional work to get the encryption feature fully complete:

> P.S. if you do turn on encryption, it will require a modification to the bindings to send the encryption key from JS, so, open a PR when the time comes and I will take a look.

– https://github.com/tursodatabase/libsql/pull/1423#issuecomment-2186259989

### How?

- The encryption feature of the libSQL dependency in the C bindings package has been enabled for all targets.
- Build scripts and configurations in the libSQL FFI package have been updated so that libSQL products for each target and the [OP SQLite Example App](https://github.com/OP-Engineering/op-sqlite/tree/main/example) build successfully; and the tests against the OP SQLite Example App pass.

I'm not actually familiar with Android/iOS/C/Rust build systems, or libSQL / OP SQLite internals so configuration decisions were based on a feedback loop of running `make <android|ios>` in `bindings/c` and making updates until builds in libSQL and the OP SQLite Example App were successful. I'm unsure if there's anything else to check and will have to defer to Turso team members and @ospfranco for their knowledge.


bindings/c/Makefile
```diff
@@ -3,6 +3,7 @@ CFLAGS := -Iinclude
 LDFLAGS := -lm
 ARCHS_IOS = x86_64-apple-ios aarch64-apple-ios aarch64-apple-ios-sim
 ARCHS_ANDROID = aarch64-linux-android armv7-linux-androideabi x86_64-linux-android i686-linux-android
+ANDROID_PLATFORM := 31
```
winghouchan, line 6: > [!NOTE]
>
> This value comes from the value already passed to the `platform` option when running `cargo ndk`. It has been pulled out and assigned here so that it can also be passed as an environment variable for libSQL FFI's build script which uses it to set the `ANDROID_PLATFORM` variable when running CMake.

libsql-ffi/build.rs
```diff
@@ -452,7 +452,7 @@ fn build_multiple_ciphers(target: &str, out_path: &Path) {
     let bundled_dir = format!("{out_dir}/sqlite3mc");
     let sqlite3mc_build_dir = env::current_dir().unwrap().join(out_dir).join("sqlite3mc");
 
-    let mut cmake_opts: Vec<&str> = vec![];
+    let mut cmake_opts: Vec<String> = vec![];
```
winghouchan, line 455: > [!NOTE]
>
> `cmake_opts` changed from `Vec<&str>` to `Vec<String>` to allow for interpolating options values into a string with their option names.

libsql-ffi/build.rs
```diff
@@ -508,20 +515,58 @@ fn build_multiple_ciphers(target: &str, out_path: &Path) {
...
+    cmake_opts.push("-DCMAKE_BUILD_TYPE=Release".to_string());
+    cmake_opts.push("-DSQLITE3MC_STATIC=ON".to_string());
+    cmake_opts.push("-DCODEC_TYPE=AES256".to_string());
+    cmake_opts.push("-DSQLITE3MC_BUILD_SHELL=OFF".to_string());
+    cmake_opts.push("-DSQLITE_SHELL_IS_UTF8=OFF".to_string());
+    cmake_opts.push("-DSQLITE_USER_AUTHENTICATION=OFF".to_string());
+    cmake_opts.push("-DSQLITE_SECURE_DELETE=OFF".to_string());
+    cmake_opts.push("-DSQLITE_ENABLE_COLUMN_METADATA=ON".to_string());
+    cmake_opts.push("-DSQLITE_USE_URI=ON".to_string());
+    cmake_opts.push("-DCMAKE_POSITION_INDEPENDENT_CODE=ON".to_string());
```
winghouchan, line 527: > [!NOTE]
>
> Changing `cmake_opts` from `Vec<&str>` to `Vec<String>` necessitated calling `to_string` on these string literals.
>
> This seems very verbose but I'm not familiar with Rust so please let me know if there is a better way.

libsql-ffi/build.rs
```diff
@@ -462,8 +462,15 @@ fn build_multiple_ciphers(target: &str, out_path: &Path) {
...
-    let cmake_toolchain_opt = "-DCMAKE_TOOLCHAIN_FILE=toolchain.cmake".to_string();
+    let ndk_cmake_toolchain_path = env::var("CARGO_NDK_CMAKE_TOOLCHAIN_PATH").ok();
+    let toolchain_path = ndk_cmake_toolchain_path
+        .clone()
+        .map(PathBuf::from)
+        .unwrap_or_else(|| sqlite3mc_build_dir.join("toolchain.cmake"));
+    let cmake_toolchain_opt = ndk_cmake_toolchain_path
+        .clone()
+        .map(|path| format!("-DCMAKE_TOOLCHAIN_FILE={}", path))
+        .unwrap_or_else(|| "-DCMAKE_TOOLCHAIN_FILE=toolchain.cmake".to_string());
```
winghouchan, line 473: > [!NOTE]
>
> Use the Android NDK toolchain if available (determined by the value of the `CARGO_NDK_CMAKE_TOOLCHAIN_PATH`). This seems to be necessary for a successful build for Android.

libsql-ffi/build.rs
```diff
@@ -508,20 +515,58 @@ fn build_multiple_ciphers(target: &str, out_path: &Path) {
...
+    cmake_opts.push("-DSQLITE_ENABLE_COLUMN_METADATA=ON".to_string());
+    cmake_opts.push("-DSQLITE_USE_URI=ON".to_string());
+    cmake_opts.push("-DCMAKE_POSITION_INDEPENDENT_CODE=ON".to_string());
 
     if target.contains("musl") {
-        cmake_opts.push("-DCMAKE_C_FLAGS=\"-U_FORTIFY_SOURCE\" -D_FILE_OFFSET_BITS=32");
-        cmake_opts.push("-DCMAKE_CXX_FLAGS=\"-U_FORTIFY_SOURCE\" -D_FILE_OFFSET_BITS=32");
+        cmake_opts.push("-DCMAKE_C_FLAGS=\"-U_FORTIFY_SOURCE\" -D_FILE_OFFSET_BITS=32".to_string());
+        cmake_opts
+            .push("-DCMAKE_CXX_FLAGS=\"-U_FORTIFY_SOURCE\" -D_FILE_OFFSET_BITS=32".to_string());
```
winghouchan, line 532: > [!NOTE]
>
> Changing `cmake_opts` from `Vec<&str>` to `Vec<String>` necessitated calling `to_string` on these string literals.
>
> This seems very verbose but I'm not familiar with Rust so please let me know if there is a better way.

libsql-ffi/build.rs
```diff
@@ -508,20 +515,58 @@ fn build_multiple_ciphers(target: &str, out_path: &Path) {
...
+            "x86_64-linux-android" => "x86_64",
+            _ => panic!("Unsupported Android target: {}", target),
+        };
+        let android_platform = std::env::var("ANDROID_PLATFORM")
+            .expect("ANDROID_PLATFORM environment variable must be set");
+
+        cmake_opts.push(cmake_toolchain_opt);
+        cmake_opts.push(format!("-DANDROID_ABI={}", android_abi));
+        cmake_opts.push(format!("-DANDROID_PLATFORM=android-{}", android_platform));
+    }
```
winghouchan, line 549: > [!IMPORTANT]
>
> This is probably the critical change to build for Android. It configures CMake in the following ways:
>
> - Sets `CMAKE_TOOLCHAIN_FILE` to the path to the Android NDK toolchain file (available via the `CARGO_NDK_CMAKE_TOOLCHAIN_PATH` environment variable).
> - Sets the `ANDROID_ABI` variable according to the defined target.
> - Sets the `ANDROID_PLATFORM` variable to the platform level specified in the C bindings package manifest, via an environment variable named `ANDROID_PLATFORM`.

libsql-ffi/build.rs
```diff
@@ -508,20 +515,58 @@ fn build_multiple_ciphers(target: &str, out_path: &Path) {
...
+                "iphoneos"
+            };
+            ("arm64", "arm64", sysroot)
+        } else {
+            panic!("Unsupported iOS target: {}", target);
+        };
+
+        cmake_opts.push(format!("-DCMAKE_OSX_ARCHITECTURES={}", arch));
+        cmake_opts.push(format!("-DCMAKE_SYSTEM_PROCESSOR={}", processor));
+        cmake_opts.push(format!("-DCMAKE_OSX_SYSROOT={}", sysroot));
```
winghouchan, line 569: > [!IMPORTANT]
>
> This is probably the critical change to build for iOS. It configures CMake in the following ways:
>
> - Sets the `CMAKE_OSX_ARCHITECTURES` variable to the architecture according to the target.
> - Sets the `CMAKE_SYSTEM_PROCESSOR` variable to the processor according to the target.
> - Sets the `CMAKE_OSX_SYSROOT` variable to `iphonesimulator` if the target is a simulator or the x86_64 architecture (there are no physical x86_64 iPhones) or `iphoneos` if the target is a physical device.

winghouchan:
@ospfranco: given my unfamiliarity, would you mind expanding on [this comment on #1423](https://github.com/tursodatabase/libsql/pull/1423#issuecomment-2186259989) please?

> P.S. if you do turn on encryption, it will require a modification to the bindings to send the encryption key from JS, so, open a PR when the time comes and I will take a look.

winghouchan:
Related: #1384

winghouchan:
I see there's some logic for cross compilation: https://github.com/tursodatabase/libsql/blob/c6ebbe1b23a5dc7d950b2a4ad33fc33c849e462f/libsql-ffi/build.rs#L457-L509

Would it be better to somehow _work within this_, instead of the changes currently proposed (79ff682)?

penberg:
@levydsa do you remember why we disabled encryption for mobile targets?

levydsa:
@penberg, because it was hard to cross-compile sqlite3mc to mobile targets. But apparently, its possible with a few tweaks.

tyrauber:
Thank you @winghouchan, for picking this up. libsql is fantastic on mobile devices (thanks to op-sqlite), but the lack of database encryption is definitely problematic.  I would love to see this get merged in.

winghouchan:
@Geal: interesting, thanks for sharing those details. Looking at a diff/blame between `0.9.17` and `0.9.19` might be interesting. Btw, if you think it's suitable, it may be worth making a PR that supersedes this PR with your changes. Your changeset is probably smaller than mine and will work with the changes introduced by #2099.

PR #2080, opened 2025-05-26
Prevent holding lock to context between sync interval ticks

### LINKED ISSUE #2076: Embedded replicas (and offline sync) databases are slow if you set `sync_interval`
If you set `sync_interval`, then `sync` calls take as much as time the `sync_interval` value is set to. This issue is observed in JS SDK (https://github.com/tursodatabase/libsql-js) and this doesn't happen in Rust client. 

There are two discord users reporting the same and following links also contain reproducers: 

https://discord.com/channels/933071162680958986/1374700570514030592

https://discord.com/channels/933071162680958986/1374075561017741312/1375111427890614435

levydsa:
Fixes #2076.

### DIFF
--- a/libsql/src/database/builder.rs
+++ b/libsql/src/database/builder.rs
@@ -645,10 +645,14 @@ cfg_sync! {
             let mut bg_abort: Option<std::sync::Arc<crate::sync::DropAbort>> = None;
 
             if let Some(sync_interval) = sync_interval {
+                let sync_span = tracing::debug_span!("sync_interval");
+                let _enter = sync_span.enter();
+
                 let sync_ctx = db.sync_ctx.as_ref().unwrap().clone();
                 {
                     let mut ctx = sync_ctx.lock().await;
                     crate::sync::bootstrap_db(&mut ctx).await?;
+                    tracing::debug!("finished bootstrap with sync interval");
                 }
 
                 // db.connect creates a local db file, so it is important that we always call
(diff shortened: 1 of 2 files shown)


libsql/src/sync.rs
```diff
@@ -893,6 +895,8 @@ pub async fn try_pull(
     let mut err = None;
 
     loop {
+        tokio::task::yield_now().await;
```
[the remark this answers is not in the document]
levydsa: I was unable to reproduce the results I was seeing yesterday, rendering my analysis completely wrong. I was thinking that `try_pull` was somehow now yield back execution and that `yield_now` would help with that. Testing now, `yield_now` does nothing. What made the difference was that `interval.tick()` was called before acquiring the lock (I still don't fully understand why that would be a problem, very curious).

I'm changing to a more robust (i hope) approach with a oneshot channel for cancellation. Seems to avoid these issues altogether.

libsql/src/database/builder.rs
```diff
@@ -644,38 +644,61 @@ cfg_sync! {
...
                 let conn = db.connect()?;
-                let jh = tokio::spawn(
+
+                tokio::spawn(
                     async move {
+                        let mut interval = tokio::time::interval(sync_interval);
+
                         loop {
-                            tracing::trace!("trying to sync");
+                            tracing::info!("trying to sync");
```
PThorpe92, line 671: nit: could this loop be a little loud for an `info!` log?

libsql/src/database/builder.rs
```diff
@@ -644,38 +644,61 @@ cfg_sync! {
...
+
+                            interval.tick().await;
```
PThorpe92, line 673: I think from looking at the issue, it looks like we probably want to sync at the beginning of the task first before the loop? Or just avoid calling `tick` before the first sync

PR #2089, opened 2025-06-01
Optimise `sync_pull` to pull frames in batches

avinassh:
Previously, we pulled frames one by one. This patch changes it pull frames in batches. Currently, the batch size is set to 128 (the maximum supported by the server)

### DIFF
--- a/libsql/src/sync.rs
+++ b/libsql/src/sync.rs
@@ -20,6 +20,7 @@ const METADATA_VERSION: u32 = 0;
 
 const DEFAULT_MAX_RETRIES: usize = 5;
 const DEFAULT_PUSH_BATCH_SIZE: u32 = 128;
+const DEFAULT_PULL_BATCH_SIZE: u32 = 128;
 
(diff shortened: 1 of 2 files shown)


avinassh:
This PR should be merged once the server side changes are deployed. Changing it to draft for now.

libsql/src/sync.rs
```diff
@@ -185,9 +190,11 @@ impl SyncContext {
             self.sync_url,
             generation,
             frame_no,
-            frame_no + 1
+            // the server expects the range of [start, end) frames, i.e. end is exclusive
+            // so we add +1 so that we can pull in the batch size of `self.pull_batch_size`
+            frame_no + self.pull_batch_size
```
khuezy, line 195: According to the comment, it says "add +1", but I don't see that.

avinassh: @khuezy thanks for taking a look! I am still experimenting with the batch sizes, so the comment is not up to date.

penberg (approved):

Issue #1859, opened 2024-11-30
Unsound Use of str::from_utf8_unchecked in keyword_token


lwz23:
**Description**
The keyword_token function uses unsafe { str::from_utf8_unchecked(word) } to convert a byte slice (&[u8]) into a string slice (&str) without validating whether the input is valid UTF-8. This introduces undefined behavior (UB) if the word parameter contains invalid UTF-8 bytes. The absence of validation makes the function unsound.
https://github.com/tursodatabase/libsql/blob/9241b00b819c6a0fe846747b1ccc91e6c244cdcf/vendored/sqlite3-parser/src/dialect/mod.rs#L69
```
pub fn keyword_token(word: &[u8]) -> Option<TokenType> {
    KEYWORDS
        .get(UncasedStr::new(unsafe { str::from_utf8_unchecked(word) }))
        .cloned()
}

```
**Problems:**
this function is a `pub` function, so I assume user can control the `word` field, it cause some problems.
1. Undefined Behavior on Invalid UTF-8:
unsafe { str::from_utf8_unchecked(word) } assumes that the word slice is valid UTF-8. If this assumption is violated, undefined behavior occurs immediately.
The function does not verify that word is valid UTF-8 before invoking the unsafe conversion.
2. No Safety Contract:
The function is not marked as unsafe, nor does it document the requirement that the word input must be valid UTF-8. This makes it easy for callers to misuse the function by passing invalid inputs.
3. Potential Exploitation:
If word is derived from untrusted or external input, it could contain invalid UTF-8. This could lead to crashes, memory corruption, or other unpredictable behavior.

**Suggestion**
1. mark this function as unsafe and provide safety doc.
2. add some check in the function body eg. use `from_utf8` instead.

**Additional Context:**
Unsafe code should only be used when safety invariants are strictly guaranteed. The current implementation assumes that the word input is always valid UTF-8, but this is not enforced or documented, making the function unsound. By switching to std::str::from_utf8, the function can remain safe and robust while handling invalid input gracefully.

lwz23:
after some test, I think it is ok to use `from_utf8_unchecked` here, I didn't meet any crush with the following code:
```
extern crate sqlite3_parser;

use sqlite3_parser::dialect::keyword_token;

fn main() {
    let invalid_utf8: Vec<u8> = vec![0xF0, 0x9F, 0x28, 0xFF, 0xF0, 0x28, 0x8C, 0x28];

[...]

PR #2093, opened 2025-06-05
fix: make keyword_token safe by validating UTF-8 input

kmr-ankitt:
This PR fixes an unsound usage of unsafe { str::from_utf8_unchecked(word) } in the public function keyword_token in mod.rs.

The function now uses std::str::from_utf8(word).ok()? to safely handle invalid UTF-8, eliminating the unsoundness.
No logic or API changes.
Code compiles and tests pass (where possible).

Closes: https://github.com/tursodatabase/libsql/issues/1859

### DIFF
--- a/vendored/sqlite3-parser/src/dialect/mod.rs
+++ b/vendored/sqlite3-parser/src/dialect/mod.rs
@@ -65,9 +65,8 @@ include!(concat!(env!("OUT_DIR"), "/keywords.rs"));
 pub(crate) const MAX_KEYWORD_LEN: usize = 17;
 
 pub fn keyword_token(word: &[u8]) -> Option<TokenType> {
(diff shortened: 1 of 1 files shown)


penberg:
Hey @kmr-ankitt, thanks, please consider submitting this upstream too. Would also love to have this in https://github.com/tursodatabase/limbo where we hard forked the parser.

penberg:
Hey @kmr-ankitt, there are other places that call `from_utf8_unchecked()` too, almost certainly for performance reasons. As far as I can tell, it's probably better to attempt to fix this in the upstream repository and then merge back, instead of fixing parts of it in `libsql.git`. Happy to also discuss fixing this in https://github.com/tursodatabase/limbo where we hard forked the parser.

PR #2108, opened 2025-06-22
libsql: Open a fresh stream in HttpConnection::prepare()

penberg:
The server expires idle streams in 10 seconds. Therefore, let's not even attempt to reuse the same stream for preparing a new SQL statement.

### DIFF
--- a/libsql/src/hrana/connection.rs
+++ b/libsql/src/hrana/connection.rs
@@ -81,7 +81,7 @@ where
     }
 
     pub async fn prepare(&self, sql: &str) -> crate::Result<Statement<T>> {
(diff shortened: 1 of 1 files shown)


libsql/src/hrana/connection.rs
```diff
@@ -81,7 +81,7 @@ where
     }
 
     pub async fn prepare(&self, sql: &str) -> crate::Result<Statement<T>> {
-        let stream = self.current_stream().clone();
```
khuezy, line 84: Did the expired stream get properly cleaned up?
Is it possible to check the current stream to see if it's a live and reuse it, if it's dead open a new stream - does that improve anything?

PR #2171, opened 2025-10-04
Add get/set reserved bytes to C bindings

levydsa:
- Exposes the API to the `Connection` struct
- Adds `libsql_set_reserved_bytes` and `libsql_get_reserved_bytes` to `libsql.h`

### DIFF
--- a/bindings/c/src/lib.rs
+++ b/bindings/c/src/lib.rs
@@ -550,6 +550,41 @@ pub unsafe extern "C" fn libsql_load_extension(
 }
 
 #[no_mangle]
+pub unsafe extern "C" fn libsql_set_reserved_bytes(
+    conn: libsql_connection_t,
+    reserved_bytes: i32,
+    out_err_msg: *mut *const std::ffi::c_char,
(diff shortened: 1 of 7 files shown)


bindings/c/src/lib.rs
```diff
@@ -549,6 +549,41 @@ pub unsafe extern "C" fn libsql_load_extension(
...
+    out_err_msg: *mut *const std::ffi::c_char,
+) -> std::ffi::c_int {
+    if conn.is_null() {
+        set_err_msg("Null connection".to_string(), out_err_msg);
+        return 1;
+    }
+    let conn = conn.get_ref();
+    match conn.get_reserved_bytes() {
+        Ok(v) => *reserved_bytes = v,
+        Err(err) => set_err_msg(err.to_string(), out_err_msg),
```
avinassh, line 582: we should return `1` if the error is set right? but this returns `0` regardless

avinassh (changes requested):

avinassh (approved):

PR #2208, opened 2026-01-08
Replace unmaintained `bincode` dependency with `wincode` for serialization/deserialization

ocean:
`bincode` has been marked as unmaintained ([RUSTSEC-2025-0141](https://rustsec.org/advisories/RUSTSEC-2025-0141.html)).

This commit replaces it with [`wincode`](https://crates.io/crates/wincode), which:
- Maintains full wire protocol compatibility with `bincode`
- Is actively maintained by the [Anza XYZ team](https://crates.io/users/anza-team)
- Offers better performance according to benchmarks

### DIFF
--- a/libsql-replication/src/generated/proxy.rs
+++ b/libsql-replication/src/generated/proxy.rs
@@ -180,7 +180,7 @@ pub struct Description {
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct Value {
-    /// / bincode encoded Value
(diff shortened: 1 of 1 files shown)


ocean:
Much harder than I thought as this has broken the Go bindings in `go-libsql` in the tests, due to bytecode incompatibility of the serialization output from `postcard`, hence I have marked this as draft while thinking about it, as well as giving time for smarter people to join in.

ocean:
After further investigation, I've now replaced `bincode` with `wincode` as discussed in the main PR text. Let's see if the tests pass now.

PR #2219, opened 2026-03-18
Docs: Enhance USER_GUIDE with JWT authentication and key generation

Andres6936:
Added complete Javascript/Typescript examples to the documentation for generating Ed25519 keys, signing JWTs with the `jose` library, and authenticating the `@libsql/client`, including token expiration details.


Andres6936:
tap, tap, tap, this is working...?

PR #2230, opened 2026-04-22
Reset replication endpoint

sle-c:


sle-c:
close, needs testing before opening on the source project

PR #2237, opened 2026-05-18
Update SQLite to 3.47.2

azizcodesthings:
## Summary
- update the vendored SQLite source tree to 3.47.2
- preserve libSQL-specific parser, shell, pager, WAL, and planner integrations during the upstream merge
- regenerate bundled sqlite3 amalgamations, headers, and Rust bindgen outputs

## Test plan
- cargo xtask build-bundled
- LIBSQL_DEV=1 LIBSQLITE3_SYS_BUNDLING=1 cargo check -p libsql-ffi
- LIBSQL_DEV=1 LIBSQLITE3_SYS_BUNDLING=1 cargo check -p libsql-ffi --features session
- ./libsql-sqlite3/libsql -version

## Notes
- build/checks emit existing generated-C warnings for ALIGN macro redefinition and a non-void generated function return path
- configure warns that Tcl regression tests cannot run locally because no tclConfig.sh was found


penberg:
Fixed by https://github.com/tursodatabase/libsql/pull/2240

PR #2252, opened 2026-06-30
savepoint forget hook

jussisaurio:


libsql-sqlite3/src/pager.c
```diff
@@ -7066,6 +7066,14 @@ int sqlite3PagerSavepoint(Pager *pPager, int op, int iSavepoint){
     nNew = iSavepoint + (( op==SAVEPOINT_RELEASE ) ? 0 : 1);
     for(ii=nNew; ii<pPager->nSavepoint; ii++){
       sqlite3BitvecDestroy(pPager->aSavepoint[ii].pInSavepoint);
+      if( pagerUseWal(pPager)
```
sivukhin, line 7068: Is it bit more safer to cal `xSavepointForget` before any destruction from SQLite? Just worry that if SQLite will add some `sqlite3DataDestroy(pPager->aSavepoint[ii].aWalData)` later here - we will screw up with ordering

sivukhin (approved):
If this is eps-effort - should we ask clanker to write simple tests which will check that savepoint methods will be called in some expected order for some statements?

PR #2255, opened 2026-07-12
libsql-server: Don't fail namespace load on corrupt stats file

### LINKED ISSUE #2254: Database getting corrupted on power loss
Database getting corrupted, there was a powerloss, the database got currepted, even when there were no active transaction being written, had to remove it.
sqlite and sql database like mariadb completely fine.

Error: Internal Error: `EOF while parsing a value at line 1 column 0`

Caused by:
    EOF while parsing a value at line 1 column 0

penberg:
If the server loses power, the stats file can end up empty because the stats persist loop did not sync the file to disk before renaming it over the old one. On restart, the stats file failed to parse, which errored out namespace load, and took the whole server down:

    Error: Internal Error: `EOF while parsing a value at line 1 column 0`

    Caused by:
        EOF while parsing a value at line 1 column 0

Stats are disposable telemetry, so fall back to default stats with a warning when the stats file does not parse, and sync the stats file to disk before renaming it to close the truncation window.

Fixes #2254

### DIFF
--- a/libsql-server/src/stats.rs
+++ b/libsql-server/src/stats.rs
@@ -246,7 +246,18 @@ impl Stats {
         let stats_path = db_path.join("stats.json");
         let mut this = if stats_path.try_exists()? {
             let data = tokio::fs::read_to_string(&stats_path).await?;
-            serde_json::from_str(&data)?
+            match serde_json::from_str(&data) {
+                Ok(stats) => stats,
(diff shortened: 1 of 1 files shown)


janus-reith:
@penberg Just ran into this. Any chance of getting this reviewed and into a version tagged release?

PR #2265, opened 2026-08-03
libsql-ffi: Fix Windows build with encryption feature enabled

sreejsanas:
When the encryption feature is enabled, `copy_with_cp` copies the `SQLite3MultipleCiphers` directory into OUT_DIR. `cp` isn't available on Windows, so it falls back to `fs::copy`, which cannot copy a directory.

The previous fallback only routed to `copy_dir_all` when `fs::copy` returned `ErrorKind::InvalidInput`, but Windows returns `PermissionDenied` for a directory, so the copy was never retried and the build failed.

Detect directories explicitly with `is_dir()` instead of relying on a platform-specific error kind. Unix/macOS/Nix behavior is unchanged.

### DIFF
--- a/libsql-ffi/build.rs
+++ b/libsql-ffi/build.rs
@@ -88,11 +88,9 @@ fn copy_with_cp(from: impl AsRef<Path>, to: impl AsRef<Path>) -> io::Result<()> 
         .status()
     {
         Ok(status) if status.success() => Ok(()),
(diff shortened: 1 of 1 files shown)


sreejsanas:
@jussisaurio Is it possible to merge this into 0.10 release? this fails windows builds when encryption is enabled on libsql.

Windows checks job is failing with visual studio not installed, i did verify this change on macos/linux/windows(x86_64, arm64)/ios too locally

